PR #14, opened 2020-06-02
Add utility function to calculate degree heating days

drewjgray:


api/utils/temperature.js
```diff
@@ -0,0 +1,61 @@
...
+
+function getDHD(reefID) {
+    /**
+     * Calculates the Degree Heating Days of a reef location. 
+     *
+     * HS = SST(daily) - MMM if SST(daily) > MMM
+     * HS = 0                if SST(daily) <= MMM
+     * HS > 1C               bleaching threshold
+     *
+     * @param {float}  SST           Sea Surface Temperature [°C] (SST)
```
ericboucher, line 26: you need to decide if these are parameters of your function, or if they are fetched within the function

drewjgray:
Could go either way. Think we'll be calling these apis from within this function?

api/utils/temperature.js
drewjgray: Could go either way, not sure how you're structuring the data calls. Think we'll be calling these apis from within this function?

api/utils/temperature.js
```diff
@@ -0,0 +1,61 @@
...
+
+    let SST = getSST(reefID);
+    let MMM = getMMM(reefID)
+    let HS = SST.map(value => {
+        return (value - MMM > 0 ? value - MMM : 0);
+    });
+
+    let DHD = 0;
+    HS.forEach(HSday => {
+        if (HSday >= 1) {
```
ericboucher, line 41: I think you could do all this in one go and test if `value - MMM > 1` directly

api/utils/temperature.js
```diff
@@ -0,0 +1,61 @@
+// Corals start to become stressed when the SST is 1°C warmer than the maxiumum monthly mean temperature (MMM).
```
ericboucher, line 1: I'd recommend using the /** notation you used below as well for this block

api/utils/temperature.js
ericboucher: I think we ca rename this function `calculateDHD` and it should take as parameters everything it needs as you described in the params.

So no reefId, but instead: SST: number[], MMM: number. I don't think we need HS as an input, right?

api/utils/temperature.js
drewjgray: I left it like this in case we want to use the HS value for anything. Also for clarity for other viewers of the code.

api/utils/temperature.js
drewjgray: Let's change when we know the final architecture.

api/src/utils/temperature.test.ts
```diff
@@ -0,0 +1,25 @@
+import { calculateDegreeHeatingDays } from '.';
```
ericboucher, line 1: @bencpeters I couldn't get this import to work with '.' nor './temperature', any ideas?

bencpeters: I wouldn't expect `.` to work, that only works with `index`. But `./temperature` should work. The problem is that `temperature` doesn't export the function `calculateDegreeHeatingDays` though...

api/src/utils/temperature.ts
```diff
@@ -0,0 +1,44 @@
...
+ *      No Stress (no heat stress or bleaching is present): HotSpot of less than or equal to 0. 
+ *      Bleaching Watch (low-level heat stress is present): HotSpot greater than 0 but less than 1; SST below bleaching threshhold. 
+ *      Bleaching Warning (heat stress is accumulating, possible coral bleaching): HotSpot of 1 or greater; SST above bleaching threshold; DHW greater than 0 but less than 4. 
+ *      Bleaching Alert Level 1 (significant bleaching likely): HotSpot of 1 or greater; SST above bleaching threshold; DHW greater than or equal to 4 but less than 8. 
+ *      Bleaching Alert Level 2 (severe bleaching and significant mortality likely): HotSpot of 1 or greater; SST above bleaching threshold; DHW greater than or equal to 8. 
+ * 
+ * DHW = (1/7)*sum[1->84](HS(i) if HS(i) >= 1C)
+**/
+
+function calculateDegreeHeatingDays(seaSurfaceTemperatures: number[], maximumMonthlyMean: number) {
```
bencpeters, line 19: Export this function.

api/src/utils/temperature.ts
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * HS = SST(daily) - MMM if SST(daily) > MMM
+     * HS = 0                if SST(daily) <= MMM
+     * HS > 1C               bleaching threshold
+     *
+     * @param {float[]}    seaSurfaceTemperatures        list of seaSurfaceTemperatures
+     * @param {float}      maximumMonthlyMean            maximumMonthlyMean for this location
+     *
+     * @return {float} DHD           Degree Heating Days
+     */
```
bencpeters, line 31: Doc strings generally go above the function declaration in JS.

api/src/utils/temperature.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (seaSurfaceTemperatures.length !== 84){
+        throw new Error('Calculating Degree Heating Days requires exactly 84 days of data.');
+    }
+
+    const degreeHeatingDays = seaSurfaceTemperatures.map(
+        value => {
+            const degreeDeviation = value - maximumMonthlyMean
+            return (degreeDeviation >= 1 ? value - maximumMonthlyMean : 0);
+        }).reduce(function(a, b){ return a + b }, 0)
```
bencpeters, line 41: Generally we prefer to use ES6 style function declarations for anonymous functions:

`reduce((a, b) => a + b, 0)`

api/src/utils/temperature.ts
```diff
@@ -0,0 +1,44 @@
...
+     * @param {float}      maximumMonthlyMean            maximumMonthlyMean for this location
+     *
+     * @return {float} DHD           Degree Heating Days
+     */
+
+    if (seaSurfaceTemperatures.length !== 84){
+        throw new Error('Calculating Degree Heating Days requires exactly 84 days of data.');
+    }
+
+    const degreeHeatingDays = seaSurfaceTemperatures.map(
```
bencpeters, line 37: This first loop through the data seems unnecessary - you can do this calculation and the cumsum operation in a single pass with reduce:

```
seaSurfaceTemperatures.reduce((sum, value) => {
  // deviation calculation
  return sum + degreeDeviation;
}, 0);
```

api/yarn.lock
```diff
@@ -777,6 +777,11 @@
   resolved "https://registry.yarnpkg.com/@types/normalize-package-data/-/normalize-package-data-2.4.0.tgz#e486d0d97396d79beedd0a6e33f4534ff6b4973e"
   integrity sha512-f5j5b/Gf71L+dbqxIpQ4Z2WlmI/mPJ0fOkGGmFgtb6sAu97EPczzbS3/tJKxmcYDj55OX6ssqwDAWOHIYDRDGA==
 
+"@types/parse-json@^4.0.0":
```
bencpeters, line 780: Are these updates all the result of adding `lint-staged`? I'm surprised the `yarn.lock` updated this much with those changes to `package.json`...

ericboucher: Looks like it...

PR #15, opened 2020-06-04
Implement db schema in TypeORM

vickykont-programize:
Implement the DB schema outlined in https://dbdiagram.io/d/5ebc6bb239d18f5553ff2a85 into TypeORM


api/src/spotters/spotters.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+  @Index({ spatial: true })
+  location: string;
+
+  @Column({ length: 50 })
+  hardware_version: string;
+
+  @ManyToOne(() => Reef)
+  @JoinColumn({ name: "reef_id" })
+  reef_id: Reef;
+}
```
ericboucher, line 25: Let's make sure we have newlines at the end of files for consistency

api/migration/1591285575946-InitDB.ts
```diff
@@ -0,0 +1,84 @@
...
+        await queryRunner.query(`ALTER TABLE "monthly_temperature_threshold" DROP CONSTRAINT "FK_f76ea5d5e9fa63e2c08747a3d2d"`);
+        await queryRunner.query(`ALTER TABLE "monthly_temperature_threshold" DROP CONSTRAINT "FK_d69b2e6aaf16f588be54301915a"`);
+        await queryRunner.query(`ALTER TABLE "daily_data" DROP CONSTRAINT "FK_281bc33ffb88dab13d8112d22e1"`);
+        await queryRunner.query(`ALTER TABLE "reef_point_of_interest" DROP CONSTRAINT "FK_4a2bdb6a3cff9c746be703224dc"`);
+        await queryRunner.query(`ALTER TABLE "reef" DROP CONSTRAINT "FK_63c63e16bcebffa7ab94aaddbdb"`);
+        await queryRunner.query(`ALTER TABLE "reef" DROP CONSTRAINT "FK_dc56bfd6bfcd1f221ec83885294"`);
+        await queryRunner.query(`ALTER TABLE "reef" DROP CONSTRAINT "FK_073078b9b04d1501c3c373fe1b8"`);
+        await queryRunner.query(`ALTER TABLE "region" DROP CONSTRAINT "FK_2547f55866c0274d35fdd9a29c0"`);
+        await queryRunner.query(`DROP INDEX "IDX_332a5f87cb3709f88980816dab"`);
+        await queryRunner.query(`DROP TABLE "survey_video"`);
```
ericboucher, line 64: I am a bit surprised that we are ending with all these DROPs without recreating any index and tables after, no?

api/src/reefs/interfaces/reefs.dto.ts
```diff
@@ -1,13 +1,13 @@
 export class ReefDto {
     readonly name: string;
     readonly polygon: string;
-    readonly region_id: number;
+    readonly region_id: any;
```
ericboucher, line 4: why do we use "any" here?

api/migration/1591285575946-InitDB.ts
```diff
@@ -0,0 +1,84 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class InitDB1591285575946 implements MigrationInterface {
+    name = 'InitDB1591285575946'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE TABLE "region" ("id" SERIAL NOT NULL, "name" character varying(50) NOT NULL, "polygon" polygon NOT NULL, "parent_id" integer, CONSTRAINT "PK_5f48ffc3af96bc486f5f3f3a6da" PRIMARY KEY ("id"))`);
```
bencpeters, line 7: In Postgres, it's generally negligible from a performance standpoint to use `varchar` vs. `text` fields, so I'd almost always go for `text` instead for the added flexibility. Let's make sure all these text fields are `text` columns.

api/migration/1591285575946-InitDB.ts
```diff
@@ -0,0 +1,84 @@
...
+        await queryRunner.query(`CREATE INDEX "IDX_9229452ed71aae8c51844ce86e" ON "survey_image" USING GiST ("location") `);
+        await queryRunner.query(`CREATE TABLE "survey_video" ("id" SERIAL NOT NULL, "location" point NOT NULL, "url" character varying NOT NULL, "start_timestamp" TIMESTAMP NOT NULL, "end_timestamp" TIMESTAMP NOT NULL, "upload_timestamp" TIMESTAMP NOT NULL, "quality" integer NOT NULL DEFAULT 1, "important" boolean NOT NULL, "hidden" boolean NOT NULL, "metadata" json NOT NULL, "survey_id" integer, "user_id" integer, "poi_id" integer, CONSTRAINT "PK_9272f4b8dbe71d5f65d56c22c7a" PRIMARY KEY ("id"))`);
+        await queryRunner.query(`CREATE INDEX "IDX_332a5f87cb3709f88980816dab" ON "survey_video" USING GiST ("location") `);
+        await queryRunner.query(`ALTER TABLE "region" ADD CONSTRAINT "FK_2547f55866c0274d35fdd9a29c0" FOREIGN KEY ("parent_id") REFERENCES "region"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`);
+        await queryRunner.query(`ALTER TABLE "reef" ADD CONSTRAINT "FK_073078b9b04d1501c3c373fe1b8" FOREIGN KEY ("region_id") REFERENCES "region"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`);
```
bencpeters, line 27: All these foreign keys have `on delete no action`. This will prevent deletion if the related entity is deleted, making it impossible to e.g. delete a user. In general when there's an ownership condition, we probably want to use the `onDelete: 'CASCADE'` in the decorator.

api/migration/1591285575946-InitDB.ts
bencpeters: This is the down migration, it's expected.

api/src/reef-pois/reef-pois.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+  @Column()
+  poi_label_id: number;
+
+  @Column()
+  image_url: string;
+
+  @Column({ length: 50 })
+  name: string;
+
+  @ManyToOne(() => Reef)
```
bencpeters, line 20: We should have a `onDelete` specifier here.

api/src/reef-pois/reef-pois.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+} from 'typeorm';
+import { Reef } from '../reefs/reefs.entity';
+
+@Entity()
+export class ReefPointOfInterest {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column()
+  poi_label_id: number;
```
bencpeters, line 12: Let's make the Entities use `camelCase` identifiers to be consistent with the rest of the codebase (e.g. `imageUrl`). We can then use [this plugin](https://www.npmjs.com/package/typeorm-naming-strategies) to `snake_case` the DB column names.

bencpeters, line 12: Is this a join column? We should specify a relationship if yes...

api/src/reefs/daily-data.entity.ts
```diff
@@ -0,0 +1,62 @@
...
+  @Column("float")
+  min_bottom_temperature: number;
+
+  @Column("float")
+  max_bottom_temperature: number;
+
+  @Column("float")
+  avg_bottom_temperature: number;
+
+  @Column("float")
```
bencpeters, line 23: Can this be a decimal value? I wonder if it should be an integer type instead?

api/src/reefs/daily-data.entity.ts
```diff
@@ -0,0 +1,62 @@
...
+  max_wind_speed: number;
+
+  @Column("float")
+  avg_wind_speed: number;
+
+  @Column()
+  wind_direction: number;
+
+  @ManyToOne(() => Reef)
```
bencpeters, line 59: `onDelete`

api/src/reefs/interfaces/reefs.dto.ts
bencpeters: +1

bencpeters:
Also, when you submit a PR, can you please write a brief description of the changes in it?

api/migration/1591285575946-InitDB.ts
ericboucher: My bad, I hadn't seen the up/down structure

api/src/reefs/daily-data.entity.ts
ericboucher: Yes it can be a decimal

api/migration/1591285575946-InitDB.ts
bencpeters: Actually, I think it basically just doesn't matter with `varchar` vs `text`. Disregard this, it's going to be more trouble than it's worth to manually specify if TypeORM "wants" to generate `varchar` instead of `text`.

PR #16, opened 2020-06-04
Initialize reef detail page

tleraj:


website/src/helpers/requests.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface Request {
+  method: AxiosRequestConfig["method"];
+  url: AxiosRequestConfig["url"];
+  data?: AxiosRequestConfig["data"];
+  params?: AxiosRequestConfig["params"];
+  token?: string;
+  responseType?: AxiosRequestConfig["responseType"];
+}
+
+export type promiseType = (arg0: Request) => Promise<AxiosResponse<any>>;
```
bencpeters, line 36: I think you can make these types a little more useful (`any` isn't a very useful type in general - it tosses out most of the type safety we get with TS).

Also, _generally_ types are named with `PascalCase`.

I'd probably do this:

```typescript
function send <T> (request: Request): Promise<AxiosResponse<T>> {
  const headers = ...;
  return agent.request<T>({ ... });
}

// Then to use:
type ExpectedData = {
  field: string;
}
// myRespData will be of type Promise<AxiosResponse<ExpectedData>>
const myRespData = send<ExpectedData>(req);
```

website/src/index.css
```diff
@@ -1,4 +1,6 @@
 body {
+  width: 100vw;
+  height: 100vh;
```
bencpeters, line 3: `app.css` already defines some width & height values for `html` and `body`, is this needed?

website/src/layout/App/index.tsx
```diff
@@ -1,19 +1,23 @@
...
   return (
     <ThemeProvider theme={theme}>
       <Router>
-        <NavBar />
+        {/* <NavBar /> */}
 
         <div id="app">
           <Switch>
+            <Route exact path="/" component={HomePage} />
+            <Route exact path="/reefs" component={Reefs} />
```
bencpeters, line 20: Are we expecting any subpaths under `/reefs`? (e.g. `/reefs/:reefId`?) If so, we probably don't want `exact` here, so that we can use nested routing...

website/src/routes/Homepage/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+};
+
+const styles = (theme: Theme) => {
+  return createStyles({
+    homepageContent: {
+      display: "flex",
+      flexDirection: "column",
+      alignItems: "center",
+      justifyContent: "center",
```
bencpeters, line 35: Could you just use the [Material UI Grid component](https://material-ui.com/components/grid/) for this?

website/src/routes/Homepage/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      height: "100vh",
```
bencpeters, line 36: Is the intention to eventually have a navbar? This will cause issues if we do - the Navbar will render above this page, so `100vh` will add a vertical scroll.

I think a better layout is to have full viewport height set on `html` & `body`, then have this height be `100%` of its parent (which is the main root/app container, `#app`), then have `#app` have a flex grow property that allows it to share space with the navbar, but take up all remaining vertical space.

website/src/routes/Homepage/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      backgroundColor: theme.palette.text.secondary
+    },
+    welcomeMessage: {
+      color: theme.palette.text.primary,
+      marginBottom: "3rem"
```
bencpeters, line 41: I'd like to avoid having custom CSS all over the place. In this case, you can easily use the `color=primary` property on [`Typography`](https://material-ui.com/api/typography/), and perhaps `gutterBottom` (and control the height of that gutter on a theme-basis so that it's consistent across the site?)

website/src/routes/Homepage/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+    },
+    buttonWrapper: {
+      display: "flex",
+      alignItems: "center",
+      justifyContent: "center"
+    },
+    button: {
+      backgroundColor: theme.palette.primary.main,
```
bencpeters, line 49: Material UI `Button` has a `color` property, could you just use that? Also, unless there's a special design reason, I think we'd rather define things like `borderRadius` for buttons in the theme so that they're consistent across the whole site.

website/src/routes/Reefs/Reef/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  Typography
+} from "@material-ui/core";
+import { Map, TileLayer } from "react-leaflet";
+import { useSelector } from "react-redux";
+
+import ReefNavBar from "../ReefNavBar";
+import { RootState } from "../../../store/configure";
+
+const Reef = ({ classes }: ReefProps) => {
+  const reef = useSelector((state: RootState) => state.reef);
```
bencpeters, line 17: Let's use the [Redux Slices Pattern](https://redux-toolkit.js.org/usage/usage-guide#creating-slices-of-state) from the redux toolkit for organizing redux state. In that pattern, selectors generally get defined in the slice that they're associated with.

website/src/routes/Reefs/ReefNavBar/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    managerInfo: {
+      display: "flex",
+      alignItems: "center"
+    },
+    managerDetails: {
+      display: "flex",
+      flexDirection: "column",
+      alignItems: "flex-end"
+    },
+    managerAvatar: {
```
bencpeters, line 87: Could you achieve this same layout with `Grid` components instead of all these custom CSS classes?

website/src/routes/Reefs/ReefNavBar/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+      marginLeft: "1rem"
+    }
+  });
+};
+
+export interface ReefNavBarStyleProps extends WithStyles<typeof styles> {}
```
bencpeters, line 93: Do you need this intermediate (I'm legitimately not sure)? I wonder if this would work:

```
type NavBarProps = WithStyles<typeof styles> & ReefNavBarIncomingProps
```

I also would only `export` things that are potentially useful to another component. In this case, I think only the incoming props is potentially useful, I'd keep the rest local to this file.

website/src/routes/Reefs/ReefNavBar/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export interface ReefNavBarIncomingProps {
+  reefName: string | null;
```
bencpeters, line 96: Do you have a reason to prefer `null` props to `undefined` for these? It seems like `undefined` might be a little easier just because that's what JS/TS give by default when accessing a property that doesn't exist on an object - with `null` you have to provide `null` explicitly. 

TS also has a special syntax for optional props if they can be `undefined`:

```typescript
interface ReefNavBarIncomingProps {
  reefName?: string;
}
```

website/src/routes/Reefs/ReefNavBar/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <div className={classes.lastSurveyed}>
+              <Typography variant="subtitle2">{`Last surveyed: ${lastSurvey}`}</Typography>
+            </div>
+          </div>
+        </div>
+        <div className={classes.managerInfo}>
+          <div className={classes.managerDetails}>
+            <Typography variant="caption">
+              {`Managed by: ${managerName}`}
+            </Typography>
```
bencpeters, line 44: In your props, you marked `reefName`, `lastSurvey`, and `managerName` as optional. If they're truly optional props, add conditional rendering to only render the relevant blocks if the prop exists. If they're not optional, make sure that the `Props` type has them as required so that Typescript can provide type safety for us (require us to provide them)

website/src/routes/Reefs/ReefNavBar/index.test.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
 import { render } from "@testing-library/react";
-import NavBar from ".";
+import ReefNavBar from ".";
 
 test("renders as expected", () => {
-  const { container } = render(<NavBar />);
+  const { container } = render(<ReefNavBar />);
```
bencpeters, line 6: Hrm, it doesn't look like this snapshot is rendering anything... Let's make sure it's working (it might require a mock of the Material UI components).

website/src/routes/Reefs/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+    dispatch(reefActions.reefRequest("1"));
+  }, [dispatch]);
+
+  return (
+    <div>
+      {/* 
+        Should render a list of all available Reefs. By clicking to a specific
+        reef, the user will be able to see its details. 
+      */}
+      <Reef />
```
bencpeters, line 20: Let's set up routing for this. Add `Reef` as a nested route that takes `reefId` as a path param (you'll have to remove `exact` from the `Reefs` definition in `App`), then have the `Reef` component dispatch its own data request.

website/src/services/reefServices.ts
```diff
@@ -0,0 +1,14 @@
+import { AxiosResponse } from "axios";
+
+import requests from "../helpers/requests";
+
+const getReef = (id: string): Promise<AxiosResponse<any>> => {
+  return requests.send({
+    url: `/reefs/${id}`,
+    method: "GET"
+  });
+};
```
bencpeters, line 10: If we're going through the trouble of defining a service, I'd avoid requiring the consumer of this service to have to unwrap the `AxiosResponse`. This service function should return `Promise<Reef>`

(Optionally, I think it might be worth modifying `requests.send` to unwrap the payload too - basically making `send <T> = (request: Request) => Promise<T>` rather than `Promise<AxiosResponse<T>>`)

website/src/store/Reefs/actions.ts
```diff
@@ -0,0 +1,58 @@
...
+    payload: error
+  };
+};
+
+const reefRequestSetLoading = (loading: Reef["loading"]): reefActionTypes => {
+  return {
+    type: constants.REEF_REQUEST_SET_LOADING,
+    payload: loading
+  };
+};
```
bencpeters, line 28: Use Redux toolbox, it encodes a bunch of Redux best practices that I'd like to use for this app. This should be a [slice](https://redux-toolkit.js.org/usage/usage-guide#creating-slices-of-state).

website/src/store/Reefs/actions.ts
```diff
@@ -0,0 +1,58 @@
...
+      };
+      dispatch(reefRequestSuccess(reef));
+    } catch (error) {
+      dispatch(reefRequestFailure(error.message));
+      throw error;
+    } finally {
+      dispatch(reefRequestSetLoading(false));
+    }
+  };
+};
```
bencpeters, line 51: Use Redux toolkit's [`createAsyncThunk`](https://redux-toolkit.js.org/usage/usage-guide#async-requests-with-createasyncthunk) for this.

website/src/store/Reefs/types.ts
```diff
@@ -0,0 +1,31 @@
+import constants from "./constants";
+
+export interface Reef {
+  details: {
+    id: string | null;
+    name: string | null;
+    managerName: string | null;
```
bencpeters, line 7: Let's make this the `Reef` interface (again, figure out whether `id`, `name`, `managerName` are actually required properties...). Then we'll also have a type `ReefState` which has `loading` & `error`, but this way we can use `Reef` as a type for components/functions that act on reef information:

```typescript
type Reef = {
  id: string;
  name: string;
  managerName?: string;
}

type ReefState = {
  reef?: Reef;
  loading: boolean;
  error?: string;
}

File: website/src/store/configure.ts [hunk 19]
```diff
@@ -0,0 +1,38 @@
...
+);
+
+export type RootState = ReturnType<typeof store.getState>;
+export type AppThunk<ReturnType = void> = ThunkAction<
+  ReturnType,
+  RootState,
+  unknown,
+  Action<string>
+>;
+export type AppActions = reefActionTypes;
```
bencpeters, line 38: Use Redux Toolkit to set up the store, and it will automatically include `redux-thunk` middleware.

They have a [good guide for usage with Typescript](https://redux-toolkit.js.org/usage/usage-with-typescript) too.

bencpeters:
Also, in the future, when you submit a PR, can you please add a little summary of what is implemented in the PR? If you add pages, include screenshots of the pages.

website/README.md
```diff
@@ -1,51 +1,35 @@
...
 
-This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app), using the [Redux](https://redux.js.org/) and [Redux Toolkit](https://redux-toolkit.js.org/) template.
+### Requirements
 
-- **Styling & UI Library** Use [Material UI](https://material-ui.com/). Note that to use the [styles API](https://material-ui.com/styles/basics/) you can `import @material-ui/core/styles`. We prefer the Hooks styles API.
-- **Routing** Uses [React Router](https://reacttraining.com/react-router/web/guides/quick-start).
-- **State Management** Uses [Redux](https://redux.js.org/introduction/getting-started) and the [Redux Toolkit](https://redux-toolkit.js.org/api/configureStore) with slices.
-- **Testing** Uses [Jest](https://jestjs.io/) with [Enzyme](https://enzymejs.github.io/enzyme/)
+* [Node.js](https://nodejs.org/) = *14.0.0*
+* [yarn](https://yarnpkg.com/lang/en/) >= *1.22.4*
```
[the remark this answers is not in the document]
tleraj: `yarn` added.

website/package.json
```diff
@@ -15,20 +15,27 @@
...
     "lodash": "^4.17.15",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-leaflet": "^2.7.0",
     "react-redux": "^7.2.0",
     "react-router-dom": "^5.1.2",
     "react-scripts": "3.4.1",
     "redux": "^4.0.5",
+    "redux-actions": "^2.6.5",
+    "redux-thunk": "^2.3.0",
```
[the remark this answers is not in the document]
tleraj: Packages deleted.

website/src/helpers/requests.ts
tleraj: Updated `send` to receive a type parameter `T`.

packages/website/src/layout/App/__snapshots__/index.test.tsx.snap
```diff
@@ -2,13 +2,79 @@
...
+            <span
+              class="MuiButton-label"
+            >
+              See Reefs List
+            </span>
+            <span
+              class="MuiTouchRipple-root"
+            />
+          </a>
+        </div>
```
bencpeters, line 75: Let's mock this component for snapshots (I think mocking `Route` should be good - you might also need mocks for the arguments to `component` in each of the routes defined - you don't want the full component to render in this snapshot). We don't want _this_ snapshot to change every time the sub routes change.

packages/website/src/layout/App/theme.ts
```diff
@@ -1,3 +1,78 @@
...
+      secondary: darkGreyBlue,
+    },
+    grey: {
+      500: lightGray,
+    },
+  },
+  overrides: {
+    MuiAppBar: {
+      root: {
+        height: "122px",
```
bencpeters, line 28: This doesn't matter either way, but just a FYI - for material UI's JSS style definitions if you want to specify values in pixels, you can also just use a numeric value.

packages/website/src/routes/ReefRoutes/Reef/ReefNavBar/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <Grid item>
+            <Grid alignItems="center" container spacing={2}>
+              <Grid item>
+                <IconButton
+                  edge="start"
+                  color="inherit"
+                  aria-label="menu"
+                  onClick={() => history.push("/reefs")}
+                >
+                  <ArrowBack />
```
bencpeters, line 35: Can we just use the [`Link`](https://reacttraining.com/react-router/web/api/Link) component for this?

packages/website/src/routes/ReefRoutes/Reef/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+} from "@material-ui/core";
+import { useSelector, useDispatch } from "react-redux";
+import { RouteComponentProps } from "react-router-dom";
+import { Map, TileLayer } from "react-leaflet";
+
+import ReefNavBar from "./ReefNavBar";
+import { reefDetailsSelector, reefRequest } from "../../../store/Reefs/slice";
+
+const Reef = ({ match, classes }: ReefProps) => {
+  const reefDetails = useSelector(reefDetailsSelector);
```
bencpeters, line 18: This selector almost certainly wants an `id`, no?

packages/website/src/routes/ReefRoutes/Reef/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(reefRequest(reefId));
+  }, [dispatch, reefId]);
+
+  return (
+    <>
+      <ReefNavBar
+        reefName={reefDetails.region_name}
```
bencpeters, line 29: I think these properties are going to be `camelCase` (e.g. `regionName`) as soon as you pull in changes from the latest `master`.

packages/website/src/routes/ReefRoutes/Reef/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+              <div className={classes.container}>
+                <Map
+                  dragging={false}
+                  minZoom={6}
+                  className={classes.map}
+                  center={[37.773, -122.43]}
+                  zoom={6}
+                >
+                  <TileLayer
+                    // attribution="Tiles &copy; Esri &mdash; Source: Esri, i-cubed, USDA, USGS, AEX, GeoEye, Getmapping, Aerogrid, IGN, IGP, UPR-EGP, and the GIS User Community"
```
bencpeters, line 47: Is there a reason this is commented out? Generally speaking if we have commented out lines, give an explanation for why they're commented out instead of removed.

packages/website/src/routes/ReefRoutes/index.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import { Switch, Route } from "react-router-dom";
+
+import Reef from "./Reef";
+import ReefsList from "./ReefsList";
+
+const ReefRoutes = () => {
+  return (
```
bencpeters, line 8: If you're using ES6 style arrow functions, you can just do:

```tsx
const ReefRoutes = () => (
  <Switch>
    ...
  </Switch>
);
```

(single line implicit return)

packages/website/src/routes/ReefRoutes/ReefsList/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const history = useHistory();
+
+  return (
+    <>
+      <div className={classes.root}>
+        <List component="nav">
+          <ListItem button>
+            <ListItemText
+              primary="Reef 1"
+              onClick={() => {
```
bencpeters, line 23: Let's use React Router's `Link` component instead.

packages/website/src/store/Reefs/slice.ts
```diff
@@ -0,0 +1,80 @@
...
+import reefServices from "../../services/reefServices";
+
+const selectedReefInitialState: ReefState = {
+  details: {
+    id: "",
+    region_name: "",
+    manager_name: "",
+  },
+  loading: false,
+  error: "",
```
bencpeters, line 15: Consider nulling out `error` instead of having an empty string?

packages/website/src/store/Reefs/slice.ts
```diff
@@ -0,0 +1,80 @@
...
+          error: action.payload,
+          loading: false,
+        };
+      }
+    );
+
+    builder.addCase(reefRequest.pending, (state) => {
+      return {
+        ...state,
+        loading: true,
```
bencpeters, line 62: This reducer should probably reset the `error` state too - any errors from previous requests aren't relevant after we make a new request.

packages/website/src/store/Reefs/slice.ts
```diff
@@ -0,0 +1,80 @@
...
+      };
+    });
+  },
+});
+
+export const reefDetailsSelector = (state: RootState): ReefState["details"] => {
```
bencpeters, line 68: Let's use one-line implicit return format for these ES6 arrow functions

```tsx
 export const reefDetailsSelector = (state: RootState): ReefState["details"] => state.selectedReef.details;
```

packages/website/src/store/configure.ts
```diff
@@ -1,13 +1,22 @@
...
+
+import reducer from "./reducer";
+
+export const store = configureStore({ reducer });
+
+export type RootState = ReturnType<typeof store.getState>;
+
+export type AppDispatch = typeof store.dispatch;
+
+export type AppThunk<ReturnType = void> = ThunkAction<
```
bencpeters, line 11: Is this used anywhere?

packages/website/src/routes/ReefRoutes/Reef/index.tsx
tleraj: I was thinking that the application's `RootState` could look like this  
```
{
  ... ,
  selectedReef: {...},
  reefs: [...],
  ...
}
```
When the user visits the url `/reefs/:id`, the action `reefRequest(id)` will be dispatched and set the `selectedReef` equal to the reef with `id`. So, there is going to be only one selected reef, and the `reefDetailsSelector` will fetch its details. Perhaps I could rename the selector to `selectedReefDetails` so that its purpose is more clear?

packages/website/src/routes/Homepage/index.test.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import { BrowserRouter as Router } from "react-router-dom";
+
+import Homepage from ".";
+
+test("renders as expected", () => {
+  const { container } = render(
+    <Router>
```
bencpeters, line 9: I don't _think_ you need `Router` providers for rendering these snapshots?

packages/website/src/routes/ReefRoutes/Reef/index.tsx
bencpeters: No, that sounds good! I actually meant to delete this comment once I read further down and saw the full slice definition. This looks good to me, but I think this component should have a `loading` guard for when the reef details are loading (and perhaps error handling?)

packages/website/src/routes/ReefRoutes/Reef/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders as expected 1`] = `<div />`;
```
bencpeters, line 3: This snapshot should render the Reef detail page - it doesn't really do us any good to just have a mocked `div` ;-)

packages/website/src/routes/ReefRoutes/Reef/index.test.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { render } from "@testing-library/react";
+import { BrowserRouter as Router, Route } from "react-router-dom";
+
+import Reef from ".";
+
+test("renders as expected", () => {
+  const { container } = render(
+    <Router>
+      <Route exact path="/reefs/1" component={Reef} />
+    </Router>
```
bencpeters, line 11: You don't need to use the router at all in these tests - just render something like this:

```jsx
<Reef match={mockMatch} />
```

You can then construct the `mockMatch` to do whatever you want.

However, you probably need to mock `reefDetailsSelector` to provide mock data to display properly.

packages/website/src/routes/Homepage/index.test.tsx
tleraj: This component (as well ass `ReefNavBar` and `ReefsList`) has a react router `Link` element in it. When I render it without `Router` and execute `yarn run test` in order to create snapshots, it gives me the following error  
```
Invariant failed: You should not use <Link> outside a <Router>
```
That's why I wrapped it in a `Router`.

packages/website/src/routes/ReefRoutes/Reef/index.test.tsx
```diff
@@ -1,14 +1,50 @@
+/* eslint-disable fp/no-mutation */
```
bencpeters, line 1: Not part of this PR, but we probably should add an override to the `eslintrc` to disable this for all tests...

PR #21, opened 2020-06-08
Organized code into monorepo

bencpeters:
# Organized code into monorepo

This PR organizes the codebase into a mono repo managed by [Lerna](https://github.com/lerna/lerna) using [Yarn Workspaces](https://classic.yarnpkg.com/en/docs/workspaces/).

API is now in `packages/api`, web app is in `packages/website`.

ESLint & TS config are (mostly) controlled from the root `.estlintrc` and `tsconfig.json` files. If a sub repo wants to override any part of this config, it can do so with nested options (e.g. see the custom `packages/api/.eslintrc` that overrides a few options for some common Nest.js paradigms).

Set up Husky & `lint-staged` to automatically run pre-commit hooks.

**Important** if you have installed the website node packages from before the mono-repo, you will likely need to delete `.git/hooks/husky*` and then re-install `yarn install` from the root for the precommit hooks to work correctly.

You should make sure to rebuild your node modules with this refactor so that hoisting works properly. Make sure to delete `node_modules` in each of the sub repos and run `yarn install`.


ericboucher:
Looks good to me. Although you have unearthed a series of:

src/reefs/interfaces/reefs.dto.ts:12:14 - error TS2564: Property 'updated_at' has no initializer and is not definitely assigned in the constructor.
```  readonly updated_at: Date; ```

PR #22, opened 2020-06-09
Updates to API TypeORM configuration and Reef entity model

bencpeters:
# Updates to API TypeORM configuration and Reef entity model

This PR cleans up how the TypeORM connection is configured & updates the model Reef connection object/API.

Configuration Updates:
- Moved TypeORM config to [ormconfig.ts](./ormconfig.ts).
- Added Postgres connection url config option `DATABASE_URL` (the older individual `POSTGRES_X` options work as well)
- Added automatic translation of `camelCase` -> `snake_case` for entity field definitions.

Reef Updates
- Removed custom repository to simplify the config logic (for now we can just do the operations directly in the controller, if we ultimately want more complicated logic we can consider either an ActiveRecord pattern or wrapping  logic into a Service).
- Added autogeneration for the date columns
- Changed entity names to `camelCase`
- Removed typescript interface (we should just be able to use the classes themselves as interfaces).


packages/api/ormconfig.ts
```diff
@@ -0,0 +1,54 @@
+import { SnakeNamingStrategy } from 'typeorm-naming-strategies';
+import { join } from 'path';
+// const { join } = require('path');
```
ericboucher, line 3: 👣 ?

PR #28, opened 2020-06-18
Add CRUD endpoints for regions

vickykont-programize:
This PR adds basic CRUD functionality for the `Region` entity.

Also, adds a custom validation constraint to support validating foreign-key constraints. 

Suggestions for further improvements:
- Add validations for geospatial data.
- Transform geospatial data on fetch.


packages/api/src/app.module.ts
```diff
@@ -2,12 +2,14 @@ import { Module } from '@nestjs/common';
...
 import { AppController } from './app.controller';
 import { ReefsModule } from './reefs/reefs.module';
+import { RegionsModule } from './regions/regions.module';
 import { configService } from './config/config.service';
 
 @Module({
   imports: [
     TypeOrmModule.forRoot(configService.getTypeOrmConfig()),
     ReefsModule,
+    RegionsModule,
```
ericboucher, line 12: I had to remove this import in #31 as it was breaking `start:prod`. Maybe we need to concert the order and have `TypeOrmModule.forRoot(configService.getTypeOrmConfig()),` come last?

Make sure that you test `yarn build && yarn start:prod`

packages/api/src/regions/regions.entity.ts
```diff
@@ -21,9 +21,9 @@ export class Region {
   @Index({ spatial: true })
   polygon: string;
 
-  @ManyToOne(() => Region, { onDelete: 'CASCADE' })
+  @ManyToOne(() => Region, { onDelete: 'CASCADE', nullable: true })
   @JoinColumn({ name: 'parent_id' })
-  parentId: Region;
+  parentId: Region | number;
```
bencpeters, line 26: Let's change this to `parent?: Region` (to make it clear that it's returning the parent entity, not the parent id, and that the parent is optional)

packages/api/src/regions/regions.repository.ts
```diff
@@ -0,0 +1,20 @@
+import { EntityRepository, Repository } from 'typeorm';
+import { Region } from './regions.entity';
+import { FilterRegionDto } from './dto/filter-region.dto';
+
+@EntityRepository(Region)
+export class RegionsRepository extends Repository<Region> {
```
bencpeters, line 6: I think you _should_ be able to avoid needing to define both a `service` _and_ a `repository` for each class by using default Nest injections.

Module:

```typescript
// Inject the actual entity here:
imports: [TypeOrmModule.forFeature([Region]),
```

Service:

```typescript
  constructor(
    @InjectRepository(Region)
    private repo: Repository<Region>,
  ) {}
```

Then, implement the query logic in the appropriate service methods using `this.repo`.

I think in most cases it'll just result in slightly more consolidated code by keeping business logic mostly in the service rather than splitting between 2 different classes.

packages/api/src/regions/regions.repository.ts
```diff
@@ -0,0 +1,20 @@
...
+  async filter(filter: FilterRegionDto): Promise<Region[]> {
+    const query = this.createQueryBuilder('region');
+    if (filter.name) {
+      query.andWhere('(region.name LIKE :name)', { name: `%${filter.name}%` });
```
bencpeters, line 10: Do we want to make this case insensitive?

PR #29, opened 2020-06-19
Reef Detail Page Temperature Data

tleraj:
This pull request concerns the temperature data on the reef detail page. Here is a picture of what it looks like.
reef-detail-page


ericboucher:
@tleraj an you rebase your PR on master or merge master for clarity?

packages/website/src/routes/ReefRoutes/Reef/Temperature/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+        title={
+          <Typography
+            style={{ display: "flex", alignItems: "baseline" }}
+            component="div"
+          >
+            <Box>
+              <Typography variant="h6">CURRENT CONDITIONS</Typography>
+            </Box>
+            <Box ml={1}>
+              <Typography variant="subtitle2">(05/12/20 8:16AM PST)</Typography>
```
bencpeters, line 45: Let's either not hardcode this, or add a `TODO` comment to update this?

packages/website/src/routes/ReefRoutes/Reef/Temperature/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  return (
+    <Card className={classes.card}>
+      <CardHeader
+        className={classes.header}
+        title={
+          <Typography
+            style={{ display: "flex", alignItems: "baseline" }}
+            component="div"
+          >
```
bencpeters, line 40: Why is this wrapper a `Typography` component? It seems like `Grid` would be a better fit for this than `Typography` wrapping `Box`?

packages/website/src/routes/ReefRoutes/Reef/Temperature/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+              direction="column"
+              alignItems="flex-start"
+            >
+              <Alert
+                className={classes.alert}
+                variant="filled"
+                severity="warning"
+                icon={<ErrorIcon />}
+              >
+                <Typography variant="caption">ALERT LEVEL: HIGH</Typography>
```
bencpeters, line 141: Should this be conditional?

packages/website/src/routes/ReefRoutes/Reef/index.test.tsx
```diff
@@ -26,6 +26,7 @@ describe("Reef Detail Page", () => {
             type: "",
             coordinates: [[[0, 0]]],
           },
+          dailyData: [],
```
bencpeters, line 29: Maybe have a snapshot with no data & one with data?

packages/website/src/store/Reefs/types.ts
```diff
@@ -6,12 +6,40 @@ interface Polygon {
   type: string;
 }
 
+interface Data {
```
bencpeters, line 9: This is probably a useful interface to export...

Do we expect to guarantee that all of these values exist on every weather observation? Or should some properties be optional?

packages/website/src/routes/ReefRoutes/Reef/Map/index.tsx
```diff
@@ -1,21 +1,27 @@
...
-    <Map
-      dragging={false}
-      className={classes.map}
-      bounds={mapBounds(polygon.coordinates)}
-      boundsOptions={{ padding: [150, 150] }}
-      zoomControl={false}
-      doubleClickZoom={false}
-      scrollWheelZoom={false}
-    >
+    <Map ref={mapRef} dragging={false} className={classes.map}>
```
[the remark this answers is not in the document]
tleraj: I want the map's initial position to fit the reef's polygon and also disable any zoom out of the initial frame. In order to do so, I had to expose the `Map`'s `fitBounds` and `setMinZoom` functions. That's why I used the `ref` + `useEffect` approach.

packages/website/src/routes/ReefRoutes/Reef/Stats/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+    return colorCode[index];
+  }
+  return colorCode[len - 1];
+};
+
+const Stats = ({ dailyData, classes }: StatsProps) => {
+  const degreeHetingDays =
+    dailyData && dailyData[0] && dailyData[0].degreeHeatingDays;
+
+  const colorItem = degreeHetingDays && colorFinder(degreeHetingDays);
```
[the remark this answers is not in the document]
tleraj: The `colorItem`is defined if and only if the `degreeHeatingDays` is defined. In any other case, the degree heating days value and the color indicator on the bottom are not displayed on the UI (there is a conditional rendering for those) and looks like this:

Screenshot from 2020-06-29 20-22-02

packages/website/src/routes/ReefRoutes/Reef/Temperature/index.tsx
tleraj: I totally agree. I just didn't know what condition to use (e.g. `severity="warning"` if `temperature > someValue`), so I left it like this. I believe a `switch-case` approach would be optimal in order to manage all possible temperature scales.

ericboucher: @tleraj let's add a TODO and open an issue to make sure that we don't forget this

packages/website/src/routes/ReefRoutes/Reef/Stats/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+  }
+  if (index === 0) {
+    return colorCode[index];
+  }
+  return colorCode[len - 1];
+};
+
+const Stats = ({ dailyData, classes }: StatsProps) => {
+  const degreeHetingDays =
+    dailyData && dailyData[0] && dailyData[0].degreeHeatingDays;
```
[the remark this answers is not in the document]
tleraj: Haven't added the "No data found" message yet. I first want to make sure that this is what you meant with top-level handler. Also for the non existent daily data case, the Reef Detail page looked like this:

no-data

If there are no daily data, we want a user friendly message for the whole screen (i.e. none of the 5 cards render) or maybe a no data found message for each card (e.g. "Temperature Data Not Found", "Degree Heating Weeks Not Found", etc)?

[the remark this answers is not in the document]
tleraj: I added a no data message that looks like this:

no-data

I also added a loading progress bar below the navbar.

PR #33, opened 2020-06-24
Enable postgis extension in postgres docker container

vickykont-programize:
This PR updates the definition of the docker container that runs Postgres, so that the `postgis` package is installed and the `postgis` extension is enabled.


packages/api/docker-compose.yml
```diff
@@ -1,11 +1,18 @@
 version: '3.4'
 services:
   aqua-postgres:
-    image: postgres:12.2
+    image: postgres:12.2-postgis
```
ericboucher, line 4: Let's use a pre-made postgis image directly here

eg. kartoza/postgis:12.1 / https://hub.docker.com/r/kartoza/postgis

packages/api/postgres-postgis-init.sql
```diff
@@ -0,0 +1 @@
+CREATE EXTENSION postgis;
```
ericboucher, line 1: Let's make it easier to run on the external databses and make sure to use the `IF NOT EXISTS` syntax here.

```
CREATE EXTENSION IF NOT EXISTS POSTGIS
```

packages/api/docker-compose.yml
```diff
@@ -1,11 +1,14 @@
 version: '3.4'
 services:
   aqua-postgres:
-    image: postgres:12.2
+    image: kartoza/postgis:12.1
     restart: always
     volumes:
-      - ./db_data:/var/lib/postgresql/data
+      - db_data:/var/lib/postgresql
```
ericboucher, line 7: @vickykont-programize why this change?

ericboucher: Was it because of the build context change? Is it ok to revert?

PR #39, opened 2020-06-29
Add charts to reef detail page

tleraj:
This PR concerns the charts on the reef detail page. Before making this PR ready for review, these tasks must be completed
- [x] Find a more stable approach to implement the orange shading (currently not responsive)
- [x] Customize the graph tooltip
- [x] Add the background pattern on the graphs
- [x] Add bleaching threshold annotation

Right now the graphs look like this:

Screenshot from 2020-06-29 19-53-32


ericboucher:
For the background pattern, maybe we can get some inspiration from https://github.com/datavisyn/chartjs-plugin-barchart-background

tleraj:
The addition of the background pattern made the graph look like this:

Screenshot from 2020-06-30 12-46-23

tleraj:
The addition of the tooltip made the graph look like this:

Screenshot from 2020-07-02 14-01-36

Still some work for the tooltip to get the data displayed dynamically.

tleraj:
Added vertical chart slice. It looks like this: 

Screenshot from 2020-07-03 18-07-00

tleraj:
Added bleaching threshold annotation text. Also the tooltip data now comes from the chart data points.

Screenshot from 2020-07-06 17-48-50

ericboucher:
Nice start!

- Let's try to make it a bit more responsive, most of the objects don't look "right" if you are not exactly at the right screen size.

- We should have some mock data available, let's discuss a strategy to make this happen

packages/website/src/routes/ReefRoutes/Reef/Charts/tooltip.tsx
```diff
@@ -0,0 +1,157 @@
...
+  Card,
+  CardHeader,
+  CardContent,
+  Grid,
+  Box,
+  Typography,
+} from "@material-ui/core";
+
+const Tooltip = ({ data, classes }: TooltipProps) => {
+  const date = new Date(data.date).toLocaleString().replace(/:\d{2}\s/, " ");
```
bencpeters, line 15: Can we add a quick comment on what this is doing? It looks like it's removing the seconds from the date string?

packages/website/src/routes/ReefRoutes/Reef/Charts/tooltip.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const splitDate = date.split(", ");
+
+  return (
+    <Card className={classes.tooltip}>
+      <CardHeader
+        style={{ flex: "0 1 auto", padding: "0.5rem 2rem 0.5rem 2rem" }}
```
bencpeters, line 21: Is there a reason this is done using `style` instead of CSS? It seems like we could re-use the same padding class for this and the `CardContent` below?

packages/website/src/routes/ReefRoutes/Reef/Charts/tooltip.tsx
```diff
@@ -0,0 +1,157 @@
...
+              <Typography variant="caption">SURFACE TEMP</Typography>
+              <Typography variant="h5">
+                {data.surfaceTemperature} &#8451;
+              </Typography>
+            </Grid>
+          </Grid>
+          <Grid container direction="column" item xs={12}>
+            <Typography variant="caption">WIND</Typography>
+            <Typography
+              style={{ display: "flex", alignItems: "baseline" }}
```
bencpeters, line 76: This should be CSS too...

packages/website/src/routes/ReefRoutes/Reef/Charts/tooltip.tsx
```diff
@@ -0,0 +1,157 @@
...
+      height: 230,
+      width: 240,
+      display: "flex",
+      flexFlow: "column",
+      backgroundColor: "#404b6b",
+      opacity: 0.9,
+    },
+  });
+
+interface TooltipData {
```
bencpeters, line 142: Let's use this type to initialize `useState<TooltipData | null>` (or something like that) in the `Charts` component?

packages/website/src/routes/ReefRoutes/Reef/ReefNavBar/index.tsx
```diff
@@ -59,12 +68,21 @@ const ReefNavBar = ({ reefName, lastSurvey, managerName }: ReefNavBarProps) => (
   </AppBar>
 );
 
+const styles = (theme: Theme) =>
+  createStyles({
+    appBar: {
+      "&.MuiPaper-root": {
```
bencpeters, line 74: This may well not work on a prod build - the JSS classes get renamed (so `.MuiPaper-root` might well not exist). Is there a reason you can just add this style directly? `<AppBar classes={{ root: classes.appBar }} />`?

packages/website/src/routes/ReefRoutes/Reef/Temperature/index.tsx
```diff
@@ -15,12 +15,14 @@ import Alert from "@material-ui/lab/Alert";
...
-  }
-
-  const bottomTemperature = dailyData[0].maxBottomTemperature;
-  const date = new Date(dailyData[0].date);
+  const sortByDate = Object.values(dailyData).sort((item1, item2) => {
+    if (item1.date > item2.date) {
+      return -1;
+    }
+    return 1;
+  });
```
bencpeters, line 23: This same logic is getting repeated a lot in different files - extract to a utility function?

packages/website/src/routes/ReefRoutes/Reef/ReefNavBar/index.tsx
tleraj: For some unknown reason, if I don't use this, the AppBar's color turns to white. I also tried the solution you suggested, but still doesn't work.

PR #40, opened 2020-06-29
Add CRUD endpoints for reefs

vickykont-programize:
This PR adds CRUD endpoints for reefs.

### DIFF
--- a/packages/api/src/reefs/reefs.service.ts
+++ b/packages/api/src/reefs/reefs.service.ts
@@ -1,0 +1,69 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+import { Reef } from './reefs.entity';
+import { CreateReefDto } from './dto/create-reef.dto';
+import { FilterReefDto } from './dto/filter-reef.dto';
+import { UpdateReefDto } from './dto/update-reef.dto';
+
+@Injectable()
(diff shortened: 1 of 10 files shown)


packages/api/src/reefs/reefs.service.ts
```diff
@@ -0,0 +1,69 @@
...
+      relations: ['region', 'admin', 'stream'],
+    });
+    if (!found) {
+      throw new NotFoundException(`Reef with ID ${id} not found.`);
+    }
+    return found;
+  }
+
+  async update(id: number, updateReefDto: UpdateReefDto) {
+    const result = await this.reefsRepository.update(id, updateReefDto);
```
bencpeters, line 57: Let's return the result - CRUD paths generally want to return the updated entity.

bencpeters:
@vickykont-programize Let's just make sure the tests pass before merging - it looks like the Reefs controller needs a few more dependencies mocked before it gets injected.

packages/api/src/reefs/reefs.service.ts
vickykont-programize: I looked into TypeORM's documentation and it provides 2 functions to handle updates, `save()` and `update()`. `save()` only returns some partial data for the updated entity and `update()` only some info about the query.

Should I add a second query to return the updated entity?

bencpeters: Ya, I think it probably makes the most sense to add a second query for the updated entity.

PR #44, opened 2020-07-01
Add ability to getMMM (maximum monthly mean) in the backend

ericboucher:
Initially, we thouhght about loading the raster into Postgis. Sadly, `rasters` are not supported under Google cloud managed Postgres at the moment.

So I created a small script to transform the netcdf into a cloud optimized geotiff, and loaded it on: https://storage.googleapis.com/reef_climatology/sst_clim_mmm.tiff

And a simple Typescript function to get the value for a reef location (lat, on).


packages/api/src/utils/coordinates.ts
```diff
@@ -0,0 +1,18 @@
+export type Extent = [number, number, number, number];
+
+function pointToPixel(
+  lat: number, long: number, boundingBox: Extent, height: number, width: number
+) {
+  lat3 = lat < 0 ? lat + 180 : lat
+  long3 = long < 0 ? long + 360 : long
```
drewjgray: Have we converged on a standard for longitude? Most mapping systems use -180 to +180. While NOAA weather data uses 0 to 360. It looks like you took the NOAA convention here, is this consistent with the rest of the app?

packages/api/scripts/load_raster.sh
```diff
@@ -0,0 +1,7 @@
+# Download NOAA Climatology NETCDF file
+wget -nc ftp://ftp.star.nesdis.noaa.gov/pub/sod/mecb/crw/data/5km/v3.1/climatology/nc/ct5km_climatology_v3.1.nc
+
+# Transform into a Cloud Optimized Geotiff - https://www.cogeo.org/
+gdal_translate NETCDF:"./ct5km_climatology_v3.1.nc":sst_clim_mmm sst_clim_mmm.tiff -co COMPRESS=LZW -co TILED=YES
```
drewjgray, line 5: Maybe pull out just the single MMM band so the COG is smaller?

packages/api/src/utils/coordinates.ts
```diff
@@ -0,0 +1,23 @@
+export type Extent = [number, number, number, number];
+
+export function pointToPixel(
```
drewjgray, line 3: Why do we need to get the pixel value? Can't we use lat/lon directly as the file is a geotiff?

packages/api/scripts/load_raster.sh
ericboucher: Yup, that what the `sst_clim_mmm` is doing :)

packages/api/src/utils/coordinates.ts
ericboucher: Sadly I did not find a way to do that efficiently in javascript...

PR #46, opened 2020-07-02
Clean up migrations and fix down migrations

vickykont-programize:
This PR removes all redundant drops/re-creations of indices from TypeORM's migrations, while keeping the structure and names of the existing migrations for compatibility reasons.

Also, some issues were found while testing the down migrations. This PR also fixes these issues.


packages/api/migration/1592510640883-InitDB.ts
```diff
@@ -4,6 +4,7 @@ export class InitDB1592510640883 implements MigrationInterface {
     name = 'InitDB1592510640883'
 
     public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE EXTENSION IF NOT EXISTS "postgis"`);
         await queryRunner.query(`CREATE TABLE "region" ("id" SERIAL NOT NULL, "name" character varying(50) NOT NULL, "polygon" polygon NOT NULL, "created_at" TIMESTAMP NOT NULL DEFAULT now(), "updated_at" TIMESTAMP NOT NULL DEFAULT now(), "parent_id" integer, CONSTRAINT "PK_5f48ffc3af96bc486f5f3f3a6da" PRIMARY KEY ("id"))`);
```
ericboucher, line 8: I'm not sure we should put arbitrary limits in character length at this point. Maybe we could just use `text` for all these columns?

bencpeters: I agree with this. I think unless there's a strong reason to limit characters we shouldn't do it, especially at a DB level.

Postgres actually handles `varchar` (with no character limit) and `text` identically internally, so it doesn't really matter which we use (if I was writing the migrations by hand, I'd certainly use `text`, but I don't know that it's worth spending time figuring out how to make TypeORM do that by default).

ericboucher:
I investigated this TypeORM issue a bit and it seems to be a known one. Eg:

- https://github.com/typeorm/typeorm/issues/5117
- https://github.com/typeorm/typeorm/issues/1686

`
no, its okay that typeorm drops and re-create exist keys. To make a proper comparison of exist keys TypeORM uses its own naming strategy for them.
So it simply checks that it does not have fks and there are outdated fks - it removes old one a create new ones. Its expected behaviour.
`

Not sure I agree with the "its okay" part here though haha

PR #47, opened 2020-07-07
Connect the Frontend to the API

ericboucher:
After adding #51 seeding to the API we can now connect reef and daily data endpoints with the frontend


packages/api/src/reefs/reefs.entity.ts
```diff
@@ -19,9 +20,14 @@ export class Reef {
   @Column({ nullable: true })
   name: string;
 
-  @Column('geometry', { unique: true })
+  @Column({
+    type: 'geometry',
+    nullable: true,
```
vickykont-programize, line 25: I think we need to generate a migration.

packages/api/package.json
```diff
@@ -32,24 +32,30 @@
     "hashId:staging": "DOTENV_CONFIG_PATH=.env.staging ts-node -r dotenv/config scripts/url-hash.ts toHash",
     "unhashId": "ts-node -r dotenv/config scripts/url-hash.ts toId",
     "unhashId:prod": "DOTENV_CONFIG_PATH=.env.prod ts-node -r dotenv/config scripts/url-hash.ts toId",
-    "unhashId:staging": "DOTENV_CONFIG_PATH=.env.staging ts-node -r dotenv/config scripts/url-hash.ts toId"
+    "unhashId:staging": "DOTENV_CONFIG_PATH=.env.staging ts-node -r dotenv/config scripts/url-hash.ts toId",
+    "seed:config": "ts-node ../../node_modules/typeorm-seeding/dist/cli.js config",
+    "seed:run": "ts-node ../../node_modules/typeorm-seeding/dist/cli.js seed"
```
vickykont-programize, line 37: I think we could avoid defining specific paths by doing this.
```suggestion
    "typeorm-seeding": "ts-node $(yarn bin typeorm-seeding)",
    "seed:config": "yarn typeorm-seeding config",
    "seed:run": "yarn typeorm-seeding seed",
```

packages/api/package.json
```diff
@@ -32,24 +32,31 @@
...
+    "typeorm-seeding": "ts-node $(yarn bin typeorm-seeding)",
+    "seed:config": "yarn typeorm-seeding config",
+    "seed:run": "yarn typeorm-seeding seed"
   },
   "dependencies": {
     "@nestjs/common": "^7.0.0",
     "@nestjs/core": "^7.0.0",
     "@nestjs/platform-express": "^7.0.0",
     "@nestjs/typeorm": "^7.0.0",
+    "@types/faker": "^4.1.12",
```
bencpeters, line 45: Should be a dev dependency

packages/api/package.json
```diff
@@ -32,24 +32,31 @@
...
     "@types/lodash": "^4.14.156",
     "class-transformer": "^0.2.3",
     "class-validator": "^0.12.2",
+    "faker": "^4.1.0",
```
bencpeters, line 49: _Probably_ should be a dev dependency too...

packages/api/package.json
```diff
@@ -32,24 +32,31 @@
...
     "hashids": "^2.2.1",
     "lodash": "^4.17.15",
     "objects-to-csv": "^1.3.6",
     "pg": "^8.2.1",
     "reflect-metadata": "^0.1.13",
     "rxjs": "^6.5.4",
     "typeorm": "^0.2.25",
-    "typeorm-naming-strategies": "^1.1.0"
+    "typeorm-naming-strategies": "^1.1.0",
+    "typeorm-seeding": "^1.6.1"
```
bencpeters, line 59: This also probably should be a dev dependency, unless we have issues loading the TypeORM config without it in prod... but generally we want to keep the prod build as small as possible by avoiding adding all these utilities.

packages/api/src/database/create-reefs.seeds.ts
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable no-param-reassign, fp/no-mutation, class-methods-use-this */
```
bencpeters, line 1: If these types of disables are going to be commonplace for seeds, let's add an `override` for `*.seeds.ts` in the eslintrc instead of doing it file-by-file...

packages/api/src/database/create-reefs.seeds.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Factory, Seeder } from 'typeorm-seeding';
+import { Reef } from '../reefs/reefs.entity';
+import { DailyData } from '../reefs/daily-data.entity';
+
+export class CreateReef implements Seeder {
+  public async run(factory: Factory): Promise<void> {
```
bencpeters, line 7: Just a tip - you don't _have_ to specify a return type if you're not going to use it...

packages/api/src/database/create-reefs.seeds.ts
```diff
@@ -0,0 +1,33 @@
...
+    const reef = await factory(Reef)()
+      .map(async (r) => {
+        return r;
+      })
```
bencpeters, line 11: What are you trying to do here? It looks like a noop (well, perhaps wrapping in an unnecessary promise, but that shouldn't change anything)

packages/api/src/database/reefs.factory.ts
```diff
@@ -0,0 +1,50 @@
...
+    type: 'Point',
+    coordinates: [lng, lat],
+  };
+  reef.temperatureThreshold = faker.random.number({ min: 25, max: 27 });
+  reef.depth = faker.random.number({ min: 10, max: 40 });
+  reef.status = faker.random.number(1);
+  return reef;
+});
+
+define(DailyData, (faker: typeof Faker) => {
```
bencpeters, line 23: This should probably be defined in a separate file (`daily-data.factory.ts`) - it really should mimic the actual entity definitions.

packages/api/src/reefs/reefs.entity.ts
```diff
@@ -19,9 +20,13 @@ export class Reef {
...
 
-  @Column('geometry', { unique: true })
+  @Column({
+    type: 'geometry',
+    unique: true,
+    srid: 4326,
+  })
   @Index({ spatial: true })
-  polygon: string;
+  polygon: GeoJSON;
```
bencpeters, line 29: Open question - is this the best type for this? It seems like what we're really using this for is the location of the reef and we definitely have front-end code that expects that to be a single point. Does `geojson` have a point type we could use instead here?

packages/website/src/routes/ReefRoutes/Reef/ReefNavBar/index.tsx
```diff
@@ -59,10 +68,21 @@ const ReefNavBar = ({ reefName, lastSurvey, managerName }: ReefNavBarProps) => (
   </AppBar>
 );
 
-interface ReefNavBarProps {
+const styles = (theme: Theme) =>
+  createStyles({
+    root: {
+      "&.MuiPaper-root": {
```
bencpeters, line 74: What are you trying to do here? This class will almost certainly have a different name when compiled in production, I doubt this will work... Is this supposed to be a conditional background color applied to this element? Or trying to override a default Mui style?

A better approach would almost certainly be to override this at a theme level if it's the latter...

packages/website/src/routes/ReefRoutes/Reef/index.tsx
```diff
@@ -21,15 +21,17 @@ import {
   reefLoadingSelector,
   reefErrorSelector,
   reefRequest,
-} from "../../../store/Reefs/slice";
+} from "../../../store/Reefs/selectedReefSlicelice";
```
bencpeters, line 24: Typo in this file name (either in the actual rename, or in this file)

packages/website/src/routes/ReefRoutes/ReefsList/index.tsx
```diff
@@ -10,22 +11,45 @@ import {
...
+    dispatch(reefsRequest());
+  }, [dispatch]);
+
+  return (
+    <>
+      <div className={classes.root}>
+        <List component="nav">
+          {reefsList.map(
+            (reef) =>
+              reef.id && (
```
bencpeters, line 33: Is this check actually necessary? Is there any situation where we might have a non empty list of reef objects without `id`s?

packages/website/src/store/Reefs/reefsListSlice.ts
```diff
@@ -0,0 +1,79 @@
+import { createSlice, createAsyncThunk, PayloadAction } from "@reduxjs/toolkit";
+import { AxiosError } from "axios";
+
+import { ReefsListState } from "./types";
+
+/* eslint-disable-next-line import/no-cycle */
+import { RootState, CreateAsyncThunkTypes } from "../configure";
```
bencpeters, line 7: use `import type` to try to remove import cycle warning.

packages/website/src/store/Reefs/selectedReefSlicelice.ts
```diff
@@ -1,29 +1,31 @@
 import { createSlice, createAsyncThunk, PayloadAction } from "@reduxjs/toolkit";
```
bencpeters, line 1: Typo in this file name.

packages/website/src/store/Reefs/selectedReefSlicelice.ts
```diff
@@ -1,29 +1,31 @@
...
 
-import { ReefState } from "./types";
+import { SelectedReefState } from "./types";
 /* eslint-disable-next-line import/no-cycle */
 import { RootState, CreateAsyncThunkTypes } from "../configure";
 import reefServices from "../../services/reefServices";
 
-const selectedReefInitialState: ReefState = {
+const selectedReefInitialState: SelectedReefState = {
   details: {
```
bencpeters, line 10: I think this should be nulleable (or `undefined`) as an object, rather than having a "fake" details object with all null fields. This approach is just going to lead to field-by-field checks all over the place & Typescript won't be able to be as helpful.

packages/website/src/store/Reefs/types.ts
```diff
@@ -34,15 +38,26 @@ export interface Data {
...
-  managerName: string;
-  videoStream: string;
-  polygon: Polygon;
+  id: number | null;
+  name: string | null;
+  polygon: Polygon | Point;
+  temperatureThreshold: number | null;
+  depth: number | null;
+  status: number;
+  videoStream: string | null;
```
bencpeters, line 47: Some of these fields don't seem like they're actually nulleable (e.g. `id`). This type should reflect what we actually get from the API/require in the database...

packages/api/src/reefs/reefs.entity.ts
bencpeters: I should clarify this comment after reading the rest of the PR - it looks like we can have either polygons or points here. But I think the general question of the best type might still apply?

PR #51, opened 2020-07-09
Seeding - Backend

ericboucher:
Add seeding functionalities to the backend for reef and daily data


packages/api/.eslintrc
```diff
@@ -32,6 +28,19 @@
           }
         ]
       }
+    },
+    {
+      "files": ["src/database/*"],
```
bencpeters, line 33: This should maybe be a bit more specific: `src/database/*.factory.ts` (or `src/database/*.{factory,seeds}.ts`?). I don't want it to also mess up entity files.

packages/api/src/database/reefs.factory.ts
```diff
@@ -0,0 +1,50 @@
...
+    type: 'Point',
+    coordinates: [lng, lat],
+  };
+  reef.temperatureThreshold = faker.random.number({ min: 25, max: 27 });
+  reef.depth = faker.random.number({ min: 10, max: 40 });
+  reef.status = faker.random.number(1);
+  return reef;
+});
+
+define(DailyData, (faker: typeof Faker) => {
```
bencpeters, line 23: Let's move this into its own file.

packages/api/.eslintrc
ericboucher: There are no other files under database. Maybe I should rename this folder `seeds` ?

PR #55, opened 2020-07-16
Add sofar utils and daily-data workers

ericboucher:
This PR aims to provide the necessary utils and functions to fetch, calculate, and insert daily data for each reef.

This PR:
- [x] replaces the spotter table by a `spotter_id` in the reef table
- [x] makes the daily data columns nullable
- [x] creates basic functions to call sofar APIs
- [x] adding a constraint to prevent (date, reef) duplicates
- [x] Resolve issue with `geo-tz` type
- [x] Add cloud function / cron


ericboucher:
@bencpeters for some reason the type declaration for `geo-tz` that I added to `custom.d.ts` is not working and to run the script I have to turn off `strict` in tsconfig... Any ideas?

Also, what trick would you suggest to prevent duplicate data for (reef, day) pairs?

packages/api/src/reefs/reefs.entity.ts
```diff
@@ -20,6 +20,9 @@ export class Reef {
   @Column({ nullable: true })
   name: string;
 
+  @Column({ nullable: true })
+  spotterId: string;
```
bencpeters, line 24: Is this actually a string, or is this a foreign key to another table?

packages/api/src/utils/sofar.test.ts
```diff
@@ -0,0 +1,22 @@
+import { getSofarDailyData } from './sofar';
+
+test('Test', async () => {
```
bencpeters, line 3: This should be a description of the test, no?

packages/api/src/utils/sofar.ts
```diff
@@ -0,0 +1,88 @@
...
+  variableID: string,
+  latitude: number,
+  longitude: number,
+  date: Date,
+) {
+  // Get day equivalent in timezone using geo-tz to compute "start" and "end".
+  const timezone = get(geoTz(latitude, longitude), 0, '') as string;
+  const m = moment.tz(date, timezone);
+  const start = m.clone().startOf('day').utc().format();
+  const end = m.clone().endOf('day').utc().format();
```
bencpeters, line 49: So it looks like this code parses the date in local time to the location of the reef, then gets the UTC timestamps of the start and end of the day _in local_. So basically when we make these requests, the UTC hours and even date are going to vary a lot based on where the reef is located. Is that what we want? I just wanted to be very explicit about it, since it's easy to get these things confused, especially since it seems like this function is basically assuming `date` is in local time, but most of the logic in this PR is using UTC...

Or does moment parse `date` as UTC, then convert to local?

packages/api/src/utils/sofar.ts
```diff
@@ -0,0 +1,88 @@
...
+  date: Date,
+): Promise<SpotterData> {
+  // TODO - Implement Spotter Data Retrieval
+  // https://docs.sofarocean.com/spotter-sensor
+  // getSofarSpotterData()
+  // using SOFAR_SPOTTER_URL
+  return { surfaceTemperature: 20, bottomTemperature: [0] };
+}
+
+export function getValueClosestToDate(sofarValues: SofarValue[], date: Date) {
```
bencpeters, line 79: Probably worth adding a docstring, and making it very clear that `date` here should be specified in UTC.

packages/api/src/utils/sofar.ts
```diff
@@ -0,0 +1,88 @@
...
+  const index = sofarValues.reduce((r, a, i, aa) => {
+    return i &&
+      Math.abs(new Date(aa[r].timestamp).getTime() - date.getTime()) <
+        Math.abs(new Date(a.timestamp).getTime() - date.getTime())
+      ? r
+      : i;
+  }, -1);
```
bencpeters, line 86: There's nothing _wrong_ with this per se, but I find this code very difficult to read & reason about - I think it'd be much more clearly written (didn't test, but should be basically right):

```typescript
const timeDiff = (timestamp) => Math.abs(new Date(timestamp).getTime() - date.getTime())
sofarValues.reduce((prevClosest, nextPoint) => timeDiff(prevClosest.timestamp) > timeDiff(nextPoint.timestamp)
  ? nextPoint
  : prevClosest
).value;
```

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,161 @@
+/** Worker to process daily data for all reefs. */
+import { sum, get } from 'lodash';
+import { createConnection, Connection } from 'typeorm';
+import { Reef } from '../reefs/reefs.entity';
+import { DailyData } from '../reefs/daily-data.entity';
+import { getSofarDailyData, getSpotterData } from '../utils/sofar';
+// import { calculateDegreeHeatingDays } from '../utils/temperature';
```
bencpeters, line 7: Let's either explain why this is commented out or remove

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,161 @@
...
+const dbConfig = require('../../ormconfig');
+
+async function getDailyData(reef: Reef, date: Date): Promise<any> {
```
bencpeters, line 10: Can we do a better job of typing than this? If the return isn't important, use `Promise<void>`, otherwise let's type the return value (or let Typescript figure it out for us - explicitly typing it `any` gives up all type protections basically)

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,161 @@
...
+  const { polygon, spotterId } = reef;
+  const [longitude, latitude] = get(polygon, 'coordinates');
```
bencpeters, line 12: Why lodash `get` here?

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,161 @@
...
+  // Get Spotter Data
```
bencpeters, line 13: /pedantic, but these are the types of comments I find to be pure noise - I don't think the comment "Get Spotter Data" right above a function call to `getSpotterData` adds a ton of info ;-)

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,161 @@
...
+  const spotterData = await getSpotterData('reef.spotterId', date);
```
bencpeters, line 14: Shouldn't this be `getSpotterData(spotterId, date)`? (right now it's using the string literal `'reef.spotterId'` as the first argument to this function call...)

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export async function getReefsDailyData(connection: Connection, date: Date) {
+  const reefRepository = connection.getRepository(Reef);
+  const dailyDataRepository = connection.getRepository(DailyData);
+  const allReefs = await reefRepository.find();
+  allReefs.forEach(async (reef) => {
+    const dailyDataInput = await getDailyData(reef, date);
+    const dailyDataEntities = dailyDataRepository.create(dailyDataInput);
+    dailyDataRepository.save(dailyDataEntities);
+  });
```
bencpeters, line 143: I feel like this function should probably return `Promise.all(allReefs.map(...))` - even if you don't use the return value right now, it's kinda confusing to have an async function that doesn't actually wait for its operations to be finished to resolve itself...

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,161 @@
...
+
+async function run() {
+  const today = new Date();
+  today.setHours(0, 0, 0, 0);
+
+  const yesterday = new Date(today);
+  yesterday.setDate(today.getDate() - 1);
+  createConnection(dbConfig).then(async (connection) => {
+    try {
+      await getReefsDailyData(connection, yesterday);
```
bencpeters, line 154: This isn't actually waiting for the operation to be finished - see the last comment.

packages/api/ormconfig.ts
```diff
@@ -52,4 +53,4 @@ export = {
     migrationsDir: 'migration',
     subscribersDir: 'subscriber',
   },
-};
+} as unknown) as ConnectionOptions;
```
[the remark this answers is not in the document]
ericboucher: I think it was actually complains about having too many fields because of the `seeds` fields

packages/api/scripts/backfill-daily-data.ts
```diff
@@ -0,0 +1,40 @@
+import { createConnection } from 'typeorm';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import yargs from 'yargs';
+import { getReefsDailyData } from '../src/workers/dailyData';
+
+const dbConfig = require('../ormconfig');
```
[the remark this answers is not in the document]
ericboucher: It needs to be a require for some reason, otherwise it's getting a weird `name is undefined` error

packages/api/scripts/backfill-daily-data.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const { argv } = yargs
+  .scriptName('backfill-data')
+  .usage('$0 <cmd> [args]')
+  .option('d', {
+    alias: 'days',
+    describe: 'Specify how far back we should backfill',
+    type: 'number',
+  })
+  .required('d')
```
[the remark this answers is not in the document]
ericboucher: I'm envisioning this script to get more complex

packages/api/src/reefs/reefs.entity.ts
ericboucher: Just a string that we use to ping the `sofar` API. I got rid of the spotter table since all of the data should be readily available through the Sofar API

packages/api/src/utils/sofar.ts
ericboucher: Added a docstring. Is it worth changing the date name to utcDate you think?

packages/api/package.json
```diff
@@ -35,24 +35,31 @@
...
+    "backfill-data": "ts-node -r dotenv/config scripts/backfill-daily-data.ts"
   },
   "dependencies": {
     "@nestjs/common": "^7.0.0",
     "@nestjs/core": "^7.0.0",
     "@nestjs/platform-express": "^7.0.0",
     "@nestjs/typeorm": "^7.0.0",
     "@types/lodash": "^4.14.156",
-    "geotiff": "^1.0.0-beta.12",
+    "@types/moment-timezone": "^0.5.13",
```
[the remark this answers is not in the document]
ericboucher: Just confirming that we are keeping those here for now?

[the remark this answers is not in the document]
bencpeters: Types actually should always be dev dependencies _unless_ they're needed by the runtime system (e.g. we're actually running typescript for the production process, rather than using the typescript compile to build and then output a JS app that can be run by regular node. Basically this means unless we're running using `ts-node`)

packages/api/scripts/backfill-daily-data.ts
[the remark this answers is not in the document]
bencpeters: #javascriptproblems. no worries then.

packages/api/src/utils/sofar.ts
bencpeters: Doesn't hurt! I like the explicit vs. implicit thing ;-)

packages/api/.eslintrc
```diff
@@ -21,12 +21,7 @@
       "files": ["scripts/*"],
       "rules": {
         "no-console": "off",
-        "import/no-extraneous-dependencies": [
-          "error",
-          {
-            "devDependencies": true
-          }
-        ]
+        "import/no-extraneous-dependencies": "off"
```
bencpeters, line 24: We don't want to do this - this is saying that we'll let scripts import modules that don't appear in any dependency lists. I think the previous setting is actually what we want... (allows imports that only appear in `devDependencies`)

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,214 @@
...
+  const satelliteTemperatureData = await getSofarDailyData(
+    'HYCOM',
+    'HYCOM-seaSurfaceTemperature',
+    latitude,
+    longitude,
+    localTimezone,
+    date,
+  );
+
+  // Get satelliteTemperature closest to midnight local time by grabbing the last datapoint.
```
drewjgray, line 69: Hopefully the NOAA data source is already reporting calibrated data (e.g. satellite readings from midnight).

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,214 @@
...
+      date,
+    )
+  ).map(({ value }) => value);
+
+  const waveDirection = getAverage(meanDirectionWindWaves);
+
+  const meanPeriodWindWaves = (
+    await getSofarDailyData(
+      'NOAAOperationalWaveModel',
+      'NOAAOperationalWaveModel-meanPeriodWindWaves',
```
drewjgray, line 108: I believe we should use the ```NOAAOperationalWaveModel-peakPeriod``` instead of the wind waves here.

packages/api/src/workers/dailyData.ts
```diff
@@ -0,0 +1,214 @@
...
+      'GFS',
+      'GFS-magnitude10MeterWind',
+      latitude,
+      longitude,
+      localTimezone,
+      date,
+    )
+  ).map(({ value }) => value);
+
+  const windDirection = getAverage(windDirections);
```
drewjgray, line 145: We shouldn't average wind directions. Often times throughout the day the direction flips 180 from the heating and cooling of land near shore. The average will be meaningless. Let's show the closest to real-time data in the weather cards instead of averages.

ericboucher: What do we store in the dailyData then?

PR #57, opened 2020-07-16
Add user management

vickykont-programize:
This PR adds Firebase authentication and level (role) management.

Regarding authentication, a valid `accessToken` is expected to be provided as `Bearer`.

Regarding level management, an `Auth(...levels)` decorator is added. If `levels` are specified, the back-end only accepts requests from authenticated users with a `level` that belongs in `levels`. If no `levels` are provided, the back-end only checks whether the user is authenticated.

Also, as part of the PR, an endpoint `[PUT] /users/:id/level/:level` is added, to allow for managing user levels.


packages/api/example.firebase.json
```diff
@@ -0,0 +1,12 @@
+{
```
ericboucher, line 1: It's probably easier to rely on the .env for now and not add an additional config file.

packages/api/src/auth/firebase-auth.strategy.ts
```diff
@@ -0,0 +1,33 @@
...
+
+@Injectable()
+export class FirebaseAuthStrategy extends PassportStrategy(Strategy) {
+  constructor(private usersService: UsersService) {
+    super();
+  }
+
+  async authenticate(req: any): Promise<void> {
+    const self = this;
+    const token = ExtractJwt.fromAuthHeaderAsBearerToken()(req);
```
bencpeters, line 16: This seems like a little bit of a large dependency to add (`passport-jwt`) just to extract the header.

Let's just write it by hand, maybe something like this? (untested)

```typescript
function extractTokenFromHeader(req: Request): string | undefined {
  const authHeader = req.headers.authorization
  const match = authHeader && authHeader.match(/bearer (.*)/i)
  return match && match[1]
}
```

packages/api/src/main.ts
```diff
@@ -1,9 +1,24 @@
...
+  // Pass
+}
+
 async function bootstrap() {
+  // eslint-disable-next-line import/no-dynamic-require, global-require
+  const serviceAccount = require(process.env.FIREBASE_SERVICE_ACCOUNT ||
+    'firebase.json');
+  admin.initializeApp({
+    credential: admin.credential.cert(serviceAccount),
+    databaseURL: process.env.FIREBASE_DATABASE_URL || '',
```
bencpeters, line 20: I don't _think_ we're using the firebase DB for anything right now, do we need this?

packages/api/src/reefs/reefs.module.ts
```diff
@@ -5,10 +5,12 @@ import { ReefsService } from './reefs.service';
...
 import { EntityExists } from '../validations/entity-exists.constraint';
+import { UsersService } from '../users/users.service';
+import { User } from '../users/users.entity';
 
 @Module({
-  imports: [TypeOrmModule.forFeature([Reef, DailyData])],
+  imports: [TypeOrmModule.forFeature([Reef, User, DailyData])],
   controllers: [ReefsController],
-  providers: [ReefsService, EntityExists],
+  providers: [ReefsService, UsersService, EntityExists],
```
bencpeters, line 14: Why do we need this dependency? I didn't see us using the `UsersService` anywhere on this module? If it's part of the `Auth` guard, could we just import the `Auth` module instead?

packages/api/src/reefs/reefs.module.ts
```diff
@@ -5,10 +5,12 @@ import { ReefsService } from './reefs.service';
 import { Reef } from './reefs.entity';
 import { DailyData } from './daily-data.entity';
 import { EntityExists } from '../validations/entity-exists.constraint';
+import { UsersService } from '../users/users.service';
+import { User } from '../users/users.entity';
 
 @Module({
-  imports: [TypeOrmModule.forFeature([Reef, DailyData])],
+  imports: [TypeOrmModule.forFeature([Reef, User, DailyData])],
```
bencpeters, line 12: This feels a little bit redundant to me - I feel like `User` should be injected from the nest perspective over in the UserModule, and other modules shouldn't need to do it? (or they would just import the whole `User` module as a dependency?).

Does Nest complain if you don't put this in?

packages/api/src/regions/regions.module.ts
```diff
@@ -4,10 +4,12 @@ import { RegionsController } from './regions.controller';
...
 import { EntityExists } from '../validations/entity-exists.constraint';
+import { UsersService } from '../users/users.service';
+import { User } from '../users/users.entity';
 
 @Module({
-  imports: [TypeOrmModule.forFeature([Region])],
+  imports: [TypeOrmModule.forFeature([Region, User])],
   controllers: [RegionsController],
-  providers: [RegionsService, EntityExists],
+  providers: [RegionsService, UsersService, EntityExists],
```
bencpeters, line 13: Same comments as for the `reef`  module - it seems like we probably don't need these dependencies, I'd more just expect us to `import` the `auth` module?

packages/api/src/users/users.service.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  async create(createUserDto: CreateUserDto): Promise<User> {
+    const { firebaseUid } = createUserDto;
+    const alreadyExists = await this.findByFirebaseUid(firebaseUid);
+    if (alreadyExists) {
+      throw new BadRequestException(
+        `User with firebaseUid ${firebaseUid} already exists.`,
+      );
+    }
+    return this.usersRepository.save(createUserDto);
```
bencpeters, line 26: I think we need a couple more checks here:

1) Verify that the firebase token is valid. We should check with Firebase to make sure that it's not just some random string, but is actually a valid token. This check also lets us pull the `firebaseUid` and `email` associated with the token.

2) Verify that the email in the `createUserDto` matches the firebase email associated with the `firebaseUid`. With the way the code is written right now, you could potentially create an account with firebase using one email, get that token, then use the token to POST a request with a totally different email and get an account on the AquaLink platform with an email that you don't actually own.

packages/api/src/users/dto/create-user.dto.ts
```diff
@@ -0,0 +1,48 @@
...
+  @IsNotEmpty()
+  readonly fullName?: string;
+
+  @IsEmail()
+  readonly email: string;
+
+  @IsString()
+  @MinLength(1)
+  @MaxLength(128)
+  readonly firebaseUid: string;
```
bencpeters, line 24: This actually shouldn't get sent with the web request directly. We want to make sure that the user creating the account actually is who they say they are (from Firebase's perspective). To do this, the create request should have a Firebase token (e.g. a `Authorization: Bearer <token>` header) already. The backend then needs to validate that token with firebase, and use the admin SDK to get the `firebaseUid` directly from firebase, rather than relying on the user to provide it in the request.

packages/api/src/users/users.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+export class UsersController {
+  constructor(private usersService: UsersService) {}
+
+  @Post()
+  create(@Req() req: any, @Body() createUserDto: CreateUserDto): Promise<User> {
+    return this.usersService.create(req, createUserDto);
+  }
+
+  @Auth()
+  @Get('self')
```
bencpeters, line 28: Let's call this `'current'` instead (for the route).

packages/api/src/users/users.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+  getSelf(@Req() req: AuthRequest): Promise<User | undefined> {
+    return this.usersService.getSelf(req);
+  }
+
+  @Auth(AdminLevel.SuperAdmin)
+  @Put(':id/level/:level')
```
bencpeters, line 34: I like I'd probably put this param (`level`) in the body rather than the URL, although it's not a huge deal either way.

PR #58, opened 2020-07-17
Homepage

tleraj:
This PR concerns the application's homepage. The implementation right now gives users the ability to select a reef from the table and fly to that specific reef, and jump to that specific reef details page by clicking the Explore button. The design looks like this:

Screenshot from 2020-07-17 18-27-35

Features to be implemented:
- [x] Firebase authentication (log in/out, register)
- [x] Zoom in reef on table row selection
- [x] UI error handling in failed authentication

TODO in a future PR:
- Search location implementation
- Change reef detail page header
- Make reef detail page map larger if there is no video.


packages/website/src/services/userServices.ts
```diff
@@ -0,0 +1,15 @@
+import app from "../firebase";
+
+const createUser = (email: string, password: string) =>
+  app.auth().createUserWithEmailAndPassword(email, password);
+
+const signInUser = (email: string, password: string) =>
+  app.auth().signInWithEmailAndPassword(email, password);
```
bencpeters, line 7: These are going to have to also include backend requests as soon as we have the backend PR (#57) merged -> let's make sure these are async functions.

packages/website/src/store/User/userSlice.ts
```diff
@@ -0,0 +1,179 @@
...
+  async ({ email, password }: UserRequestParams, { rejectWithValue }) => {
+    try {
+      const { user } = await userServices.createUser(email, password);
+      return {
+        email: user?.email,
+        uid: user?.uid,
+        token: await user?.getIdToken(),
+      };
+    } catch (err) {
+      const error: AxiosError<UserState["error"]> = err;
```
bencpeters, line 29: This isn't actually an `AxiosError` in a lot of cases, right? If the error is throne from the Firebase operation, it's going to be a firebase error...

packages/website/src/store/configure.ts
```diff
@@ -1,9 +1,28 @@
...
+import storage from "redux-persist/lib/storage";
 
 /* eslint-disable-next-line import/no-cycle */
 import reducer from "./reducer";
 
-export const store = configureStore({ reducer });
+const persistConfig = {
+  key: "aqualink-app-storage",
+  storage,
+  whitelist: ["user"],
```
bencpeters, line 11: We don't want to store the token in local storage - that opens us to potential XSS attacks. This `redux-persist` library is storing its data there...

if you need to initialize the redux store with a token, use `firebase.auth().currentUser.getIdToken()`

packages/website/src/routes/Homepage/ReefTable/selectedReefCard.tsx
```diff
@@ -0,0 +1,161 @@
...
+  createStyles,
+  Theme,
+} from "@material-ui/core";
+import { Link } from "react-router-dom";
+
+import CardChart from "./cardChart";
+import { Reef } from "../../../store/Reefs/types";
+import { sortDailyData } from "../../../helpers/sortDailyData";
+
+const reefImage = require("../../../assets/reef-image.png");
```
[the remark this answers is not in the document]
tleraj: Changed to `import`. I added `declare module "*.png"` in the `custom.d.ts` file.

packages/website/src/routes/Homepage/Map/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const { current } = mapRef;
+    if (current && current.leafletElement) {
+      const map = current.leafletElement;
+      const southWest = L.latLng(-90, -240);
+      const northEast = L.latLng(90, 240);
+      const bounds = L.latLngBounds(southWest, northEast);
+      map.setMaxBounds(bounds);
+      map.flyTo(center, zoom, { duration: 1 });
+    }
+  }, [center, zoom]);
```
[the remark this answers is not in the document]
tleraj: Added `mapRef` as a dependency, but I have to say that eslint was not complaining without them.

packages/website/src/services/userServices.ts
tleraj: I will implement this logic as soon as the backend PR is merged.

packages/website/src/routes/Homepage/RegisterDialog/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+      event?: BaseSyntheticEvent<object, HTMLElement, HTMLElement>
+    ) => {
+      if (event) {
+        event.preventDefault();
+      }
+      const registerInfo: UserRequestParams = {
+        email: data.emailAddress,
+        password: data.password,
+      };
+      dispatch(createUser(registerInfo));
```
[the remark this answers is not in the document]
tleraj: Right now only the firebase register (which only needs email + password) is implemented. In the future, when https://github.com/aqualinkorg/aqualink-app/pull/57 will be merged, it will require other parameters (such as fullName, organization, etc).

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Environment variables
 .env*
 !.env.example
+env/
```
ericboucher, line 4: why this new folder?

packages/website/.env.example
```diff
@@ -1 +1,11 @@
-REACT_APP_API_BASE_URL="http://localhost:8080/"
+# Backend base URL
+REACT_APP_API_BASE_URL = 
+
+# Firebase
```
ericboucher, line 4: Do we actually need all of these variables?

.gitignore
tleraj: This is just to ignore the virtual environment folder.

packages/website/src/routes/Homepage/ReefTable/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ericboucher, line 1: Should these snaps be empty? This doesn't seem to be a very valuable test ;)

packages/website/.env.example
tleraj: Yes. As you can see in `src/firebase.ts`, these keys are used to configure firebase (as mentioned in the documentation: https://firebase.google.com/docs/web/setup).

packages/website/src/routes/Homepage/RegisterDialog/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ericboucher, line 1: Same here. Let's try to have useful tests / snapshots, or no snapshot comparison if we can't create a useful one easily.

packages/website/src/routes/Homepage/SignInDialog/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
```
ericboucher, line 2: Similar snapshot issue

packages/website/src/routes/Homepage/RegisterDialog/__snapshots__/index.test.tsx.snap
tleraj: My bad...I was writing `describe("Homepage", () => {...})` instead of `describe("RegisterDialog", () => {...})`. I made an update, but still I'm not sure if the produced snapshot is useful. I think we could just remove this snapshot test (along with `SignInDialog`).

packages/website/.env.example
```diff
@@ -1 +1,11 @@
...
+REACT_APP_FIREBASE_API_KEY =
+REACT_APP_FIREBASE_AUTH_DOMAIN = 
+REACT_APP_FIREBASE_DABABASE_URL =
```
ericboucher, line 7: typo?

packages/website/.env.example
ericboucher: Yes but we probably don't need the full config for our use case.

.gitignore
ericboucher: Which virtual env? 

Then add a new category 

# Virtual env

packages/website/src/routes/Homepage/Map/Popup/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ericboucher, line 1: Another small snap

PR #66, opened 2020-08-05
Augment get-reefs endpoint to include latest daily data

vickykont-programize:
This PR augments the `[GET] /reefs` endpoint to include the latest available daily data.

### DIFF
--- a/packages/api/src/reefs/reefs.entity.ts
+++ b/packages/api/src/reefs/reefs.entity.ts
@@ -7,7 +7,7 @@ import {
   ManyToOne,
   CreateDateColumn,
   UpdateDateColumn,
-  OneToMany,
+  OneToOne,
 } from 'typeorm';
 import { Region } from '../regions/regions.entity';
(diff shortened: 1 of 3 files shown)


packages/api/src/reefs/reefs.entity.ts
```diff
@@ -54,4 +57,7 @@ export class Reef {
 
   @ManyToOne(() => VideoStream, { onDelete: 'CASCADE', nullable: true })
   stream?: VideoStream;
+
+  @OneToMany(() => DailyData, (dailyData) => dailyData.reef)
+  dailyData?: DailyData[];
```
ericboucher, line 62: should this be DailyData, no array if we are only getting the latest info?

vickykont-programize: Since this is an `OneToMany` relationship, in general (if we performed a `JOIN` between the two tables), we could have multiple `DailyData` for the same `Reef`, hence the array.

PR #68, opened 2020-08-13
Add survey endpoints

vickykont-programize:
This PR adds CRUD endpoints for surveys, survey media and reef points of interest. It also introduces file-upload functionality.

To handle file uploads easier, a new decorator `@AcceptFile(param, acceptTypes, subfolder, prefix)` is added.
- `param`: is the name of the parameter that holds the file.
- `acceptTypes`: is a sub-array of `['image', 'video']`.
- `subfolder`: is the subfolder to use.
- `prefix`: is the `prefix` used to generate the name of the file stored.


packages/api/.env.example
```diff
@@ -6,3 +6,8 @@ POSTGRES_DATABASE=postgres
 FIREBASE_PROJECT_ID=
 FIREBASE_PRIVATE_KEY=
 FIREBASE_CLIENT_EMAIL=
+GC_PROJECT=
+GCS_BUCKET=
+GCS_KEYFILE=gcloud.json
```
ericboucher, line 11: Add to .gitignore?

PR #92, opened 2020-08-26
Add survey-media endpoints

avalmas-programize:
- Add get survey media endpoint
- Hide images based on hidden flag
- Return featured image when retrieving the survey list
- Add @google-cloud/storage npm package
- Add GoogleCloudService to add google cloud specific functionality (deleteFile)

### DIFF
--- a/packages/api/src/surveys/surveys.entity.ts
+++ b/packages/api/src/surveys/surveys.entity.ts
@@ -6,6 +6,7 @@ import {
   JoinColumn,
   CreateDateColumn,
   UpdateDateColumn,
+  OneToOne,
 } from 'typeorm';
 // eslint-disable-next-line import/no-cycle
 import { Reef } from '../reefs/reefs.entity';
@@ -14,6 +15,8 @@ import { User } from '../users/users.entity';
 import { DailyData } from '../reefs/daily-data.entity';
 // eslint-disable-next-line import/no-cycle
 import { ReefPointOfInterest } from '../reef-pois/reef-pois.entity';
+// eslint-disable-next-line import/no-cycle
+import { SurveyMedia } from './survey-media.entity';
 
 export enum WeatherConditions {
   Calm = 'calm',
@@ -55,6 +58,9 @@ export class Survey {
 
   @UpdateDateColumn()
   updatedAt?: Date;
+
+  @OneToOne(() => SurveyMedia, (surveyMedia) => surveyMedia.surveyId)
+  featuredSurveyMedia?: SurveyMedia;
 
   latestDailyData?: DailyData;
 
--- a/packages/api/src/uploads/file.decorator.ts
+++ b/packages/api/src/uploads/file.decorator.ts
@@ -33,19 +33,24 @@ export const AcceptFile = (
     ? parseInt(process.env.STORAGE_MAX_FILE_SIZE_MB, 10)
     : 1;
   const maxFileSizeB = maxFileSizeMB * 1024 * 1024;
+  // Detach config object from MulterConfigurationOptions because
+  // param acl is not currently documented in the types file of the extension 'multer-google-storage'
+  // although the functionality for access control exists
+  const config: any = {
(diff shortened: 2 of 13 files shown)


packages/api/src/uploads/file.decorator.ts
```diff
@@ -33,19 +33,24 @@ export const AcceptFile = (
     ? parseInt(process.env.STORAGE_MAX_FILE_SIZE_MB, 10)
     : 1;
   const maxFileSizeB = maxFileSizeMB * 1024 * 1024;
+  // Detach config object from MulterConfigurationOptions because
+  // param acl is not currently documented in the types file of the extension 'multer-google-storage'
+  // although the functionality for access control exists
+  const config: any = {
```
ericboucher, line 39: Maybe you can use the type of MulterGoogleCloudStorage instead of any?

packages/api/src/surveys/surveys.entity.ts
```diff
@@ -56,6 +59,9 @@ export class Survey {
   @UpdateDateColumn()
   updatedAt?: Date;
 
+  @OneToOne(() => SurveyMedia, (surveyMedia) => surveyMedia.surveyId)
+  featuredSurveyMedia?: SurveyMedia;
```
ericboucher, line 63: The statement below doesn't guarantee a oneToOne relationship, no?

```
.leftJoinAndSelect(
        'survey.featuredSurveyMedia',
        'featuredSurveyMedia',
        'featuredSurveyMedia.featured = True',
      )
```

packages/api/src/uploads/file.decorator.ts
avalmas-programize: No we cannot.
The constructor of a `MulterGoogleCloudStorage` accepts an `opts` parameter which is of type:
```js
(1): ConfigurationObject & { filename?: any, bucket?:string, contentType?: ContentTypeFunction }
``` 

Also in the `MulterGoogleCloudStorage` class an `options` global variable is created of type: 
```js
(2): ConfigurationObject & { acl?: string, bucket?: string, contentType?: ContentTypeFunction }
``` 
The two types are almost the same except the first one has the filename param which is of type any and the second has the acl which is of type string
At the end of the constructor the `opts` is assigned to the `options` variable which is used later in the class.
So in order to create a MulterGoogleCloudStorage I need to pass an `opts` of type (1) which does not have the acl parameter which I need to specify to "publicread" (the default is "private") to upload a file which public access.
As a result I use `any` instead of any other specific type to prevent typechecking because the multer-google-storage package hasn't documented its types properly.
All of the above can be seen in the node_modules/multer-google-storage/src/index.ts file.

packages/api/src/surveys/surveys.entity.ts
avalmas-programize: There only one featured survey media per survey, right?
So every will be related to only one featured survey media.
Should I include more checks in this query to ensure the OneToOne relationship?

antoniosValmas:
Featured survey media guards have been implemented according to the following principles.
**Create new survey media:**
- If no other featured survey media exists in the survey then the newly added media is the featured media of the survey
- If a featured survey media already exists in the database and the newly created one is also featured, the new becomes featured media and the old is "unfeatured"
- A featured media **cannot** be hidden.

**Edit survey media:**
- If a survey media is updated from non-featured to featured then the old one is "unfeatured"
- If a featured media is updated to hidden then it is "unfeatured" and a random media (from the survey) is chosen for the new featured media of the survey
- The `hidden` and `featured` params are now required in the updateMedia endpoint in order to not have to deal with undefined values.

**Delete survey media:**
- If the featured media gets deleted, a random media (from the survey) is chosen for the new featured media of the survey

ericboucher:
```The hidden and featured params are now required in the updateMedia endpoint in order to not have to deal with undefined values.``` @antoniosValmas having the `= false` default was not enough to not have undefined?

ericboucher (approved):

Issue #78, opened 2020-08-19
Implement new table designs for mobile


ericboucher:
Implement mobile designs for the home page and make the app mobile friendly

![image](https://user-images.githubusercontent.com/16843267/90683949-35ee6900-e21c-11ea-85a0-65451650589f.png)

![image (1)](https://user-images.githubusercontent.com/16843267/90683959-3981f000-e21c-11ea-91a6-b90cb66ddba5.png)

SimplyBallistic:
Where are we at currently with this issue? Is the last step to make the table look more like what's in the screenshots?

SimplyBallistic:
* working on restyling the table to fit the given mockup

PR #97, opened 2020-08-26
Display map on mobile and move ReefTable to pullout drawer component

dianakw8591:
The main goal of this PR is to refactor the homepage on mobile devices to put the ReefTable in a drawer component. This component is openable by clicking on it when it is in the closed state, and closes when any item is clicked in the list or when the NavBar is clicked. To my great frustration, I was not able to implement this component as Material UIs `<SlideableDrawer>` and still maintain a permanently visible part of the drawer, so while the drawer *looks* slideable it is actually open/closed by clicking. Obviously this is not ideal - another implementation would be to have a button to open when the drawer is in its closed state (technically a slideable drawer can be slid open if you know it's hiding at the bottom of the screen, although that can interfere with map dragging) and the drawer could be closed by sliding. I opted for this implementation that matches the original designs but lacks some of the desired functionality.
Screen Shot 2020-08-26 at 1 01 50 PM
Screen Shot 2020-08-26 at 12 43 54 PM

This PR also adds viewport metadata and map container styling that should fix map rendering issues on mobile.

Addresses issue #78 , additional styling needed on featured reef card for mobile view.


packages/website/src/routes/Homepage/ReefTable/index.test.tsx
```diff
@@ -41,11 +41,13 @@ describe("ReefTable", () => {
       },
     });
 
+    const openDrawer = false;
```
ericboucher, line 44: should we test the positive case as well?

ericboucher (approved):

PR #101, opened 2020-08-28
Added mobile responsive design to SelectedReefCard

dianakw8591:
SelectedReefCard display improvements for web and mobile views.
Screen Shot 2020-08-28 at 11 50 05 AM
Screen Shot 2020-08-28 at 11 50 26 AM
Screen Shot 2020-08-28 at 11 50 56 AM
The third image is from an iPad-sized screen - in this size range there are definite improvements to make still but all the content is at visible. Could consider using the mobile layout for this screen size?

### DIFF
--- a/packages/website/src/common/NavBar/__snapshots__/index.test.tsx.snap
+++ b/packages/website/src/common/NavBar/__snapshots__/index.test.tsx.snap
@@ -54,7 +54,7 @@ exports[`NavBar should render with given state from Redux store 1`] = `
           </h4>
         </div>
         <div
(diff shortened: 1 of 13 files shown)


ericboucher:
@dianakw8591 maybe we could have the font size adapt a bit to keep the temp on one line?

dianakw8591:
@ericboucher after playing with font sizes on the different screen sizes a bit, the font has to be so small to fit in one line on the smallest of the 'small' sizes (right before it hits the breakpoint to change views) that I'm not sure it's better than keeping the text on two lines...what do you think?
Screen Shot 2020-08-28 at 5 19 37 PM
It certainly loses the element of importance that a larger font size gives.

PR #103, opened 2020-08-31
🎨 Enhance Navbar on different screens

mamadOuologuem:
Screen Shot 2020-09-02 at 1 33 15 PM
Screen Shot 2020-09-02 at 1 33 00 PM

### DIFF
--- a/packages/website/src/common/NavBar/index.tsx
+++ b/packages/website/src/common/NavBar/index.tsx
@@ -48,7 +48,7 @@ const NavBar = ({ searchLocation, classes }: NavBarProps) => {
     <>
       <AppBar className={classes.appBar} position="static" color="primary">
         <Toolbar className={classes.toolbar}>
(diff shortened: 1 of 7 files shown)


ericboucher (approved):

packages/website/src/common/NavBar/index.tsx
```diff
@@ -119,7 +119,7 @@ const NavBar = ({ searchLocation, classes }: NavBarProps) => {
 
             {searchLocation && (
               <Hidden smUp>
-                <Grid item xs={12}>
+                <Grid item xs={12} style={{ margin: 0, paddingTop: 0 }}>
```
mamadOuologuem, line 122: Not a big deal

But we should try to [avoid inline styling](https://www.codecademy.com/articles/html-inline-styles) as much as we can.

mamadOuologuem (commented):
Looks good to me

Issue #111, opened 2020-09-02
Detect new users as reef admins


ericboucher:
If a user signs up with an email address that already exists in our users table, we want:

- that they take ownership of this row
- that they become a "reef_admin" and get access to upload surveys to reefs they administer (and only those)

PR #128, opened 2020-09-04
Add reef admin relationships

avalmas-programize:
- Add `reef-to-admin` table (custom intermediate table for manyToMany relationship between admin and reef)
- Change reef \ admin relationship from manyToOne (users can have many reefs but a reef has only one admin) to manyToMany (reefs now can have more than one admin)
- Refactor reef endpoints to follow the manyToMany relationship between user and reef
- Add IsReefAdminGuard to restrict access to reef_managers to only their reefs
- Refactor surveys endpoints:
  - Change survey's global prefix to `reefs/:reef_id/surveys` so that we always track the reef_id, which is needed for IsReefAdminGuard
  - Change GET /surveys/reefs/:id to /reefs/:reef_id/surveys
  - Make necessary changes in the DTOs and survey service

Some front-end refactor is needed due to the prefix change

Related issue: https://github.com/aqualinkorg/aqualink-app/issues/111


packages/api/src/users/users.service.ts
```diff
@@ -75,4 +88,28 @@ export class UsersService {
...
+
+    const reefToAdminEntities: Promise<ReefToAdmin>[] = reefAssociations.map(
+      (reefAssociation) => {
+        const reefToAdmin = new ReefToAdmin();
+        // eslint-disable-next-line fp/no-mutation
+        reefToAdmin.adminId = user.id;
+        // eslint-disable-next-line fp/no-mutation
+        reefToAdmin.reefId = reefAssociation.reef.id;
+
+        return this.reefToAdminRepository.save(reefToAdmin);
```
ericboucher, line 109: I think you can just do:

this.reefToAdminRepository.save({adminId: user.id, reefId: reefAssociation.reef.id})

packages/api/src/users/users.service.ts
```diff
@@ -9,12 +9,20 @@ import { AuthRequest } from '../auth/auth.types';
...
+import { ReefToAdmin } from '../reefs/reef-to-admin.entity';
+import { ReefApplication } from '../reef-applications/reef-applications.entity';
 
 @Injectable()
 export class UsersService {
   constructor(
     @InjectRepository(User)
     private usersRepository: Repository<User>,
+
+    @InjectRepository(ReefToAdmin)
```
ericboucher, line 21: Let's add an endpoint to return the list of reefs administered by the current logged in user

packages/api/src/reefs/reef-to-admin.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+  @PrimaryGeneratedColumn()
+  reefToAdminId: number;
+
+  @Column()
+  reefId: number;
+
+  @Column()
+  adminId: number;
+
+  @ManyToOne(() => Reef, (reef) => reef.reefToAdmin, { cascade: true })
```
ericboucher, line 18: Is it safe to have a cascade true here? Doesn't that mean that we would delete the reef if we delete this relationship?

ericboucher:
@avalmas-programize can you list the necessary frontend changes? Or even have a stab at them? They shouldn't be too complicated at the moment.

packages/api/migration/1599209835916-AddReefToAdminTable.ts
```diff
@@ -0,0 +1,46 @@
...
+export class AddReefToAdminTable1599209835916 implements MigrationInterface {
+  name = 'AddReefToAdminTable1599209835916';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      CREATE TABLE "reef_to_admin" (
+        "reef_to_admin_id" SERIAL NOT NULL,
+        "reef_id" integer NOT NULL,
+        "admin_id" integer NOT NULL,
+        CONSTRAINT "PK_481618ed9eb45a5765dd8710596" PRIMARY KEY ("reef_to_admin_id")
```
ericboucher, line 12: We should probably limit duplicates of reef_id, admin_id?

packages/api/src/reefs/reefs.controller.ts
```diff
@@ -44,17 +43,6 @@ export class ReefsController {
     return this.reefsService.findDailyData(id);
   }
 
-  @Get(':id/surveys/:poi')
```
ericboucher, line 47: is this not needed anymore?

packages/api/src/users/users.service.ts
avalmas-programize: Yes indeed. I forgot about this syntax.

packages/api/src/reefs/reef-to-admin.entity.ts
avalmas-programize: No. Cascade means that when the Reef is deleted the relationship is deleted also. (the same goes for users)

packages/api/src/reefs/reefs.controller.ts
avalmas-programize: As I see it, this endpoint was using a mock response which is already implemented in the survey details endpoint (GET /reefs/:reef_id/surveys/:id). The only difference was that I fetch all POIs of a survey and this was designed to fetch only one POI of a survey. I don't see any use for it anymore.
Also it was conflicting with the new endpoints (surveys endpoints with the updated prefix)

packages/api/migration/1599209835916-AddReefToAdminTable.ts
avalmas-programize: I will add a unique key and some checks when the relation is created to prevent any duplicates.

avalmas-programize:
Front-end changes:
All `surveys/` endpoints URLs should changes:
- POST /surveys/                                => POST /reefs/:reef_id/surveys/
- POST /surveys/:survey_id/media/    => POST /reefs/:reef_id/surveys/:survey_id/media/
- POST /surveys/upload/                     => POST /reefs/:reef_id/surveys/upload/
- GET /surveys/reefs/:reef_id              => GET /reefs/:reef_id/surveys/
- GET /surveys/:survey_id/                  => GET /reefs/:reef_id/surveys/:survey_id/
- GET /surveys/:survey_id/media/       => GET /reefs/:reef_id/surveys/:survey_id/media/
- PUT /surveys/:survey_id/                  => PUT /reefs/:reef_id/surveys/:survey_id/
- PUT /surveys/media/:media_id/        => PUT /reefs/:reef_id/surveys/:survey_id/
- DELETE /surveys/:survey_id/           => PUT /reefs/:reef_id/surveys/:survey_id/
- DELETE /surveys/media/:media_id/ => PUT /reefs/:reef_id/surveys/:survey_id/

packages/api/migration/1599209835916-AddReefToAdminTable.ts
bencpeters: @avalmas-programize A better approach would be to make the primary key itself `(reef_id, admin_id)`

packages/api/src/reefs/reef-to-admin.entity.ts
```diff
@@ -0,0 +1,30 @@
...
+  Unique,
+} from 'typeorm';
+// eslint-disable-next-line import/no-cycle
+import { Reef } from './reefs.entity';
+// eslint-disable-next-line import/no-cycle
+import { User } from '../users/users.entity';
+
+@Entity()
+@Unique('unique-relationship', ['reefId', 'adminId'])
+export class ReefToAdmin {
```
bencpeters, line 15: You shouldn't need to define a TypeORM entity at all for a join table unless you're trying to store custom information on the join table (e.g. an attribute that describes a condition on the join). In this case you're just trying to set up a many-to-many join between two entities, so defining it in this way has numerous disadvantages:

- This is a class based entity that TypeORM will actually try to instantiate if you load records in some cases, increasing the overhead.
- You have a completely extraneous extra column (`reefToAdminId`) that conveys no extra information and requires a hacky `unique` constraint (the "right" way to set up a table like this would just use the `reefId, adminId` as the primary key.
- This entity forces you to specify a naming convention (which in this case you didn't use the "standard" SQL naming convention for join tables across most web frameworks which would be `reefs_admins`) rather than relying on predictable & standard defaults.

See the [docs on Many-To-Many in TypeORM](https://typeorm.io/#/many-to-many-relations).

This is what you want to do:

```typescript
class Reef {
  // Other attributes
  @ManyToMany(() => User, (user) => user.administered_reefs)
  @JoinTable()
  admins: User[]
}

class User {
  // Other attributes
  @ManyToMany(() => Reef, (reef) => reef.admins)
  administered_reefs: Reef[]
}
```

packages/api/src/auth/is-reef-admin.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+    const request = context.switchToHttp().getRequest();
+    const { user }: { user: User } = request;
+    const reefId: number = request.params.reef_id;
+
+    if (user.adminLevel === AdminLevel.SuperAdmin) {
+      return true;
+    }
+
+    if (reefId) {
+      const isReefAdmin = await this.reefToAdminRepository.findOne({
```
bencpeters, line 23: We have an `adminLevel` that corresponds to a reef manager, do we want to enforce that that admin level is set as well?

packages/api/src/reefs/reefs.entity.ts
```diff
@@ -11,12 +11,13 @@ import {
...
 } from 'typeorm';
 import { Region } from '../regions/regions.entity';
-import { User } from '../users/users.entity';
 // eslint-disable-next-line import/no-cycle
 import { DailyData } from './daily-data.entity';
 import { VideoStream } from './video-streams.entity';
 // eslint-disable-next-line import/no-cycle
 import { Survey } from '../surveys/surveys.entity';
+// eslint-disable-next-line import/no-cycle
+import { ReefToAdmin } from './reef-to-admin.entity';
```
bencpeters, line 20: `import type { ReefToAdmin }` (the other no cycle imports should use this as well - get rid of all the linter disable comments)

packages/api/src/surveys/surveys.controller.ts
```diff
@@ -20,11 +21,13 @@ import { CreateSurveyMediaDto } from './dto/create-survey-media.dto';
...
 import { EditSurveyDto } from './dto/edit-survey.dto';
 import { EditSurveyMediaDto } from './dto/edit-survey-media.dto';
+import { IsReefAdminGuard } from '../auth/is-reef-admin.guard';
 
-@Controller('surveys')
+@Controller('reefs/:reef_id/surveys')
 export class SurveysController {
   constructor(private surveyService: SurveysService) {}
 
+  @UseGuards(IsReefAdminGuard)
```
bencpeters, line 30: Rather than requiring us to specify this guard on every single handler, it might be worth adding a `@Public` decorator that does `setMetadata("isPublic", true)` and then look up the metadata on the handler inside the `isReefAdminGuard` function and only lets the request through without the check if that metadata is set. Then you could specify the guard on a controller-wise level here.

packages/api/src/surveys/surveys.entity.ts
```diff
@@ -10,6 +10,7 @@ import {
 } from 'typeorm';
 // eslint-disable-next-line import/no-cycle
 import { Reef } from '../reefs/reefs.entity';
+// eslint-disable-next-line import/no-cycle
```
bencpeters, line 13: again, use `import type { ... }`

packages/api/src/users/users.entity.ts
```diff
@@ -6,8 +6,11 @@ import {
   Index,
   CreateDateColumn,
   UpdateDateColumn,
+  OneToMany,
 } from 'typeorm';
 import { Exclude } from 'class-transformer';
+// eslint-disable-next-line import/no-cycle
+import { ReefToAdmin } from '../reefs/reef-to-admin.entity';
```
bencpeters, line 13: another `import type { }`

packages/api/src/users/users.service.ts
```diff
@@ -75,4 +88,37 @@ export class UsersService {
...
+        const relationshipExists = await this.reefToAdminRepository.find({
+          where: {
+            adminId: user.id,
+            reefId: reefAssociation.reef.id,
+          },
+        });
+
+        // If relationship already exists, skip
+        if (relationshipExists) {
+          return Promise.resolve();
```
bencpeters, line 112: This is already an sync function, you don't need to return anything at all to get it to wrap in a `Promise` (or you could `return true` if that feels better)

bencpeters (changes requested):
Some suggested clean-ups throughout, but the big thing is re-doing the schema & entities a bit...

packages/api/src/reefs/reef-to-admin.entity.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  Entity,
+  PrimaryGeneratedColumn,
+  Column,
+  ManyToOne,
+  Unique,
+} from 'typeorm';
+// eslint-disable-next-line import/no-cycle
```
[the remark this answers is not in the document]
avalmas-programize: I removed all the lint disables from all the entities.
However the relations decorator expects a entity returned to use as a type for the relation.
To bypass this I had to change the way we declare a relationship from:
```js
@ManyToOne(() => Reef, (reef) => reef.reefToAdmin, { cascade: true })
reef: Reef;
```
to
```js
@ManyToOne<Reef>('Reef', (reef) => reef.reefToAdmin, { cascade: true })
reef: Reef;
```
But this does not solve all of the issues. After compiling the code an error appears in the dist files:
```
/home/antonios/Documents/Programize/aqualink-app-backend/packages/api/dist/src/reefs/reef-to-admin.entity.js
    __metadata("design:type", User)

ReferenceError: User is not defined
```
To solve this we have to create an interface for each entity and reference this as the type of the class member
https://github.com/typeorm/typeorm/issues/4190
Are we okay with this refactor?

packages/api/src/reefs/reef-to-admin.entity.ts
avalmas-programize: I wanted to add a custom join table in case we need it in the future.
I will revert my change

packages/api/src/auth/is-reef-admin.guard.ts
avalmas-programize: The routes this guard is added are the already protected ones, i.e the ones that do not allow the Default user to access them, only ReefManager and SuperAdmin.
In the first lines of this guard I bypass the guard if the level of the user is SuperAdmin.
So in this we are left with the case that the user is a ReefManager.
Also if in the future this is added to a not protected route, i.e a default user is tested to pass this check they will fail because the are not a manager of the reef, which makes perfect sense.
We should not allow a default user access more than a reef manager.

packages/api/src/reefs/reefs.entity.ts
avalmas-programize: Like I said this is more complicated than it seems.

packages/api/src/surveys/surveys.entity.ts
avalmas-programize: Like I said this is more complicated than it seems.

packages/api/src/users/users.entity.ts
avalmas-programize: Like I said this is more complicated than it seems.

packages/api/src/users/users.service.ts
avalmas-programize: It was not letting me to just write `return;` because map expects something returned so I thought I could return an empty Promise since that's what I want.
But I will change it to null since I agree it is a bit ugly.

packages/api/migration/1599209835916-AddReefToAdminTable.ts
avalmas-programize: Custom intermediate table has been deleted.
So these are by default primary keys now.

packages/api/src/auth/is-reef-admin.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+        where: {
+          reefId,
+          adminId: user.id,
+        },
+      });
+
+      return !!isReefAdmin;
+    }
+
+    return true;
```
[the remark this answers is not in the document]
avalmas-programize: Hmm I thought it from a different angle, but I see your point. I will return false by default.

packages/api/src/surveys/surveys.controller.ts
avalmas-programize: Interesting approach.
However I would name the decorator something more specific like `@ManagersOnly` because you would assume a `@Public` decorator to also have an affect to the `@Auth` decorator.

packages/api/src/users/users.service.ts
avalmas-programize: Added
GET users/current/administrated-reefs

packages/api/src/surveys/surveys.controller.ts
avalmas-programize: Hmmm. IsReefAdminGuard needs the Auth decorator to run first and since the first one is now a Controller decorator it will always run first. So we need a different approach if we want to implement this.
We could edit the Auth decorator in order to dynamically add the IsReefAdminGuard to the list of decorators used or have the new decorator (the Public one you suggested) affect both guards' behavior and add those as global or controller decorators

packages/api/src/reefs/reef-to-admin.entity.ts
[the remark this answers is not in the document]
bencpeters: Ah, I see the issue, we need to actually use the types as a return value, not just a type annotation.

New plan: Since circular dependencies are pretty much baked into how TypeORM does many-to-many relations, let's add a new ESlint rule override to disable that check in `*.entity.ts` files only.

packages/api/src/auth/is-reef-admin.guard.ts
bencpeters: That may be true in the current use-case - wherever we use this is already a protected route, but it would be fairly easy down the line for someone not to realize that. That's why I was suggesting adding an `adminLevel` check inside this guard so that you don't _have_ to ensure that a completely independent guard is also set whenever you use it.

packages/api/src/surveys/surveys.controller.ts
bencpeters: > have the new decorator (the Public one you suggested) affect both guards' behavior and add those as global or controller decorators

This is what I was imagining. If you move `IsReefAdminGuard` to the controller level, you definitely have to move the Auth guard as well.

packages/api/src/users/users.controller.ts
```diff
@@ -44,4 +45,10 @@ export class UsersController {
   delete(@Param('id', ParseIntPipe) id: number): Promise<void> {
     return this.usersService.delete(id);
   }
+
+  @Auth()
+  @Get('current/administrated-reefs')
+  getAdministratedReefs(@Req() req: AuthRequest): Promise<Reef[]> {
+    return this.usersService.getAdministratedReefs(req);
```
bencpeters, line 52: Let's call these "administered" (not "administrated") ([ref](https://site.uit.no/english/words/administrate/))

packages/api/src/users/users.service.ts
```diff
@@ -75,4 +101,41 @@ export class UsersService {
...
+
+        // Add new user as admin
+        return this.reefRepository.update(
+          {
+            id: reef.id,
+          },
+          {
+            admins: [...reef.admins, user],
+          },
+        );
```
bencpeters, line 135: Can we add this relation from the "other side" (e.g. using the `userRepository`) instead? I think this is the only dependency the UserService has on the `reefRepository`, so if we could add it from the "other" side of the relation we could drop that dependency from this service...

```typescript
this.usersRepository.update({ id: user.id }, { administeredReefs: ... })
```
Maybe?

packages/api/src/users/users.service.ts
```diff
@@ -75,4 +101,41 @@ export class UsersService {
...
+  /**
+   * Transfer the associations between the user and the reefs from the reef-application table
+   */
+  private async migrateUserAssociations(user: User) {
+    const reefAssociations = await this.reefApplicationRepository.find({
+      where: { user },
+      relations: ['reef', 'reef.admins'],
+    });
+
+    const reefEntities: Promise<UpdateResult | null>[] = reefAssociations.map(
```
bencpeters, line 114: Implicitly, a user who has user reef admin associations should also get the reef manager role, no?

bencpeters (changes requested):
Getting close!

A few small cleanups, then:

- for the cyclic dependencies, let's just leave them in the entity code, but add an eslint rule override for `*.entity.ts` files for that specific rule so that we don't need line-by-line overrides.
- For the `@Public` decorator - I think you'd need to have it affect the auth guard too, then it should work. I don't feel super strongly about it either way, but I generally think that it might be a cleaner way to define a controller where almost all the routes are protected but there are 1 or 2 public ones... if it's taking too long don't worry about it though.

packages/api/src/reefs/reef-to-admin.entity.ts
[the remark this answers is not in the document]
avalmas-programize: Okay

packages/api/src/users/users.controller.ts
avalmas-programize: Oops typo. Fixed

packages/api/src/users/users.service.ts
avalmas-programize: Done.

packages/api/src/users/users.service.ts
avalmas-programize: Yeah I thought it would have been set from the application. Done.

avalmas-programize:
I have added the Public decorator. Implementation:
- Override method `AuthGuard('custom').canActivate` so that it will not apply the strategy if isPublic equals to true
- Override Levels and ReefAdmin guards to return true if isPublic equals to true
- Add OverrideLevelAccess decorator which overrides the level defined in the Auth decorator on the Controller in order to have different level of access per Controller
- Refactor Auth decorator in order for the above change to work without re-applying extra decorators

ericboucher:
When I tried it, I did not get promoted to "reef_manager" in the DB for some reason.

But I am actually wondering if this is still useful, if a user has a `reef - user` row, we can probably assume that they are an admin this way. And just keep `default` / `super_admin`

Thoughts @avalmas-programize @bencpeters

avalmas-programize:
No I think keeping all three roles is much useful and easier to maintain.
Also there was a bug regarding the promotion of the user to reef_manager.
Saves after the update operation for the user promotion were reverting the level of the user to their prior one.
Added a new commit that resolves this.

ericboucher (approved):

packages/api/src/users/users.service.ts
```diff
@@ -75,4 +101,45 @@ export class UsersService {
...
+
+      // eslint-disable-next-line fp/no-mutating-methods
+      newAdministeredReefs.push(reef);
+    });
+
+    const newUser = {
+      id: user.id,
+      administeredReefs: newAdministeredReefs,
+    };
+    return this.usersRepository.save(newUser);
```
ericboucher, line 143: if we pass in an id, should we use .update instead of save?

packages/api/src/users/users.service.ts
```diff
@@ -39,6 +44,14 @@ export class UsersService {
...
+
+    if (priorAccount && !uidExists) {
+      const newUser = await this.migrateUserAssociations(priorAccount);
+      // override adminLevel of priorAccount
+      // eslint-disable-next-line fp/no-mutation
+      priorAccount.adminLevel = newUser.adminLevel;
+    }
+
     const data = {
       ...priorAccount,
```
ericboucher, line 56: should we reverse the order here so that priorAccount info is prioritized over the createUserDto?

packages/api/src/users/users.service.ts
avalmas-programize: Updating ManyToMany relations is not working with `update` only with `save`.
https://github.com/typeorm/typeorm/issues/2821

packages/api/src/users/users.service.ts
avalmas-programize: Do we want to do that?
I mean the createUserDto has the information added from the user in the registration form.

Issue #114, opened 2020-09-03
Clicking on a reef resets the table sort


peterive:
Sort by DHW descending, then click on the first reef. When that happens the table on the right is re-ordered. It should retain the DHW (or other) sort.

PR #149, opened 2020-09-08
Retain table sort

tleraj:
The purpose of this PR is to fix https://github.com/aqualinkorg/aqualink-app/issues/114. The reef table on homepage was implemented using the `material-table` package, which caused issues with the `redux-state`. This implementation uses just custom MUI components

### DIFF
--- a/packages/website/src/routes/Homepage/ReefTable/index.tsx
+++ b/packages/website/src/routes/Homepage/ReefTable/index.tsx
@@ -29,7 +29,7 @@ import {
   // degreeHeatingWeeksCalculator,
 } from "../../../helpers/degreeHeatingWeeks";
 import { alertColorFinder } from "../../../helpers/bleachingAlertIntervals";
-import type { Order } from "./types";
+import type { Order, OrderKeys } from "./types";
 import EnhancedTableHead from "./tableHead";
 
 function descendingComparator<T>(a: T, b: T, orderBy: keyof T) {
@@ -44,13 +44,13 @@ function descendingComparator<T>(a: T, b: T, orderBy: keyof T) {
 
 function getComparator(
   order: Order,
-  orderBy: "locationName" | "temp" | "depth" | "dhw"
+  orderBy: OrderKeys
 ): (
   a: {
-    [key in "locationName" | "temp" | "depth" | "dhw"]: number | string | null;
+    [key in OrderKeys]: number | string | null;
   },
   b: {
-    [key in "locationName" | "temp" | "depth" | "dhw"]: number | string | null;
+    [key in OrderKeys]: number | string | null;
   }
 ) => number {
   return order === "desc"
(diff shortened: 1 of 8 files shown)


packages/website/src/routes/Homepage/ReefTable/index.tsx
```diff
@@ -173,36 +123,82 @@ const ReefTable = ({ openDrawer }: ReefTableProps) => {
...
+                          reef.tableData.id === selectedRow
+                            ? colors.lighterBlue
+                            : "white",
+                        cursor: "pointer",
+                      }}
+                      onClick={(event) => handleClick(event, reef)}
+                      role="button"
+                      tabIndex={-1}
+                      // eslint-disable-next-line react/no-array-index-key
+                      key={`${reef.locationName}-${index}`}
```
bencpeters, line 152: This should probably be `reef.id`. Using the index like this is potentially problematic, since it is theoretically possible to get conflicting/duplicate results when we flip the sort order (unlikely, but possible)

packages/website/src/routes/Homepage/ReefTable/tableHead.tsx
```diff
@@ -0,0 +1,114 @@
...
+import {
+  TableHead,
+  TableRow,
+  TableCell,
+  TableSortLabel,
+  Typography,
+} from "@material-ui/core";
+import type { Order, OrderKeys } from "./types";
+
+const columnTitle = (title: string, unit?: string) => (
```
bencpeters, line 11: Let's make this a proper React component:

```typescript
const ColumnTitle = ({ title, unit }) => (
  ...
)
```

packages/website/src/routes/Homepage/ReefTable/tableHead.tsx
```diff
@@ -0,0 +1,114 @@
...
+  <>
```
bencpeters, line 12: No reason to wrap this in a Fragment, it's a single element.

packages/website/src/routes/Homepage/ReefTable/tableHead.tsx
```diff
@@ -0,0 +1,114 @@
...
+          >
+            <TableSortLabel
+              active={props.orderBy === headCell.id}
+              direction={props.orderBy === headCell.id ? props.order : "asc"}
+              onClick={
+                headCell.id !== "alert"
+                  ? createSortHandler(headCell.id)
+                  : () => {}
+              }
+            >
```
bencpeters, line 87: Could we make `TableSortLabel` conditional on `headCell.id !== "alert"` instead of having a dummy click handler? I don't want it to seem like we have a sortable column that just has nothing happen when you click...

bencpeters (changes requested):
Good work! A few clean-up suggestions, but it looks good overall.

Is this PR intended to merge to `master`?

bencpeters:
Closing for now in favor of #152.

PR #152, opened 2020-09-09
Update reef table to use Material UI table

### LINKED ISSUE #114: Clicking on a reef resets the table sort

bencpeters:
This PR switches the reef table to Material UI's built-in table and separates different state dependencies to minimize reloading.

This PR replaces #149. It fixes #114.

### DIFF
--- a/packages/website/src/store/Homepage/types.ts
+++ b/packages/website/src/store/Homepage/types.ts
@@ -6,7 +6,7 @@ export interface HomePageState {
 
 export interface TableRow {
   locationName: string | null;
-  temp: string | 0;
+  temp: number | 0;
   depth: number | null;
   dhw: number | null;
   tableData: {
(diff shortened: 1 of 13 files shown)


packages/website/src/store/Homepage/types.ts
```diff
@@ -3,3 +3,13 @@ import { Reef } from "../Reefs/types";
 export interface HomePageState {
   reefOnMap: Reef | null;
 }
+
+export interface TableRow {
+  locationName: string | null;
+  temp: number | 0;
```
ericboucher, line 9: I think the temp default should be null and displayed as `- -` if possible as we do in other parts of the app

ericboucher (approved):

Issue #142, opened 2020-09-07
Add a delete survey button


ericboucher:
For users with manager access, allow them to delete surveys from the timeline. Maybe a simple button next to "View Survey": "Delete".

And ask for confirmation: "Are you sure you would like to delete the survey for `date`"

PR #154, opened 2020-09-09
Delete survey button

tleraj:
This PR refers to https://github.com/aqualinkorg/aqualink-app/issues/142. The designs implemented look like this:

Screenshot from 2020-09-09 18-22-16
Screenshot from 2020-09-09 18-22-06


ericboucher (changes requested):
This looks like a complete overkill to me. We should be able to do this in just a few lines of code with https://material-ui.com/components/dialogs/

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
```diff
@@ -169,6 +176,20 @@ const SurveyTimeline = ({ addNew, reefId, classes }: SurveyTimelineProps) => {
                           </Link>
                         </Grid>
                       </Grid>
+                      <Grid
```
ericboucher, line 179: Let's make sure that only reef admins can see it?

packages/website/src/routes/ReefRoutes/Reef/Surveys/DeleteButton.tsx
```diff
@@ -0,0 +1,145 @@
...
+      </Dialog>
+    </>
+  );
+};
+
+const styles = () =>
+  createStyles({
+    dialogTitle: {
+      backgroundColor: "#F4F4F4",
+      borderBottom: "1px solid black",
```
ericboucher, line 130: I don't like this border

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
tleraj: The button's visibility is set inside the component. Now it's only visible for `super_admin` or `reef_manager`.

ericboucher: Why set it inside the component? I think it probably makes more sense within the Surveys? Maybe set an `isAdmin` flag that we can use to show/hide the create survey option and the delete options?

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
```diff
@@ -169,6 +176,22 @@ const SurveyTimeline = ({ addNew, reefId, classes }: SurveyTimelineProps) => {
...
+                        className={classes.buttonContainer}
+                        container
+                        alignContent="flex-end"
+                        justify="flex-end"
+                        item
+                        xs={1}
+                      >
+                        {isAdmin && (
```
bencpeters, line 187: This check should be one level higher - there's no reason to render the `Grid.buttonContainer` if `isAdmin` is `false`...

packages/website/src/routes/ReefRoutes/Reef/Surveys/TimelineMobile/index.tsx
```diff
@@ -171,6 +179,22 @@ const TimelineMobile = ({ reefId, addNew, classes }: TimelineMobileProps) => {
...
                   </Grid>
+                  <Grid
+                    className={classes.buttonContainer}
+                    container
+                    alignContent="flex-end"
+                    justify="flex-end"
+                    item
+                    xs={1}
+                  >
+                    {isAdmin && (
```
bencpeters, line 190: This check should be one level higher

bencpeters (changes requested):
Minor clean-ups. I'm also wondering what the purpose of passing in `isAdmin` from the higher component is? It seems like all that accomplishes is establish a new redux dependency (on user) in a higher level component, which is going to mean that the whole component (`ReefDetails` or `NewSurvey`) re-renders on every change to the user redux slice. It may not matter _that_ much in this case since I don't know that the state changes much, but I don't see any reason to hoist that dependency "higher" in the tree than it needs to be here?

packages/website/src/routes/ReefRoutes/Reef/Surveys/TimelineMobile/index.tsx
ericboucher: @tleraj Ben means that ths isAdmin should be set on the grid rather than the button

ericboucher (approved):

PR #155, opened 2020-09-09
Use latest featured image in reef details

### LINKED ISSUE #131: Map is too big and change featured video to featured image
Screen Shot 2020-09-04 at 10 38 11 AM

The triple wide map is just too big. Let's make it back to double-wide and make it much smaller, similar to the cards below it. The featured video can be changed to featured image.

ericboucher:
Fixes #131


bencpeters (changes requested):
This mostly looks good, I think it'd just be nice to clean up the media card layout a bit...

PR #69, opened 2020-08-13
[DO NOT MERGE] Surveys

tleraj:
This PR concerns the surveys UI functionality. The designs implemented look like this:
survey-1
survey-2

PR #157, opened 2020-09-09
Fix comments on surveys PR

tleraj:
The purpose of this PR is to address all the changes requested on https://github.com/aqualinkorg/aqualink-app/pull/69.


packages/website/src/routes/Homepage/RegisterDialog/index.tsx
```diff
@@ -35,6 +34,7 @@ import {
   userErrorSelector,
 } from "../../../store/User/userSlice";
 import { UserRegisterParams } from "../../../store/User/types";
+import incomingStyles from "../styles";
```
mamadOuologuem, line 37: To handle shared styles, we may want to use [nested theme](https://material-ui.com/styles/advanced/) instead.

The overall goal is to avoid component per component styling as much as we can in the app. We should be able to define once our theme in (packages/website/src/layout/App/theme.ts) and then let Material UI works for us.

Something time, we will need a specific styling for a given part of the app. For that, we can use a nested theme that will inherit the main app theme and overrides what is needed.

Let me know if you think it will be cleaner if I open an issue about it or if you want to have a look to that into this PR

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
```diff
@@ -76,125 +77,123 @@ const SurveyTimeline = ({ isAdmin, reefId, classes }: SurveyTimelineProps) => {
...
-                <Grid container item xs={12}>
-                  <Paper elevation={0} className={classes.surveyCard}>
+                <Paper elevation={0} className={classes.surveyCard}>
+                  <Grid
+                    style={{ height: "100%" }}
+                    container
+                    alignItems="center"
+                    justify="space-between"
+                    item
+                    xs={12}
```
mamadOuologuem, line 87: I shouldn't have `item` either `xs={12}` passed as props to a `Grid` when this `Grid` is not contained in a `<Grid container>`

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
```diff
@@ -76,125 +77,123 @@ const SurveyTimeline = ({ isAdmin, reefId, classes }: SurveyTimelineProps) => {
...
+                  >
+                    <Grid style={{ height: "100%" }} item xs={4}>
```
mamadOuologuem, line 89: `style={{ height: "100%" }}` can be removed here.

To stretch the `<Grid item>`s, you can replace in the parent `Grid` `alignItems="center"` by `alignItems="stretch"`

You may want to test if ofter elements in the `<Grid container>` are rendering as expected

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
```diff
@@ -76,125 +77,123 @@ const SurveyTimeline = ({ isAdmin, reefId, classes }: SurveyTimelineProps) => {
...
+                      </Grid>
                     </Grid>
-                  </Paper>
-                </Grid>
+                    {isAdmin && (
+                      <Grid
+                        className={classes.buttonContainer}
+                        container
+                        alignContent="flex-end"
+                        justify="flex-end"
```
mamadOuologuem, line 184: Those props should be removed.

You must not use those props when there is not `<Grid item>` as a child.
If you want to add this style to this element
```
  display: "flex",
  alignItems="flex-end",
  justifyContent="flex-end",
``` 
You can do it directly from `classes.buttonContainer`

PS: Where you intendly using `alignContent="flex-end"`? or you mean `alignItems="flex-end"`

packages/website/src/routes/ReefRoutes/Reef/Surveys/TimelineMobile/index.tsx
```diff
@@ -179,22 +180,22 @@ const TimelineMobile = ({ reefId, isAdmin, classes }: TimelineMobileProps) => {
...
-                    item
-                    xs={1}
-                  >
-                    {isAdmin && (
+                  {isAdmin && (
+                    <Grid
+                      className={classes.buttonContainer}
+                      container
+                      alignContent="flex-end"
+                      justify="flex-end"
```
mamadOuologuem, line 188: same comment as before

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
tleraj: That would result into this:

Screenshot from 2020-09-11 12-49-35

meaning that both the photo and the info would be aligned on the top. In this case I want the info to be centrally aligned, that's why I used this approach. If you strongly disagree with that then I will fix it (perhaps by creating a class `cardImageWrapper` and pass `height: "100%"`).

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
tleraj: I guess you're right. The purpose of this was to place the button on the bottom-right corner of the container. I think it is a better practice to handle it on the `buttonContainer`. Fixed it.

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
mamadOuologuem: Oh okay, 
Lets keep it as you did it then.

mamadOuologuem (approved):
Looks good to me

PR #162, opened 2020-09-11
List administered reefs

ericboucher:
Screen Shot 2020-09-11 at 1 41 56 AM


packages/website/src/common/NavBar/index.tsx
```diff
@@ -115,7 +116,22 @@ const NavBar = ({ searchLocation, classes }: NavBarProps) => {
                       open={Boolean(anchorEl)}
                       onClose={handleMenuClose}
                     >
+                      {user.administeredReefs?.map((reef, index) => {
```
mamadOuologuem, line 119: Maybe worth destructuring reef
```suggestion
                      {user.administeredReefs?.map(({ name, region, id }, index) => {
```

mamadOuologuem (approved):
Looks good to me

PR #189, opened 2020-09-11
Improve daily updates and ensure they happen in utc time

ericboucher:


packages/api/src/utils/maths.ts
```diff
@@ -0,0 +1,15 @@
+import { sum } from 'lodash';
```
bencpeters, line 1: pedantic, but I'd call this `math` (singular) in American English. The Brits like plural "maths" ;-)

packages/api/src/utils/maths.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const getAverage = (numbers: number[], round = false) => {
+  const average =
+    numbers.length > 0 ? sum(numbers) / numbers.length : undefined;
+  return average !== undefined && round ? Math.round(average) : average;
```
bencpeters, line 6: This might be a slightly awkward utility function to use with Typescript - the type of the return value is going to `boolean | number`, which means that you'll basically want to have a boolean check guard any use in a formula. That _might_ be what you want to do, but another (perhaps better?) approach would be to use NaN instead, which would simplify the function:

```
const avg = sum(numbers) / numbers.length;
return round ? Math.round(avg) : avg
```

`NaN` is "safe" to use in computations but it _does_ "taint" everything it touches (when used in computation the results also become `NaN`). But you can also do an explicit `isNaN` check to replicate the functionality of the mixed return type function here.

bencpeters (approved):

packages/api/src/utils/maths.ts
ericboucher: It's actually returning a `number | undefined`. The idea is to avoid populating the DB with NaNs so I made all these utils function return `number | undefined`

PR #190, opened 2020-09-11
Disable videos

ericboucher:


packages/website/src/routes/Surveys/New/UploadMedia.tsx
```diff
@@ -294,8 +294,8 @@ const UploadMedia = ({
               "image/png",
               "image/jpeg",
               "image/gif",
-              "video/mp4",
-              "video/mpeg",
+              // "video/mp4",
+              // "video/mpeg",
             ]}
```
bencpeters, line 299: I'd probably remove these instead of comment out, but at the very least add a note about why they're commented out.

bencpeters (approved):
I _really_ don't approve of commenting out code to disable things - that's the whole purpose of git. It's very easy to look up old code and not have commented code all over the place. However, if you feel that this is actually the best way to do it, at least add a comment explaining why this code is commented out.

ericboucher:
Agreed it's cleaner this way, esp. since this should be short-lived so the memory of this PR will be fresh

PR #196, opened 2020-09-13
Remove icon from map pins

ericboucher:
This is what removing the sat logo looks like. Let me know what you think @drewjgray @peterive @ewolfer67 

Screen Shot 2020-09-13 at 5 23 04 PM


drewjgray:
I like it 👍

Issue #173, opened 2020-09-11
Survey history filters are not populated on reef detail


peterive:
The filters for survey history and observation are not populated on 
https://os-aqualink.web.app/reefs/2941

PR #207, opened 2020-09-14
Activate surveys filters

tleraj:
The purpose of this PR is to implement survey filtering based on survey point and observation (https://github.com/aqualinkorg/aqualink-app/issues/173).

- [x] filter based on observations
- [x] filter based on survey point


packages/api/src/surveys/surveys.service.ts
```diff
@@ -326,4 +359,13 @@ export class SurveysService {
     const trimmedComments = comments.trim();
     return trimmedComments === '' ? undefined : trimmedComments;
   }
+
+  private groupBySurveyId(object: any[], key: string) {
```
ericboucher, line 363: We can do better than any here @avalmas-programize

ericboucher (approved):

PR #212, opened 2020-09-14
Use radio buttons for tile layers

ericboucher:
Screen Shot 2020-09-14 at 7 30 40 PM


drewjgray:
Instead of No Overlay, let's just call it Satellite. Besides that looks great.

drewjgray (changes requested):
Instead of No Overlay, let's just call it Satellite. Besides that looks great.

PR #216, opened 2020-09-15
Calculate better yAxisMin and yAxisMax for temperature charts

ericboucher:


packages/website/src/routes/ReefRoutes/Reef/Charts/utils.ts
```diff
@@ -51,9 +54,32 @@ export const calculateAxisLimits = (dailyData: Data[]) => {
...
   ];
 
+  const { bottomTemperatureData, surfaceTemperatureData } = createDatasets(
+    dailyData
+  );
+
+  const temperatureData = [
+    ...bottomTemperatureData,
+    ...surfaceTemperatureData,
+  ].filter((value) => value);
```
SimplyBallistic, line 64: Would 0 be a valid answer here? If so an explicit null check might be good

packages/website/src/routes/ReefRoutes/Reef/Charts/utils.ts
```diff
@@ -51,9 +54,32 @@ export const calculateAxisLimits = (dailyData: Data[]) => {
...
+
+  const yAxisMinTemp = Math.min(...temperatureData) - 2;
+
+  const yAxisMaxTemp = Math.max(...temperatureData) + 2;
+
+  const yAxisMin = temperatureThreshold
+    ? Math.min(yAxisMinTemp, Math.round(temperatureThreshold) - 5)
```
SimplyBallistic, line 71: there's inconsistent rounding happening on yAxisMinTemp and MaxTemp based off whether temperatureThreshold exists or not. Was this intended?

packages/website/src/routes/ReefRoutes/Reef/Charts/Tooltip/index.tsx
```diff
@@ -132,7 +132,7 @@ const styles = () =>
 export interface TooltipData {
   date: string;
   depth: number | null;
-  bottomTemperature: number;
+  bottomTemperature: number | null;
```
SimplyBallistic, line 135: While working on refactoring Charts, I noticed this is done quite often instead of something like `bottomTemperature?: number`. Is this so users are reminded of optional fields by setting `null` explicitly? Otherwise using the above pattern might be better. This is just personal preference - interested to learn the patterns of this project!

ericboucher: This might be better yes. I think we're doing this in a few spots for simplicity because the backend is returning null values (and not undefined)

packages/website/src/routes/ReefRoutes/Reef/Charts/utils.ts
```diff
@@ -67,13 +67,17 @@ export const calculateAxisLimits = (
 
   const yAxisMaxTemp = Math.max(...temperatureData) + 2;
 
-  const yAxisMin = temperatureThreshold
-    ? Math.min(yAxisMinTemp, Math.round(temperatureThreshold) - 5)
-    : Math.round(yAxisMinTemp);
+  const yAxisMin = Math.round(
+    temperatureThreshold
+      ? Math.min(yAxisMinTemp, temperatureThreshold - 5)
+      : Math.round(yAxisMinTemp)
```
SimplyBallistic, line 74: Not a big deal, but the Math.round on the third part of the ternary could be left out. Otherwise all looks good!

SimplyBallistic (approved):

packages/website/src/routes/ReefRoutes/Reef/Charts/utils.ts
ericboucher: Good catch!

PR #217, opened 2020-09-15
Refactor survey details endpoint

avalmas-programize:
Return the entire list of survey media along with their respective points of interest


packages/website/src/routes/Surveys/View/SurveyMediaDetails.tsx
```diff
@@ -40,13 +41,22 @@ const carouselSettings = {
   ],
 };
 
-const SurveyMediaDetails = ({ points, classes }: SurveyMediaDetailsProps) => {
+const SurveyMediaDetails = ({
+  surveyMedia,
+  classes,
+}: SurveyMediaDetailsProps) => {
+  if (!surveyMedia) {
```
tleraj, line 48: We could get rid of that if statement and in the return do something like:
```
{surveyMedia &&
  getSurveyPointsByName(surveyMedia).map((point) => {
    ...
  })}
```

tleraj (approved):

PR #218, opened 2020-09-15
Creating Common Chart Component to reduce duplication

### LINKED ISSUE #140: DRY chart components
We currently have three independent chart components for
1. featured reef
2. reef details page
3. survey details

90% of the code is duplicated across these components. Let's create one reusable and configurable chart component.

SimplyBallistic:
Closes #140 
Deployed: http://aqua-charts.surge.sh - make sure to test the charts function exactly like master

Components which will be cut down in size from the planned component: 
[ReefRoutes/Reef/Charts](https://github.com/aqualinkorg/aqualink-app/blob/c005a5821b7934e1a47d88ca5c3e91cfe3a629f2/packages/website/src/routes/ReefRoutes/Reef/Charts/index.tsx)
[Surveys/View/Charts](
https://github.com/aqualinkorg/aqualink-app/blob/c005a5821b7934e1a47d88ca5c3e91cfe3a629f2/packages/website/src/routes/Surveys/View/Charts/index.tsx)
[SelectedReefCard/cardChart](https://github.com/aqualinkorg/aqualink-app/blob/c005a5821b7934e1a47d88ca5c3e91cfe3a629f2/packages/website/src/routes/Homepage/ReefTable/SelectedReefCard/cardChart.tsx)

New files:
- Chart/index.tsx (an extension of cardChart with some abstraction)
- Chart/ChartWithTooltip.tsx (Extends above as a HOC adding tooltips)
- Chart/useProcessedChartData.ts (custom hook that loads all chart data like axis limits and datasets. Made to reduce duplication between the above 2 components)


SimplyBallistic:
Issue: `merge` from lodash doesn't merge arrays, but just replaces indicies. How would we make it so that arrays are added together? (Link to conflicting code will be here soon)

SimplyBallistic:
Currently tooltips don't work. Not exactly sure why. Will take a bit of effort to relook everything to work out what settings didn't pass to this new system correctly

packages/website/src/routes/Homepage/ReefTable/SelectedReefCard/cardChart.tsx
```diff
@@ -1,160 +1,11 @@
...
-            },
-          ],
-        },
-      }}
-      height={100}
-      data={createChartData(chartLabels, surfaceTemperatureData, true)}
-    />
-  );
+const CardChart = ({ ...rest }: CardChartProps) => {
+  // TODO temp until its confirmed this component is no longer needed.
```
SimplyBallistic, line 7: This file could be removed. Also, noticed annotations weren't done correctly. This was fixed in this PR.

ericboucher:
Great start!

I noticed a few things on the commit: 
- the dates get hidden on the x.
- the orange coloring is acting weird and did not appear properly before resizing the window.

Screen Shot 2020-09-21 at 12 52 16 PM

Screen Shot 2020-09-21 at 12 52 56 PM

ericboucher:
Can we look into removing the additional points I mentioned last time (the mean temp at the beginning and the end of the graph)

1. I am not sure why we need those in the first place. @tleraj can you maybe give use some insights?
2. If we do need those, let's make sure they don't have a tooltip. It could be as simple as using: `if (index > -1 && index !== 0 && index !== chartLabels.length - 1)` when setting the tooltip.

SimplyBallistic:
> Great start!
> 
> I noticed a few things on the commit:
> 
> * the dates get hidden on the x.
> * the orange coloring is acting weird and did not appear properly before resizing the window.
 
Both are now fixed!

packages/website/src/common/Chart/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+}: ChartProps) {
+  const chartRef = useRef<Line>(null);
+
+  if (forwardRef) {
+    // this might be doable with forwardRef or callbacks, but its a little hard since we need to
+    // use it in two places
+    // eslint-disable-next-line no-param-reassign
+    forwardRef.current = chartRef.current;
+  }
+  const [updateChart, setUpdateChart] = useState<boolean>(true);
```
SimplyBallistic, line 44: I had to flip this to true (it was previously false) to fix the orange fill not appearing until you resize. I couldn't work out what it does exactly. Will this cause an issue down the line? (the boolean value is used in `fillPlugin`

SimplyBallistic:
> Can we look into removing the additional points I mentioned last time **(the mean temp at the beginning and the end of the graph)**

What exactly are these? A screenshot would be great!

tleraj:
I added two dummy points at the graph in order to make it continuous: one at the left (start) and one at the right (end), both equal to the graph's average. Without those points, the graph would start a little after the left edge and end a little before the right edge. The problem with the tooltip being shown can be seen in this screenshot.

Screenshot from 2020-09-22 15-55-20

packages/website/src/common/Chart/Tooltip/index.test.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { render } from "@testing-library/react";
 
-import Tooltip from ".";
+import Tooltip from "./index";
```
SimplyBallistic, line 4: * replace with `.`

SimplyBallistic, line 4:
```suggestion
import Tooltip from ".";
```

packages/website/src/common/Chart/index.tsx
ericboucher: @tleraj any thoughts? do we actually need this?

ericboucher (approved):

PR #221, opened 2020-09-15
Adjust timeline spacing

### LINKED ISSUE #186: Survey history and reef detail page should be center-aligned, as shown in the designs (attached).
Designs:
Screen Shot 2020-09-11 at 9 07 33 AM

Page:
Screen Shot 2020-09-11 at 9 07 11 AM

ericboucher:
Fixes #186 

Screen Shot 2020-09-15 at 7 50 32 PM


packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
```diff
@@ -45,7 +45,11 @@ const SurveyTimeline = ({ isAdmin, reefId, classes }: SurveyTimelineProps) => {
       <Timeline>
         {isAdmin &&
           !(window && window.location.pathname.includes("new_survey")) && (
-            <TimelineItem>
+            <TimelineItem className={classes.timelineItem}>
+              <TimelineOppositeContent
+                className={classes.timelineOppositeContent}
+                style={{ padding: "0 10px" }}
```
SimplyBallistic, line 51: I feel like this could be moved into the corresponding class since you have it set already. Otherwise, looks good!

SimplyBallistic (approved):

packages/website/src/routes/ReefRoutes/Reef/Surveys/Timeline/index.tsx
ericboucher: I added a comment but this one needs to be a bit different than the other ones because we want to center the + sign for adding surveys. I thought it was simpler than creating another class but maybe there are better options to achieve the correct "center"

tleraj (approved):

drewjgray (approved):

PR #223, opened 2020-09-16
Hacky fix to get user data on sign-in

### LINKED ISSUE #222: User info is not loaded right after LOGIN

ericboucher:
The problem we have is that getSelf is called on `auth` change. But during sign-in, the auth change happens at the very beginning, and getSelf finishes before the signIn is over and the data from the signIn chunck ends up overriding the data we just got.

So as a hacky way to fix #222 I fetch all the data during the signIn chunck, similar to what we do during the create chunck.

There is probably a better way to handle this, so I am open to suggestions.

### DIFF
--- a/packages/website/src/store/User/userSlice.ts
+++ b/packages/website/src/store/User/userSlice.ts
@@ -38,6 +38,7 @@ export const createUser = createAsyncThunk<
       return {
         email: data.email,
         fullName: data.fullName,
+        adminLevel: data.adminLevel,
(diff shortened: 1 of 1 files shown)


tleraj (approved):
A I understand the Sign In thunk now awaits the `userServices.getSelf(...)`, preventing the issue of signing in without the necessary backend data (`adminLevel`, `administeredReefs`, ...). It seems alright with me.

PR #225, opened 2020-09-16
Improve bleaching alert

ericboucher:
Update the alertFinder functions to make sure we follow NOAA definitions as closely as possible

https://coralreefwatch.noaa.gov/subscriptions/vs.php

### DIFF
--- a/packages/website/src/helpers/bleachingAlertIntervals.ts
+++ b/packages/website/src/helpers/bleachingAlertIntervals.ts
@@ -21,21 +21,17 @@ const findInterval = (
   satelliteTemperature: number | null,
   degreeHeatingWeeks?: number | null
 ): Interval => {
-  if (
+  const hotSpot =
     satelliteTemperature &&
     maxMonthlyMean &&
-    satelliteTemperature < maxMonthlyMean
-  ) {
+    satelliteTemperature - maxMonthlyMean;
+  if (hotSpot && hotSpot < 0) {
     return {
       image: noStress,
       color: "#C6E5FA",
     };
   }
-  if (
-    satelliteTemperature &&
-    maxMonthlyMean &&
-    satelliteTemperature < maxMonthlyMean + 1
-  ) {
+  if (hotSpot && hotSpot < 1) {
     return {
       image: watch,
       color: "#FFF200",
--- a/packages/website/src/helpers/bleachingAlertIntervals.ts
+++ b/packages/website/src/helpers/bleachingAlertIntervals.ts
@@ -4,55 +4,83 @@ import watch from "../assets/alert_watch.svg";
 import lvl1 from "../assets/alert_lvl1.svg";
 import lvl2 from "../assets/alert_lvl2.svg";
 
-const bleachingAlertIntervals = [
-  {
-    max: 5,
-    image: noStress,
-    color: "#C6E5FA",
-  },
-  {
-    max: 10,
-    image: warning,
-    color: "#FFF200",
-  },
-  {
-    max: 15,
-    image: watch,
-    color: "#F8AB00",
-  },
-  {
-    max: 20,
-    image: lvl1,
-    color: "#EF0000",
-  },
-  {
-    max: Infinity,
-    image: lvl2,
-    color: "#940000",
-  },
-];
-
 type Interval = {
-  max: number;
   image: string;
   color: string;
 };
 
-const findInterval = (degreeHeatingWeeks?: number | null): Interval => {
-  const interval = bleachingAlertIntervals.find(
-    (item) => (degreeHeatingWeeks || 0) < item.max
-  );
-  return interval || bleachingAlertIntervals[0];
+/**
+ * Calculating bleaching alert level based on NOAA defintions:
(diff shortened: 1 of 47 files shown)


packages/website/src/helpers/bleachingAlertIntervals.ts
```diff
@@ -4,55 +4,79 @@ import watch from "../assets/alert_watch.svg";
...
+const findInterval = (
+  maxMonthlyMean: number | null,
+  satelliteTemperature: number | null,
+  degreeHeatingWeeks?: number | null
+): Interval => {
+  const hotSpot =
+    satelliteTemperature &&
+    maxMonthlyMean &&
+    satelliteTemperature - maxMonthlyMean;
+  if (hotSpot && hotSpot < 0) {
```
SimplyBallistic, line 28: These if statements look a little unusual. Isn't there a better way to lay this out?
For example, comparisons with constants `0 < 4` doesn't look right.

With a quick glance at the link mentioned, using this: https://lodash.com/docs/4.17.15#inRange
might help simplify things

packages/website/src/helpers/bleachingAlertIntervals.ts
```diff
@@ -4,55 +4,79 @@ import watch from "../assets/alert_watch.svg";
...
+ * @param satelliteTemperature
+ * @param degreeHeatingWeeks
+ */
+const findInterval = (
+  maxMonthlyMean: number | null,
+  satelliteTemperature: number | null,
+  degreeHeatingWeeks?: number | null
+): Interval => {
+  const hotSpot =
+    satelliteTemperature &&
```
SimplyBallistic, line 25: If any of these were 0, should it default to 0?

SimplyBallistic:
here's a suggestion on how the if statements could be simplified:
```js
import { inRange, isNil } from 'lodash';

let hotSpot = 1;
let dhw = 1;

switch (true) {
    case isNil(hotSpot) || isNil(dhw):
        console.log('nil case');
        break;

    case hotSpot <= 0:
        console.log('no bleaching');
        break;
    // >0 past this point
    case hotSpot < 1:
        console.log('watch');
        break;
    // >=1 past this point
    // start dhw checks
    // 0 < dhw < 4
    case 0 < dhw && dhw < 4:
        console.log('possible bleaching');
        break;
    // 4 <= dhw < 8
    case inRange(dhw, 4, 8):
        console.log('bleaching likely');
        break;
    case dhw >= 8:
        console.log('mortality likely');
        break;
    default:
        // dhw is likely <= 0
        console.log('default case')


}
```

packages/website/src/helpers/bleachingAlertIntervals.ts
ericboucher: The Nil case for temp and maxMonthlyMean is a tricky one, because it likely means we are missing info. My thinking was that we should therefore bypass the hotSpot test (because it is invalid) and use DHW directly.

packages/website/src/helpers/bleachingAlertIntervals.ts
```diff
@@ -25,40 +26,51 @@ const findInterval = (
...
+      };
+
+    // Hotspot >=1 or nil past this point, start dhw checks.
+    case isNil(degreeHeatingWeeks):
+      return {
+        image: noStress,
+        color: "#C6E5FA",
+      };
+
+    case inRange(degreeHeatingWeeks!, 0, 4):
```
SimplyBallistic, line 50: According to https://coralreefwatch.noaa.gov/subscriptions/vs.php, warning applies if dhw is > 0 (exclusive), but unfortunately `inRange` is arg1 (0) inclusive to arg2 (4) exclusive. Not sure if this is a potentially dangerous case, but could be fixed with a manual check like what I did with my comment

packages/website/src/helpers/bleachingAlertIntervals.ts
```diff
@@ -25,40 +26,51 @@ const findInterval = (
...
+        color: "#F8AB00",
+      };
+
+    case inRange(degreeHeatingWeeks!, 4, 8):
+      return {
+        image: lvl1,
+        color: "#EF0000",
+      };
+
+    case degreeHeatingWeeks! > 8:
```
SimplyBallistic, line 62: If dhw is 8, I think it'll jump to default. According to the specification I think it should be >=8

packages/website/src/helpers/bleachingAlertIntervals.ts
```diff
@@ -25,40 +26,51 @@ const findInterval = (
...
+      return {
+        image: lvl2,
+        color: "#940000",
+      };
+
+    default:
```
SimplyBallistic, line 68: Could add a comment to explain why default might get called. I think its if dhw <=0, if the above comments are addressed.

packages/website/src/helpers/bleachingAlertIntervals.ts
ericboucher: Yeah it's not clear what's supposed to happen for 1 <= HotSpot and DHW === 0 ...

drewjgray (approved):

SimplyBallistic (approved):

PR #227, opened 2020-09-17
Add live data API endpoint

avalmas-programize:
- Add GET /reefs/<reef_id>/live_data: Fetch live data for reef with id <reef_id>
- Add liveData.ts in utils to handle the fetching of the latest data from sofar
- Refactor dailyData.ts to remove constant string (modelId and variableIDs)
- Refactor sofar.ts so that all data fetching functions return SofarValues so that we have a uniform approach
- expose util funcions (extractSofarValues, getLatestDailyData) as they are needed in liveData.ts and dailyData.ts
- Add sofar.types.ts where public types regarding sofar are stored

### DIFF
--- a/packages/api/src/utils/constants.ts
+++ b/packages/api/src/utils/constants.ts
@@ -1,5 +1,3 @@
-import { SofarModels } from './sofar.types';
-
 // dotenv is a dev dependency, so conditionally import it (don't need it in Prod).
 try {
   // eslint-disable-next-line import/no-extraneous-dependencies, global-require
@@ -14,19 +12,25 @@ export const SOFAR_MARINE_URL =
   'https://api.sofarocean.com/marine-weather/v1/models/';
 export const SOFAR_SPOTTER_URL = 'https://api.sofarocean.com/api/wave-data';
 
+export enum SofarModels {
+  NOAAOperationalWaveModel = 'NOAAOperationalWaveModel',
+  NOAACoralReefWatch = 'NOAACoralReefWatch',
+  GFS = 'GFS',
+}
+
 // Sofar variables
 export const sofarVariableIDs = {
   [SofarModels.NOAAOperationalWaveModel]: {
-    significantWaveHeight: 'NOAAOperationalWaveModel-significantWaveHeight',
-    meanDirectionWindWaves: 'NOAAOperationalWaveModel-meanDirectionWindWaves',
-    peakPeriod: 'NOAAOperationalWaveModel-peakPeriod',
+    significantWaveHeight: 'significantWaveHeight',
+    meanDirectionWindWaves: 'meanDirectionWindWaves',
+    peakPeriod: 'peakPeriod',
   },
   [SofarModels.NOAACoralReefWatch]: {
     degreeHeatingWeek: 'degreeHeatingWeek',
     analysedSeaSurfaceTemperature: 'analysedSeaSurfaceTemperature',
   },
   [SofarModels.GFS]: {
-    magnitude10MeterWind: 'GFS-magnitude10MeterWind',
-    direction10MeterWind: 'GFS-direction10MeterWind',
+    magnitude10MeterWind: 'magnitude10MeterWind',
+    direction10MeterWind: 'direction10MeterWind',
   },
 };
--- a/packages/api/src/utils/constants.ts
+++ b/packages/api/src/utils/constants.ts
@@ -1,3 +1,5 @@
+import { SofarModels } from './sofar.types';
+
 // dotenv is a dev dependency, so conditionally import it (don't need it in Prod).
 try {
   // eslint-disable-next-line import/no-extraneous-dependencies, global-require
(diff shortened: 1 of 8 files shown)


packages/api/src/utils/liveData.ts
```diff
@@ -0,0 +1,114 @@
...
+} from './sofar';
+import { SofarLiveData, SofarModels } from './sofar.types';
+import { getDegreeHeatingDays } from '../workers/dailyData';
+
+export const getLiveData = async (reef: Reef): Promise<SofarLiveData> => {
+  const { polygon, spotterId, maxMonthlyMean } = reef;
+  // TODO - Accept Polygon option
+  const [longitude, latitude] = (polygon as Point).coordinates;
+
+  const endOfDate = new Date();
```
ericboucher, line 18: Maybe call it "now" or similar for clarity?

packages/api/src/utils/sofar.ts
```diff
@@ -4,26 +4,25 @@ import { isNil } from 'lodash';
...
+  waveMeanDirection: SofarValue[];
 };
 
 type SensorData = {
   sensorPosition: number;
   degrees: number;
 };
 
-const extractSofarValues = (sofarValues: SofarValue[]): number[] =>
+export const getLatestDailyData = (sofarValues: SofarValue[]): SofarValue =>
```
ericboucher, line 22: They should be ordered by timestamp but it probably makes sense for this function to guarantee it by fetching the latest timestamp, and not just slice(-1)

packages/api/src/utils/liveData.ts
```diff
@@ -0,0 +1,114 @@
...
+  };
+
+  const degreeHeatingDays = await getDegreeHeatingDays(
+    maxMonthlyMean,
+    latitude,
+    longitude,
+    endOfDate,
+  );
+
+  const satelliteTemperature = getLatestDailyData(
```
ericboucher, line 47: Maybe rename function sofarHindcast to have a more streamlined approach.

packages/api/src/utils/sofar.ts
```diff
@@ -4,26 +4,25 @@ import { isNil } from 'lodash';
...
 import { getStartEndDate } from './dates';
 import { SOFAR_MARINE_URL, SOFAR_SPOTTER_URL } from './constants';
-
-type SofarValue = {
-  timestamp: string;
-  value: number;
-};
+import { SofarValue } from './sofar.types';
 
 type SpotterData = {
```
ericboucher, line 9: might be worth moving to sofar types

packages/api/src/utils/constants.ts
```diff
@@ -11,3 +13,20 @@ export const { SOFAR_API_TOKEN } = process.env;
 export const SOFAR_MARINE_URL =
   'https://api.sofarocean.com/marine-weather/v1/models/';
 export const SOFAR_SPOTTER_URL = 'https://api.sofarocean.com/api/wave-data';
+
+// Sofar variables
```
ericboucher, line 17: I am wondering if we could structure models and variables as nested type instead? what do you think?

ericboucher: They are not very different from models from our perspective. The fact that they are called variable is more because of sofar's structure than our own choice / need.

packages/api/src/workers/dailyData.ts
```diff
@@ -6,10 +6,54 @@ import Bluebird from 'bluebird';
...
+      longitude,
+      endOfDate,
+      96,
+    );
+
+    // Check if there are any data returned
+    // Grab the last one and convert it to degreeHeatingDays
+    return degreeHeatingWeek.length > 0
+      ? {
+          value: getLatestDailyData(degreeHeatingWeek).value * 7,
```
ericboucher, line 46: getLatestDailyData -> getLatestData ?

packages/api/src/utils/sofar.ts
```diff
@@ -53,26 +52,60 @@ export async function sofarHindcast(
...
       } else {
-        console.error(`An error occured accessing the Sofar API - ${error}`);
+        console.error(
+          `An error occurred accessing the Sofar Hindcast API - ${error}`,
+        );
+      }
+    });
+}
+
+export async function sofarForecast(
```
ericboucher, line 65: Let's add tests for these new functions

packages/api/src/utils/liveData.ts
avalmas-programize: Ok

packages/api/src/utils/sofar.ts
avalmas-programize: Ok a simple iteration of the array should do the trick

packages/api/src/utils/sofar.ts
avalmas-programize: We don't really need elsewhere but ok

packages/api/src/workers/dailyData.ts
avalmas-programize: Ok

packages/api/src/utils/sofar.ts
avalmas-programize: Tests added

packages/api/src/utils/constants.ts
avalmas-programize: Some more input is needed

packages/api/src/utils/liveData.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  const satelliteTemperature = getLatestData(
+    await getSofarDailyData(
+      SofarModels.NOAACoralReefWatch,
+      sofarVariableIDs[SofarModels.NOAACoralReefWatch]
+        .analysedSeaSurfaceTemperature,
+      latitude,
+      longitude,
+      now,
+      96,
```
ericboucher, line 53: let's lower this to 72

avalmas-programize: Done

packages/api/src/utils/liveData.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  const degreeHeatingDays = await getDegreeHeatingDays(
+    maxMonthlyMean,
+    latitude,
+    longitude,
+    now,
+  );
+
+  const satelliteTemperature = getLatestData(
+    await getSofarDailyData(
```
ericboucher, line 46: Given how we use this function now, we could consider renaming `getSofarDailyData` to `sofarHindcast`?

PR #229, opened 2020-09-18
Map tweaks

tleraj:
The purpose of this PR is to add the features below on the homepage map

 - [x] Color cluster icon based on its inner reefs DHW average
 - [x] Add sea surface temperature legend on the map


packages/website/src/routes/Homepage/Map/Markers/index.tsx
```diff
@@ -74,6 +78,34 @@ const buoyIcon = (colorClass: string) =>
...
+    const { maxMonthlyMean } = reef;
+    const { satelliteTemperature, degreeHeatingDays } = reef.latestDailyData;
+    const degreeHeatingWeeks = degreeHeatingWeeksCalculator(degreeHeatingDays);
+    return findInterval(
+      maxMonthlyMean,
+      satelliteTemperature,
+      degreeHeatingWeeks
+    );
+  });
+  const { color } = alerts.reduce(
```
ericboucher, line 93: I think this logic belongs to the alert file. Maybe we could have a "findMaxLevel" function and a "getColorByLevel"?

packages/website/src/routes/Homepage/Map/Markers/index.tsx
```diff
@@ -74,6 +78,34 @@ const buoyIcon = (colorClass: string) =>
...
+      image: "",
+      color: "#C6E5FA",
+      level: 1,
+    }
+  );
+  const count = cluster.getChildCount();
+  return L.divIcon({
+    html: `<div style="background-color: ${color}"><span>${count}</span></div>`,
+    className: `leaflet-marker-icon marker-cluster custom-cluster-icon marker-cluster-small leaflet-zoom-animated leaflet-interactive`,
+    iconSize: L.point(40, 40, true),
```
ericboucher, line 105: Could we make the size vary slightly with the number of elements? Since we don't have the color as an indicator anymore?

ericboucher:
When clicking on the layer selector, it removes the layer if one was selected. It should not.

ericboucher (approved):

PR #234, opened 2020-09-23
Change position of featured image star icon

tleraj:
The purpose of this PR is to change the position of the star icon and place it on the media card's right. The design implemented looks like this:

Screenshot from 2020-09-23 13-38-52


packages/website/src/routes/Surveys/New/UploadMedia.tsx
```diff
@@ -109,6 +109,9 @@ const UploadMedia = ({
     setPreviews(previews.filter((item, key) => key !== index));
     setFiles(files.filter((item, key) => key !== index));
     setMetadata(metadata.filter((item, key) => key !== index));
+    if (index === featuredFile) {
```
ericboucher, line 112: why this change?

tleraj: If the user deletes the card that was set to featured, then we need to reset the first image as featured. Doesn't this make sense?

ericboucher (approved):

PR #236, opened 2020-09-23
Use UTC times

tleraj:
The purpose of this PR is to set every date appearing in the app to UTC


packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
```diff
@@ -44,7 +44,7 @@ const ReefNavBar = ({
             {lastSurvey && (
               <Grid item>
                 <Typography variant="subtitle1">{`Last surveyed: ${moment(
-                  lastSurvey
+                  new Date(lastSurvey).toISOString()
```
ericboucher, line 47: seems weird to use moment AND Date, no?

tleraj: The use of moment is only for the display format (e.g "MM/DD/YYYY").

ericboucher: Seems like an overkill. How about creating a simple utility function like this? https://stackoverflow.com/a/11172083

tleraj: We could also get rid of the `Date` like this:

```
moment(moment(lastSurvey).toISOString()).format("MMM DD[,] YYYY")
```

Does that sound good?

ericboucher:
We will use local time instead

PR #249, opened 2020-09-24
Refactor application flow

avalmas-programize:
- Add POST reef-application endpoint: Creates virtual reef (pending approval) and reef-application


avalmas-programize:
Added approved field in reefs (Default to true for all reefs). It is set to false for all new reefs.
Prevented users from interacting with these reefs via the API (reefs API, surveys API
Pending reefs will get their daily data updated (no checks have been added)

packages/api/src/reefs/reefs.service.ts
```diff
@@ -66,11 +66,13 @@ export class ReefsService {
       'latestDailyData',
       `(latestDailyData.date, latestDailyData.reef_id) IN (${this.latestDailyDataSubquery()})`,
     );
+    query.andWhere('approved = true');
     return query.getMany();
   }
 
   async findOne(id: number): Promise<Reef> {
     const found = await this.reefsRepository.findOne(id, {
+      where: { approved: true },
```
ericboucher, line 75: If the person knows the id, maybe we can let them access the page?

packages/api/src/reefs/reefs.service.ts
```diff
@@ -80,7 +82,10 @@ export class ReefsService {
   }
 
   async update(id: number, updateReefDto: UpdateReefDto): Promise<Reef> {
-    const result = await this.reefsRepository.update(id, updateReefDto);
+    const result = await this.reefsRepository.update(
```
ericboucher, line 85: why are we changing this?

packages/api/src/reefs/reefs.service.ts
```diff
@@ -66,11 +66,13 @@ export class ReefsService {
       'latestDailyData',
       `(latestDailyData.date, latestDailyData.reef_id) IN (${this.latestDailyDataSubquery()})`,
     );
+    query.andWhere('approved = true');
```
ericboucher, line 69: I feel like this could be the only changed need for the "approved", no? Just limit what reefs appear in the main app, other than that I think it's fine

packages/api/src/reefs/reefs.service.ts
avalmas-programize: Yeah. I just thought they would gain access to those reefs only via the application API. Like GET pending-reefs. But we can reuse those endpoints as well.

packages/api/src/reefs/reefs.service.ts
avalmas-programize: I just thought that we want to restrict completely the access to the pending reefs.  Reverting

packages/api/src/reefs/reefs.service.ts
avalmas-programize: Ok

packages/api/src/reef-applications/reef-applications.service.ts
```diff
@@ -1,38 +1,96 @@
...
+        type: 'Point',
+        coordinates: [latitude, longitude],
+      },
+      maxMonthlyMean,
+      timezone: timezones[0],
+      approved: false,
+      region,
+    });
+
+    // Change user's admin level only if he is not a super admin
```
ericboucher, line 56: use "they/their" instead of he.
also, why do we need this check here? can't we just add a guard to the endpoint?

packages/api/src/reef-applications/reef-applications.service.ts
```diff
@@ -1,38 +1,96 @@
...
+    // Also make him admin of the reef
+    if (user.adminLevel !== AdminLevel.SuperAdmin) {
+      // We need to fetch the user from the database again because the administeredReefs are not loaded
+      const reefAdmin = await this.userRepository.findOne({
```
ericboucher, line 60: Do we really need this check here? can't we just add a guard to the endpoint?

packages/api/src/reef-applications/reef-applications.service.ts
avalmas-programize: Fixed

packages/api/src/reef-applications/reef-applications.service.ts
avalmas-programize: We just want to know if we have to update the reef-admins list and change the user's role to ReefManager. We don't want to change a SuperAdmin to ReefManager just because they filled their application.
It is not a authorization check so a guard wouldn't help.

packages/api/src/reef-applications/reef-applications.service.ts
```diff
@@ -1,38 +1,96 @@
...
+      });
+
+      if (!reefAdmin) {
+        throw new NotFoundException(`User with id ${user.id} was not found.`);
+      }
+
+      await this.userRepository.save({
+        id: reefAdmin.id,
+        adminLevel: AdminLevel.ReefManager,
+        administeredReefs: [...reefAdmin.administeredReefs, reef],
```
ericboucher, line 72: Don't we have a way to "append" a new reef to administeredReefs without fetching the previous ones? 

We just need to add (reef_id, user_id) to `users_administered_reefs_reef`

Also, I think this part should happen even for a super admin

packages/api/src/reef-applications/reef-applications.service.ts
```diff
@@ -1,38 +1,96 @@
...
+    const timezones = getTimezones(latitude, longitude) as string[];
+
+    const reef = await this.reefRepository.save({
+      depth,
+      polygon: {
+        type: 'Point',
+        coordinates: [latitude, longitude],
+      },
+      maxMonthlyMean,
+      timezone: timezones[0],
```
ericboucher, line 51: this could fail if timezones is []

avalmas-programize: No. Javascript does not have ArrayOutOfBoundsException. 
Instead it returns undefined if no element exists in that index.
So it will set timezone to NULL which is okay since the field is nullable.

packages/api/src/reef-applications/reef-applications.service.ts
avalmas-programize: We can't access the `users_administered_reefs_reef` because it is a derived model.
The many to many relation in typeorm is quite tricky but I found a way to do it via the query builder api.
Okay I will let the super admin have entries on `users_administered_reefs_reef` as well

ericboucher (approved):

Issue #240, opened 2020-09-23
The site logo in the top left should link to home page. 


drewjgray:

ericboucher:
@tleraj let's also add a link to the footer logo

PR #254, opened 2020-09-25
Link navbar logo to homepage

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/240.

### DIFF
--- a/packages/website/src/common/Footer/index.tsx
+++ b/packages/website/src/common/Footer/index.tsx
@@ -8,6 +8,7 @@ import {
   WithStyles,
   createStyles,
   Theme,
+  Link,
 } from "@material-ui/core";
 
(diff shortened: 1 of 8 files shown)


packages/website/src/common/Footer/index.tsx
```diff
@@ -34,6 +37,15 @@ const styles = (theme: Theme) =>
         backgroundColor: theme.palette.primary.main,
       },
     },
+    navBarLink: {
+      display: "flex",
+      color: "inherit",
```
ericboucher, line 42: do we need these?

tleraj: `display: "flex"` is needed in order to place "Aqua" and "link" next to each other. Removed the color property.

ericboucher (approved):

Issue #238, opened 2020-09-23
When you click a specific site on the home page, replace the "Featured Site " content with the site details


drewjgray:
If there is a featured image, then display it. If not just expand the time-series chart all the way across.

PR #255, opened 2020-09-25
Homepage featured site

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/238. The designs implemented look like this:

 - Featured reef:
   
   Screenshot from 2020-09-25 15-26-14

 - Reef with no featured media:

   Screenshot from 2020-09-25 15-27-20

 - Reef with featured media:

   Screenshot from 2020-09-25 15-29-01

We need to decide what will happen to the explore button in case of a reef with no featured media, where the graph takes all the remaining space.


packages/website/src/routes/ReefRoutes/Reef/Surveys/index.tsx
```diff
@@ -36,9 +36,15 @@ const Surveys = ({ reefId, classes }: SurveysProps) => {
     : false;
 
   useEffect(() => {
-    reefServices
-      .getReefPois(`${reefId}`)
-      .then((response) => setPointOptions(response.data));
+    let mounted = true;
```
ericboucher, line 39: Why do we need this?

If it's needed, let's try to do it without reassignments

tleraj: Without this, React throws the following warning:
```
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory
leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup
function.
```
which happens because we do a state update inside an asynchronous function with the component not being mounted yet.
Found a way to avoid reassignments, by using `axios.CancelToken`, as proposed [here](https://github.com/axios/axios#cancellation).

ericboucher:
On mobile, after a reef is selected, the arrow to pull the drawer disappears and the reef name in the drawer can be "null"

IMG_4635

tleraj:
The designs now look like this:

 - Reef with image:
   
   Screenshot from 2020-09-30 13-31-56
   Screenshot from 2020-09-30 13-31-41
   Screenshot from 2020-09-30 13-31-25

 - Reef without image:

   Screenshot from 2020-09-30 13-34-28
   Screenshot from 2020-09-30 13-34-15
   Screenshot from 2020-09-30 13-34-02

ericboucher (approved):

PR #261, opened 2020-09-28
Add getReefNameAndRegion  helper

ericboucher:
This helps make sure that we can use the display name for the search tool.

### DIFF
--- a/packages/website/src/common/Search/index.tsx
+++ b/packages/website/src/common/Search/index.tsx
@@ -18,9 +18,22 @@ import { formatReefName } from "../../store/Reefs/helpers";
 const Search = ({ classes }: SearchProps) => {
   const [searchedReef, setSearchedReef] = useState<Reef | null>(null);
   const dispatch = useDispatch();
-  const reefs = useSelector(reefsListSelector).filter((reef) =>
-    formatReefName(reef)
-  );
+  // eslint-disable-next-line fp/no-mutating-methods
+  const reefs = useSelector(reefsListSelector)
+    .filter((reef) => formatReefName(reef))
+    // Sort by formatted name
+    .sort((a, b) => {
+      const nameA = formatReefName(a) || "";
+      const nameB = formatReefName(b) || "";
+
+      if (nameA > nameB) {
+        return 1;
+      }
+      if (nameA < nameB) {
+        return -1;
+      }
+      return 0;
+    });
 
   const onChangeSearchText = (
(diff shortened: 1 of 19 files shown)


packages/website/src/common/Search/index.tsx
```diff
@@ -13,18 +13,34 @@ import Autocomplete from "@material-ui/lab/Autocomplete";
...
 import type { Reef } from "../../store/Reefs/types";
 import { reefsListSelector } from "../../store/Reefs/reefsListSlice";
+import { formatReefName } from "../../store/Reefs/helpers";
 
 const Search = ({ classes }: SearchProps) => {
   const [searchedReef, setSearchedReef] = useState<Reef | null>(null);
   const dispatch = useDispatch();
-  const reefs = useSelector(reefsListSelector).filter(({ name }) => name);
+  const reefs = useSelector(reefsListSelector)
+    .filter((reef) => formatReefName(reef))
```
SimplyBallistic, line 22:
```suggestion
    .filter(formatReefName)
```
I think you can just pass the function reference here if you wish

packages/website/src/store/Reefs/helpers.ts
```diff
@@ -4,6 +4,9 @@ import type { Reef } from "./types";
 import { degreeHeatingWeeksCalculator } from "../../helpers/degreeHeatingWeeks";
 import { alertFinder } from "../../helpers/bleachingAlertIntervals";
 
+export const formatReefName = (reef: Reef) =>
```
SimplyBallistic, line 7: More of a personal opinion, but I feel `format` as a function verb more-so implies the string is edited in some way. Instead, what we're doing here is 'getting' the reef name, checking various sources. Do you think `getReefName` would be better here?

packages/website/src/common/Search/index.tsx
```diff
@@ -13,18 +13,34 @@ import Autocomplete from "@material-ui/lab/Autocomplete";
...
+    // Sort by formatted name
+    .sort((a, b) => {
+      const nameA = formatReefName(a) || "";
+      const nameB = formatReefName(b) || "";
+
+      if (nameA > nameB) {
```
SimplyBallistic, line 28: Using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare might be better here as in the future, special characters could be used if the application is truly international.

tleraj (approved):

SimplyBallistic (commented):
Looks good!
Edit: Didn't realize this was merged

PR #263, opened 2020-09-28
Calculate weekly alert level in daily data

ericboucher:


tleraj (approved):
Frontend changes look good!

packages/api/src/workers/dailyData.ts
```diff
@@ -222,6 +231,37 @@ export async function getDailyData(
...
+    .select('MAX(dailyData.dailyAlertLevel)', 'weeklyAlertLevel')
+    .andWhere('dailyData.date >= :pastWeek', { pastWeek })
+    .andWhere('dailyData.date <= :date', { date })
+    .andWhere('dailyData.reef = :reef', { reef: reef.id })
+    .getRawOne();
+
+  return isNumber(query.weeklyAlertLevel) ? query.weeklyAlertLevel : undefined;
+}
+
+export function mergeDailyAndWeeklyAlertLevel(
```
ericboucher, line 252: Let's simplify this function

packages/api/src/workers/dailyData.ts
```diff
@@ -246,7 +298,13 @@ export async function getReefsDailyData(connection: Connection, date: Date) {
               reef,
               date: entity.date,
             },
-            filteredData,
+            {
+              ...filteredData,
+              ...(isNumber(weeklyAlertLevel) ? { weeklyAlertLevel } : {}),
```
ericboucher, line 303: Isn't this duplicating what omitBy(entity, isNil) is doing? @avalmas-programize

avalmas-programize: No. The omitBy with the isNil as a predicate is omitting all values that are falsy. This means that a 0 alert level is omitted and therefore a weeklyAlertLevel that is supposed to be 0 will never update to zero from NULL because it will be omitted.
So this is an override of this behavior, by using the isNumber after the omitBy.

packages/api/src/workers/dailyData.ts
avalmas-programize: Hmm I am not that good at naming

ericboucher: I meant the process ;)

packages/api/src/workers/dailyData.ts
ericboucher: Hmm not sure about that - https://lodash.com/docs/4.17.15#isNil
But if that's the case we should make the filtering smarter, not implement a case by case check
we also want the temps to allow 0's

ericboucher: There are plenty of 0's in the daily data table

Issue #171, opened 2020-09-11
Allow managers to go back to set featured images


peterive:
Once I've uploaded an image I can't go back and set that image to the featured image from any of the screens

ericboucher:
@peterive for point number one, where would you expect to be able to edit it? do we want to have an edit mode for reef admins?

ewolfer67:
We’re not currently affording the reef manager the ability to fully edit a survey/images right now. The "Survey Detail Page" (if logged in as a Reef Manager) should have the Fav "star" controls displayed in the upper right corner of the image carousels as shown in the page design on Zeplin. 

![Uploading image.png…]()

PR #265, opened 2020-09-29
Allow reef managers to update featured media

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/171. The designs implemented look like this:

Screenshot from 2020-09-29 16-33-20
Screenshot from 2020-09-29 16-33-01


packages/website/src/routes/Surveys/View/index.tsx
```diff
@@ -136,7 +164,11 @@ const SurveyViewPage = ({ reef, surveyId, classes }: SurveyViewPageProps) => {
             </Typography>
           </Grid>
           <Grid style={{ width: "100%" }} item>
-            <SurveyMediaDetails surveyMedia={surveyDetails?.surveyMedia} />
+            <SurveyMediaDetails
```
ericboucher, line 167: If isAdmin is only used in SurveyMediaDetails, then let's move the `const user = useSelector(userInfoSelector);` into `SurveyMediaDetails`

And maybe create a util function `isUserAdmin(user)` and use it where we had a similar isAdmin as well so that it's easier to manage?

ericboucher (changes requested):
See comments inline

packages/website/src/routes/Surveys/View/index.tsx
tleraj: `user` is also needed in `onSurveyMediaUpdate`, so if we moved it into the `SurveyMediaDetails`, then we would have to also move the `onSurveyMediaUpdate` function, along with the `reefId`. Does this sound good?

ericboucher: If it's easy to move `onSurveyMediaUpdate` I would say let's do that. If it's not easy then it's ok to keep it here.

ericboucher (approved):

PR #284, opened 2020-10-05
Add google cloud endpoints to handle the files in the storage

avalmas-programize:
GET /google-cloud/ : Fetch dangling files
DELETE /google-cloud/ : Delete dangling files

### DIFF
--- a/packages/api/src/google-cloud/google-cloud.service.ts
+++ b/packages/api/src/google-cloud/google-cloud.service.ts
@@ -3,20 +3,62 @@ import {
   InternalServerErrorException,
   Logger,
 } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
 import { Storage } from '@google-cloud/storage';
+import { Repository } from 'typeorm';
+import { SurveyMedia } from '../surveys/survey-media.entity';
 
 @Injectable()
 export class GoogleCloudService {
   private logger = new Logger('GCS');
   private storage: Storage;
 
-  constructor() {
+  constructor(
+    @InjectRepository(SurveyMedia)
+    private surveyMediaRepository: Repository<SurveyMedia>,
+  ) {
     this.storage = new Storage({
       autoRetry: true,
       keyFilename: process.env.GCS_KEYFILE,
       projectId: process.env.GCS_PROJECT,
       maxRetries: 3,
     });
+  }
+
+  public async findDanglingFiles() {
+    if (!process.env.GCS_BUCKET || !process.env.STORAGE_FOLDER) {
+      this.logger.error(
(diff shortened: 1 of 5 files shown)


packages/api/src/google-cloud/google-cloud.service.ts
```diff
@@ -19,6 +25,45 @@ export class GoogleCloudService {
...
+      throw new InternalServerErrorException();
+    }
+
+    const surveyMedia = await this.surveyMediaRepository.find({
+      select: ['url'],
+    });
+
+    const mediaSet = new Set(
+      surveyMedia.map(
+        (media) => media.url.split(`${process.env.GCS_BUCKET}/`)[1],
```
ericboucher, line 42: let's create a util to get the file name out of the GCS url

packages/api/src/google-cloud/google-cloud.service.ts
```diff
@@ -19,6 +25,45 @@ export class GoogleCloudService {
     });
   }
 
+  public async findDanglingFiles(): Promise<string[]> {
+    if (!process.env.GCS_BUCKET || !process.env.STORAGE_FOLDER) {
```
ericboucher, line 29: In general, I prefer using process.env at the very top of the file only to create the constants that you will need. This makes it a bit clearer what depends on external configs

ericboucher:
Looks good, just a few minor comments

packages/api/src/google-cloud/google-cloud.service.ts
avalmas-programize: Done

packages/api/src/google-cloud/google-cloud.service.ts
avalmas-programize: Done

packages/api/src/google-cloud/google-cloud.service.ts
```diff
@@ -3,30 +3,79 @@ import {
...
+import { SurveyMedia } from '../surveys/survey-media.entity';
+import { getFileFromURL } from '../utils/google-cloud.utils';
 
 @Injectable()
 export class GoogleCloudService {
   private logger = new Logger('GCS');
   private storage: Storage;
 
-  constructor() {
+  private readonly GCS_BUCKET = process.env.GCS_BUCKET;
```
ericboucher, line 17: Is it possible to use destructuring here maybe?

ericboucher (approved):
One more minor comment but looks good, you can merge after

packages/api/src/google-cloud/google-cloud.service.ts
avalmas-programize: No not inside a class.

Issue #285, opened 2020-10-05
Backfill is not updating all data points


ericboucher:
When running a backfill, I noticed that not all the data points where actually updated properly.

One hypothesis is that in Typeorm `.update` function, the find condition does not actually return the element and hence no data is updated. It could be that the date used in the find condition differs? Could we use a "day is the same" type of condition instead maybe?

PR #290, opened 2020-10-06
Update entities based on date only

avalmas-programize:


avalmas-programize:
Related issue: https://github.com/aqualinkorg/aqualink-app/issues/285

ericboucher (approved):

Issue #287, opened 2020-10-05
Add points for survey dates on charts


ericboucher:

tleraj:
On which chart should we place the points? Surface temperature or bottom temperature. Right now each survey comes with a value for the bottom temperature, but we only display the surface temperature.

ericboucher:
@tleraj  let's put them on the bottom temp when it becomes available

PR #291, opened 2020-10-06
Surveys on chart

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/287. The design implemented looks like this:

Screenshot from 2020-10-06 19-13-44

### DIFF
--- a/packages/website/src/common/Chart/ChartWithTooltip.tsx
+++ b/packages/website/src/common/Chart/ChartWithTooltip.tsx
@@ -104,6 +104,12 @@ function ChartWithTooltip({
             display: true,
             rtl: true,
             labels: {
+              filter: (legendItem: any) => {
+                return (
+                  legendItem.text === "SURFACE TEMP" ||
+                  legendItem.text === "SURVEYS"
+                );
+              },
               fontSize: 14,
               fontColor: "#9ea6aa",
             },
--- a/packages/website/src/common/Chart/ChartWithTooltip.tsx
+++ b/packages/website/src/common/Chart/ChartWithTooltip.tsx
@@ -94,7 +94,7 @@ function ChartWithTooltip({
           },
           tooltips: {
             filter: (tooltipItem: any) => {
-              return tooltipItem.datasetIndex === 0;
+              return tooltipItem.datasetIndex === 1;
             },
             enabled: false,
             intersect: false,
--- a/packages/website/src/common/Chart/ChartWithTooltip.tsx
+++ b/packages/website/src/common/Chart/ChartWithTooltip.tsx
@@ -18,6 +18,7 @@ interface ChartWithTooltipProps extends ChartProps {
 function ChartWithTooltip({
   depth,
   dailyData,
+  surveys,
   temperatureThreshold,
   chartSettings,
   children,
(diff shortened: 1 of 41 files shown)


ericboucher:
Let's try to match the designs for the points

Screen Shot 2020-10-07 at 12 44 59 PM

packages/website/src/routes/Surveys/View/index.tsx
```diff
@@ -99,6 +99,7 @@ const SurveyViewPage = ({ reef, surveyId, classes }: SurveyViewPageProps) => {
                 <Grid container justify="center" item xs={12}>
                   <Charts
                     dailyData={reef.dailyData}
+                    surveys={[]}
```
ericboucher, line 102: Shouldn't we also pass surveys in here?

ericboucher:
The colors appear in the wrong order for me and the chart appears darker than it should? Maybe two layers are overlapping when they weren't?

Screen Shot 2020-10-07 at 12 56 36 PM

ericboucher:
The color of the graph is darker than the one on master (deployed on staging - https://aqualink-app-staging.web.app/ )

packages/website/src/common/Chart/ChartWithTooltip.tsx
```diff
@@ -94,7 +94,7 @@ function ChartWithTooltip({
           },
           tooltips: {
             filter: (tooltipItem: any) => {
-              return tooltipItem.datasetIndex === 0;
```
ericboucher, line 95: I don't think this is stable. Does it cover the case where there is no survey?

packages/website/src/common/Chart/index.tsx
```diff
@@ -105,7 +109,7 @@ function Chart({
           color: "rgb(158, 166, 170, 0.07)",
         },
         fillPlugin: {
-          datasetIndex: 0,
+          datasetIndex: 1,
```
ericboucher, line 112: Same instability risk here, no?

packages/website/src/common/Chart/utils.ts
```diff
@@ -1,21 +1,50 @@
...
 import type { DailyData } from "../../store/Reefs/types";
+import { SurveyListItem } from "../../store/Survey/types";
 
-export const createDatasets = (dailyData: DailyData[]) => {
+const getSurveyDates = (surveys: SurveyListItem[]): (number | null)[] => {
+  const dates = surveys.map((survey) => {
+    if (survey.diveDate) {
+      return new Date(survey.diveDate).setHours(0, 0, 0, 0);
+    }
+    return null;
```
ericboucher, line 11: why not filtering them out instead of setting them as null?

packages/website/src/common/Chart/index.tsx
ericboucher: This seems to work ok for me in both cases but just wanted to make sure.

packages/website/src/common/Chart/ChartWithTooltip.tsx
```diff
@@ -102,6 +108,12 @@ function ChartWithTooltip({
             display: true,
             rtl: true,
             labels: {
+              filter: (legendItem: any) => {
```
ericboucher, line 111: I think Drew actually wanted to remove all legends for now since they are pretty much self explanatory

packages/website/src/common/Chart/utils.ts
tleraj: It is essential that `tempWithSurvey` and `surfaceTemperature` arrays have the same length, so that each of their values corresponds to the same date.

PR #292, opened 2020-10-07
Improve table card alignment and small charts

ericboucher:
- Add a spacing before the table and align card and text
- Switch to monthly ticks when the charts are smaller (width < 400)

Screen Shot 2020-10-07 at 12 43 19 PM

### DIFF
--- a/packages/website/src/common/Chart/index.tsx
+++ b/packages/website/src/common/Chart/index.tsx
@@ -43,7 +43,7 @@ function Chart({
 
   const [xTickShift, setXTickShift] = useState<number>(0);
 
-  const [xPeriod, setXPeriod] = useState<string>("week");
+  const [xPeriod, setXPeriod] = useState<"week" | "month">("week");
 
   const {
     xAxisMax,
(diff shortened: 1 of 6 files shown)


packages/website/src/common/Chart/index.tsx
```diff
@@ -43,6 +43,8 @@ function Chart({
 
   const [xTickShift, setXTickShift] = useState<number>(0);
 
+  const [xPeriod, setXPeriod] = useState<string>("week");
```
SimplyBallistic, line 46: Could this be typed better than `string`? Don't need to worry about where its used but maybe a simple `'week'|'month'` would suffice

packages/website/src/routes/Homepage/ReefTable/body.tsx
```diff
@@ -106,4 +113,18 @@ const ReefTableBody = ({ order, orderBy }: ReefTableBodyProps) => {
...
 
-export default ReefTableBody;
+const styles = () =>
+  createStyles({
+    nameCells: {
+      paddingLeft: 10,
+    },
+  });
+
+type ReefTableBodyProps = {
```
SimplyBallistic, line 123: Not a big one, but is it better to stick to the project's naming convention for styled props: `ReefTableBodyIncomingProps`?

SimplyBallistic (approved):
Very minor details. Overall looks good!

PR #293, opened 2020-10-07
Add spotter application flow to the app

tleraj:
The purpose of this PR is to establish the new application flow in the frontend. The designs implemented look like this:

Screenshot from 2020-10-07 14-59-44
Screenshot from 2020-10-07 14-59-27


packages/website/src/routes/Apply/__snapshots__/index.test.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renders as expected 1`] = `
+exports[`Apply Page should render with given state from Redux store 1`] = `
```
ericboucher, line 3: Site registration page

packages/website/src/routes/Apply/index.tsx
```diff
@@ -56,10 +60,32 @@ const agreements = [
...
+          .set("name", user.fullName)
+          .set("org", user.organization)
+          .set("email", user.email)
+      );
+    } else {
+      setFormModel(formModel.set("name", "").set("org", "").set("email", ""));
+    }
+  }, [user, formModel]);
 
   const locationFormElements = [
```
ericboucher, line 90: Maybe move the reef name to here
  { id: "siteName", label: "Site Name" },


And disable all the contactForm elements instead?

packages/api/src/reef-applications/dto/create-reef-application.dto.ts
```diff
@@ -30,6 +30,9 @@ export class CreateReefApplicationDto {
 }
 
 export class CreateReefWithApplicationDto {
+  @IsNotEmpty()
```
ericboucher, line 33: Is this restriction also enforced on the frontend?

tleraj: Yes it is a required field.

packages/website/src/routes/Apply/index.tsx
tleraj: Then the fields will be in different order than the ones on the doc https://docs.google.com/document/d/1v8TZW2H74R2isCH-Uf57JkDkxoRYt2NjuXvLPpo7RTI/edit.

ericboucher: I think that's fine

ericboucher:
The new reef appears in my profile list as expected but it looks like we need to add a few guards to avoid errors if we try loading the reef too early:

Screen Shot 2020-10-07 at 5 45 05 PM

packages/website/src/routes/Apply/index.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    value: 26,
+    timestamp: "2020-07-01T14:25:18.008Z",
+  },
+  degreeHeatingDays: {
+    value: 32,
+    timestamp: "2020-07-01T14:25:18.008Z",
+  },
+};
+
+const dailyData = {
```
ericboucher, line 28: do you need this data in here?

tleraj: Just using it to mock a reef extensively.

packages/website/src/helpers/findAdministeredReef.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from "../store/User/types";
+import { Reef } from "../store/Reefs/types";
+
+export const findAdministeredReef = (
+  user: User | null,
+  reefId: number
+): Reef | undefined => {
+  const reef = user?.administeredReefs?.find((item) => item.id === reefId);
```
ericboucher, line 8: no need for a declaration here, just return

packages/website/src/routes/Apply/Form/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+  agreed,
+  handleFormSubmit,
+  classes,
+}: FormProps) => {
+  const [installationSchedule, setInstallationSchedule] = useState<Date | null>(
+    null
+  );
+
+  useEffect(() => {
+    if (application && application.installationSchedule) {
```
ericboucher, line 32: let's use chaining

ericboucher (approved):
LGTM with minor changes

PR #297, opened 2020-10-07
Better initial region for sites

ericboucher:


drewjgray (approved):
Nice, region priority makes sense.

packages/api/src/utils/reef.utils.ts
```diff
@@ -8,7 +12,29 @@ import { Region } from '../regions/regions.entity';
...
 const logger = new Logger('Reef Utils');
 
-const getCountry = async (
+const getLocality = (results: GeocodeResult[]) => {
+  const localityPreference = [
+    'administrative_area_level_2',
+    'administrative_area_level_1',
+    'locality',
+    'country',
+  ] as AddressType[];
```
avalmas-programize, line 21: Why using string when we can use the AddressType which is an enum to access its properties.
```
const localityPreference = [
    AddressType.administrative_area_level_2,
    AddressType.administrative_area_level_1,
    AddressType.locality,
    AddressType.country,
  ];
```

avalmas-programize (changes requested):

avalmas-programize (approved):

PR #301, opened 2020-10-10
[HOLD] Add `fixed` to all possible containers

Ry-DS:
Material UI's fixed property locks the possible max widths a container can be to 5 sizes:

image

This could help make responsive development more predictable since only 5 sizes need to be considered.
However, its important to test all pages aren't affected by this change.

Routes affected:
http://aqua-fixed.surge.sh/apply
http://aqua-fixed.surge.sh/reefs/67
http://aqua-fixed.surge.sh/reefs/67/survey_details/61
http://aqua-fixed.surge.sh/reefs/67/new_survey

Test: http://aqua-fixed.surge.sh

Results are most noticeable on iPad size on Devtools


tleraj (approved):
Everything is looking good!!

PR #306, opened 2020-10-12
Add spotter exclusion dates

avalmas-programize:
- Add exclusion dates table
- Filter out spotter data from spotter on excluded dates

### DIFF
--- a/packages/api/src/reefs/exclusion-dates.entity.ts
+++ b/packages/api/src/reefs/exclusion-dates.entity.ts
@@ -1,7 +1,8 @@
 import { Column, Entity, PrimaryGeneratedColumn, Unique } from 'typeorm';
 
 @Entity()
-@Unique('no_duplicates_exclusion_dates', ['spotterId', 'date'])
+@Unique('no_duplicates_start_date', ['spotterId', 'startDate'])
+@Unique('no_duplicate_end_date', ['spotterId', 'endDate'])
 export class ExclusionDates {
   @PrimaryGeneratedColumn()
   id: number;
@@ -9,6 +10,11 @@ export class ExclusionDates {
   @Column()
   spotterId: string;
 
+  // The start of the exclusive range. Inclusive
   @Column()
-  date: Date;
+  startDate: Date;
+
+  // The end of the exclusive range. Inclusive
+  @Column()
+  endDate: Date;
 }
--- a/packages/api/src/reefs/exclusion-dates.entity.ts
+++ b/packages/api/src/reefs/exclusion-dates.entity.ts
@@ -1,0 +1,14 @@
+import { Column, Entity, PrimaryGeneratedColumn, Unique } from 'typeorm';
+
+@Entity()
+@Unique('no_duplicates_exclusion_dates', ['spotterId', 'date'])
+export class ExclusionDates {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column()
+  spotterId: string;
+
+  @Column()
+  date: Date;
+}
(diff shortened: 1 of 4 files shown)


packages/api/src/reefs/exclusion-dates.entity.ts
```diff
@@ -0,0 +1,14 @@
...
+@Unique('no_duplicates_exclusion_dates', ['spotterId', 'date'])
+export class ExclusionDates {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column()
+  spotterId: string;
+
+  @Column()
+  date: Date;
```
ericboucher, line 13: I was thinking we could use a start_date and end_date, this way it might be easier to exclude long periods of time, and for one day we can just have start_date=end_date, what do you think?

packages/api/src/workers/dailyData.ts
```diff
@@ -277,7 +280,19 @@ export async function getReefsDailyData(
   await Bluebird.map(
     allReefs,
     async (reef) => {
-      const dailyDataInput = await getDailyData(reef, date);
+      const excludeSpotterData = await exclusionDatesRepository
```
ericboucher, line 283: Let's test for `spotterId` before even running this query.

ericboucher (commented):
It might be easier to use `inlcudeSpotterData` instead to have fewer negations.

packages/api/src/reefs/exclusion-dates.entity.ts
avalmas-programize: Okay. I will add a comment to remember this convention

packages/api/src/workers/dailyData.ts
avalmas-programize: Okay

ericboucher:
Let's address this:
```
It might be easier to use inlcudeSpotterData instead to have fewer negations.
```

packages/api/src/workers/dailyData.ts
```diff
@@ -277,7 +280,22 @@ export async function getReefsDailyData(
...
+            spotterId: reef.spotterId,
+          })
+          .andWhere('DATE(exclusion.startDate) <= DATE(:date)', { date })
+          .andWhere('DATE(exclusion.endDate) >= DATE(:date)', { date })
+          .getOne());
+
+      const dailyDataInput = await getDailyData(
+        reef,
+        date,
+        isNil(excludeSpotterData),
```
ericboucher, line 297: It's seems a bit weird that if reef.spotterId is undefined, then we would set `includeSpotterData = true`...

Maybe set:

const includeSpotterData = Boolean(
        reef.spotterId &&
        isNil(
			await exclusionDatesRepository
        	  .createQueryBuilder('exclusion')
          	  .where('exclusion.spotter_id = :spotterId', {
                spotterId: reef.spotterId,
              })
             .andWhere('DATE(exclusion.startDate) <= DATE(:date)', { date })
             .andWhere('DATE(exclusion.endDate) >= DATE(:date)', { date })
          .getOne());
     )

Or something like that?

This way you can also replace this line in dailyData.ts:

`spotterId && includeSpotterData` -> `includeSpotterData`

avalmas-programize: Yeah I have done something similar to that.

packages/api/src/workers/dailyData.ts
ericboucher: you undid this change

ericboucher (approved):

PR #318, opened 2020-10-19
Tweak card height and table scrollability

tleraj:
This PR affects the table scrollability and the selected reef card height. Specifically, for windows higher than `720px` only the table is scrollable. Otherwise the whole `div` that contains the card and the table is scrollable. Also, it fixes the image being too long issue. The designs look like this:

Screenshot from 2020-10-19 13-20-28
Screenshot from 2020-10-19 13-18-51


ericboucher:
Looks like it's not scaling properly on mobile?

Screen Shot 2020-10-19 at 4 29 24 PM

packages/website/src/routes/Homepage/ReefTable/SelectedReefCard/index.tsx
```diff
@@ -152,7 +152,7 @@ const SelectedReefContent = ({ reef, url }: SelectedReefContentProps) => {
       spacing={1}
     >
       {url && (
-        <Grid item xs={12} sm={5} lg={5}>
+        <Grid item xs={12} sm={6} lg={4}>
```
ericboucher, line 155: why these changes

ericboucher (approved):

PR #319, opened 2020-10-19
Avoid overlapping steps on chart y-ticks

tleraj:
The goal of this PR is to set constant step on charts y axis ticks. The designs look like this:

Screenshot from 2020-10-19 14-49-34  -----> Screenshot from 2020-10-19 14-49-59


ericboucher:
The problem with this method is that it goes against our goal of setting yMin and yMax to maximize the spread of data while showing the thresholds

tleraj:
Then another option could be to change the `yAxes` display callback like this:

```typescript
callback: (value: number) => {
  if (value % 5 === 0) {
    return `${value}\u00B0  `;
  }
  return "";
},
```

That way we do not change the yAxes range, while we show only the ticks that are multiples of the step.

ericboucher:
Yeah let's try the callback. Also use the literal `°C`

ericboucher:
I think we might need something a bit smarter, let's think about it a bit more. Also, the behavior you are seeing is probably much more likely with mock data than with real data. Let's make sure we test this with real data eg. on https://aqualink-app-staging.web.app/ or connecting to the prod db.


Screen Shot 2020-10-19 at 6 06 15 PM

packages/website/src/common/Chart/index.tsx
```diff
@@ -201,7 +201,10 @@ function Chart({
               stepSize: 5,
               max: yAxisMax,
               callback: (value: number) => {
-                return `${value}\u00B0  `;
+                if (value % 5 === 0) {
```
ericboucher, line 204: So what we actually wanna do is hide yMin and yMax if they are too close to the last (value % 5 === 0)

So how about?

	if (value % 5 == 1) return "" else return `${value}\u00B0  `;

packages/website/src/common/Chart/index.tsx
```diff
@@ -201,7 +201,10 @@ function Chart({
               stepSize: 5,
               max: yAxisMax,
               callback: (value: number) => {
-                return `${value}\u00B0  `;
+                if (![1, 4].includes(value % 5)) {
```
tleraj, line 204: I think we should introduce a variable `const stepSize = 5` and change the criteria to `if (![1, stepSize - 1].includes(value % 5))`.

ericboucher (approved):

PR #321, opened 2020-10-19
Update Site Location via Map

tleraj:
The purpose of this PR is to allow users to update the site's location by dragging the map's marker.


ericboucher:
Nice!
- Let's update the pin position when changing the lat lon fields
- Let's add a text below the lat/lon fields: `You can also change your site position by dragging the pin on the map.`

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -36,10 +47,53 @@ const EditForm = ({ reef, onClose, onSubmit, classes }: EditFormProps) => {
     [onSubmit]
   );
 
+  const onFieldChange = useCallback(
+    (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      const { name: field, value: newValue } = event.target;
+
+      if (draftReef?.coordinates) {
+        switch (field) {
```
ericboucher, line 55: this logic seems very weird.

Why not do something like:

if (draftReef?.coordinates) and field in [latitude, longitude]
dispatch.... ([field]: parseFloat(newValue)))

packages/website/src/store/Reefs/selectedReefSlice.ts
```diff
@@ -30,6 +31,10 @@ const selectedReefSlice = createSlice({
   name: "selectedReef",
   initialState: selectedReefInitialState,
   reducers: {
+    setDraft: (state, action: PayloadAction<SelectedReefState["draft"]>) => ({
```
ericboucher, line 34: Do we really have to introduce this new concept of draft in this case?

Maybe call it `setReefDraft` to be a bit more specific?

ericboucher (approved):

PR #330, opened 2020-10-21
Sort media by survey point name

tleraj:
The purpose of this PR is to sort the media carousels on survey details page by point of interest name.

### DIFF
--- a/packages/website/src/helpers/surveyMedia.ts
+++ b/packages/website/src/helpers/surveyMedia.ts
@@ -1,4 +1,4 @@
-import { groupBy } from "lodash";
+import { groupBy, sortBy } from "lodash";
 import type { SurveyMedia } from "../store/Survey/types";
 
 export const getFeaturedMedia = (surveyMedia: SurveyMedia[]) => {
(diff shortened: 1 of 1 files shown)


packages/website/src/helpers/surveyMedia.ts
```diff
@@ -25,8 +25,9 @@ export const getNumberOfSurveyPoints = (surveyMedia: SurveyMedia[]) => {
 };
 
 export const getSurveyPointsByName = (surveyMedia: SurveyMedia[]) => {
+  const sortedByPoiName = sortBy(surveyMedia, (media) => media.poiId?.name);
```
ericboucher, line 28: would that put "other" at the end?

packages/website/src/helpers/surveyMedia.ts
```diff
@@ -25,15 +25,27 @@ export const getNumberOfSurveyPoints = (surveyMedia: SurveyMedia[]) => {
...
-    name,
-    surveyMedia: pointSurveyMedia,
-  }));
+  const groupedMediaArray = Object.entries(groupedMedia).map(
+    ([name, pointSurveyMedia]) => ({
+      name,
+      surveyMedia: pointSurveyMedia,
+    })
+  );
+  const poiNames = Object.keys(groupedMedia);
```
ericboucher, line 42: this seems overly complicated to me. see my other comment

packages/website/src/helpers/surveyMedia.ts
ericboucher: why not just make sure HERE that the null / undefined case is sorted at the end.

Not ideal, but for example doing: => `media.poiId?.name || "zzzz"`

tleraj: Turns out that `lodash`'s `sortBy` places `null` or `undefined` values at the end. So it is guaranteed that `Other` will be at the end.

ericboucher (approved):

PR #335, opened 2020-10-22
API - Get spotter raw temperature data within time range

avalmas-programize:
- Add new endpoint on reefs controller: GET /reefs/:reef_id/spotter_data
- Add parse date pipe to parse query params

### DIFF
--- a/packages/api/src/utils/sofar.ts
+++ b/packages/api/src/utils/sofar.ts
@@ -2,6 +2,7 @@
 import axios from 'axios';
 import { isNil } from 'lodash';
 import axiosRetry from 'axios-retry';
+import moment from 'moment';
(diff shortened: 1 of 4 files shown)


packages/api/src/utils/sofar.ts
```diff
@@ -2,6 +2,7 @@
 import axios from 'axios';
 import { isNil } from 'lodash';
 import axiosRetry from 'axios-retry';
+import moment from 'moment-timezone';
```
ericboucher, line 5: `momemt from 'moment'` should do it, no?

avalmas-programize: Yeah correct. Fixed

ericboucher (approved):

Issue #139, opened 2020-09-06
Delete empty surveys


ericboucher:
If a survey is started but not completed (no media is uploaded). We should make sure that it is deleted (or never actually created)

PR #336, opened 2020-10-22
Add cronjob to delete empty surveys

avalmas-programize:
- Cronjob runs every 2 hours
- It grabs all empty surveys that have been created at most 2 hours ago (in order to avoid deleting not completed surveys)

Related issue https://github.com/aqualinkorg/aqualink-app/issues/139


packages/api/src/tasks/tasks.service.ts
```diff
@@ -0,0 +1,44 @@
...
+@Injectable()
+export class TasksService {
+  private readonly logger = new Logger(TasksService.name);
+
+  constructor(
+    @InjectRepository(Survey)
+    private surveyRepository: Repository<Survey>,
+  ) {}
+
+  @Cron('0 0-23/2 * * *', { name: CronJobs.DeleteEmptySurveys })
```
ericboucher, line 17: Let's add a comment explaining the expected schedule

avalmas-programize: Done

ericboucher: I think you can just do `*/2` instead of `0-23/2`

packages/api/src/tasks/tasks.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (emptySurveys.length) {
+      const results = await this.surveyRepository
+        .createQueryBuilder('survey')
+        .where('survey.id IN (:...ids)', { ids: emptyKeys })
+        .delete()
+        .execute();
+
+      this.logger.log(`Deleted ${results.affected} empty survey(s).`);
+    } else {
+      this.logger.log('Skipping deletion');
```
ericboucher, line 42: You're not really skipping deletion here, it's just that there is nothing to delete. Let's only have logs when emptySurveys.length > 0. And maybe a simple debug log to let us know when this task is being run.

packages/api/src/tasks/tasks.service.ts
avalmas-programize: This is the way the CronJobs module is using it, but I agree that is a bit redundant.

packages/api/src/tasks/tasks.service.ts
avalmas-programize: Done

ericboucher (approved):

PR #339, opened 2020-10-27
[HOLD] Upload thumbnail

avalmas-programize:
- Add file parser middleware to check and stop requests with content size greater than the upload limit. (multer limit is not currently working with google cloud storage as storage)
- Create new custom engine storage to handle the creation of the thumbnail and the upload of both images to cloud storage (original and thumbnail)
- Refactor engine storage to be able to handle multiple parallel uploads
- Refactor survey services and google cloud services to take into consideration the thumbnail as well.

Changes needed in frontend
- We need to handle two urls, which are returned from the upload endpoint instead of one
- In order to create a survey media two urls should be passed instead of one (imageUrl, thumbnailUrl)
- The thumbnailUrl should replace the imageUrl wherever we don't want to load the full resolution image


packages/api/migration/1603786749193-AddThumbnailInMedia.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddThumbnailInMedia1603786749193 implements MigrationInterface {
+  name = 'AddThumbnailInMedia1603786749193';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "survey_media" RENAME COLUMN "url" TO "original"`,
```
ericboucher, line 8: Let's simply use `image_url` and `thumbnail_url`, no?

packages/api/src/middleware/file-parser-middleware.ts
```diff
@@ -0,0 +1,17 @@
+import { Logger, PayloadTooLargeException } from '@nestjs/common'
+
+export const fileParserMiddleware = (req, res, next) => {
+  const logger = new Logger('FileParser');
```
ericboucher, line 4: Is this being used?

packages/api/src/middleware/file-parser-middleware.ts
```diff
@@ -0,0 +1,17 @@
...
+  const contentLength = req.headers['content-length'];
+  const maxFileSizeMB = process.env.STORAGE_MAX_FILE_SIZE_MB
+    ? parseInt(process.env.STORAGE_MAX_FILE_SIZE_MB, 10)
+    : 1;
```
ericboucher, line 8: What default do we have in the app currently? Might be worth defaulting to something a bit higher, maybe 5mb?

packages/api/src/surveys/surveys.service.ts
```diff
@@ -275,13 +276,30 @@ export class SurveysService {
...
-        return this.googleCloudService.deleteFile(file).catch(() => {
-          this.logger.error(
-            `Could not delete media ${media.url} of survey ${surveyId}.`,
-          );
-        });
+        const file = getFileFromURL(media.original);
+        // Thumbnail might not exist
+        const thumbnail = media.thumbnail && getFileFromURL(media.thumbnail);
+
+        return this.googleCloudService
```
ericboucher, line 283: I think we can simplify this function by doing something like:


[media.original, media.thumbnail].map((file) => {if (file) .... })

Or a variant of that

packages/api/src/surveys/surveys.service.ts
```diff
@@ -308,14 +326,26 @@ export class SurveysService {
     // We need to grab the path/to/file. So we split the url on "{GCS_BUCKET}/"
     // and grab the second element of the resulting array which is the path we need
     await this.googleCloudService
-      .deleteFile(getFileFromURL(surveyMedia.url))
+      .deleteFile(getFileFromURL(surveyMedia.original))
```
ericboucher, line 329: Same comment as above, I don't think it's useful to differentiate the behavior between original and thumbnail

packages/api/migration/1603786749193-AddThumbnailInMedia.ts
avalmas-programize: Ok

packages/api/src/middleware/file-parser-middleware.ts
avalmas-programize: No. Deleting

packages/api/src/middleware/file-parser-middleware.ts
avalmas-programize: Currently the default is 1mb. I will make it 5mb

packages/api/src/surveys/surveys.service.ts
avalmas-programize: ok

packages/api/src/surveys/surveys.service.ts
avalmas-programize: Done

packages/api/src/surveys/surveys.controller.ts
```diff
@@ -33,11 +33,16 @@ export class SurveysController {
...
-  upload(@UploadedFile('file') file: any): string {
+  upload(
+    @UploadedFile('file') file: any,
+  ): { original: string; thumbnail: string } {
     // Override file path because file.path provided an invalid google cloud format and HTTPS is not working correctly
     // Correct format of a URL pointing to a google cloud object should be
     // https://storage.googleapis.com/{bucketName}/path/to/object/in/bucket
-    return `https://storage.googleapis.com/${process.env.GCS_BUCKET}/${file.filename}`;
+    return {
+      original: `https://storage.googleapis.com/${process.env.GCS_BUCKET}/${file['0'].filename}`,
```
ericboucher, line 43: Let's make sure we use the `image_url` `thumbnail_url` paradigm throughout the server

avalmas-programize: Done

avalmas-programize:
Added ImageModal component which shows an image as a modal with a close button on the top right corner.
- Semi transparent overlay
- Disabled scrolling
- Modal closes if user clicks on close button or user clicks anywhere outside of the image

Screenshots:
image
The design is similar on mobile

ericboucher:
@avalmas-programize actually, could we look into a google storage solution for this problem? Maybe something like https://cloud.google.com/appengine/docs/standard/go111/images ?

PR #341, opened 2020-10-28
Update tooltip with colors

ericboucher:
Screen Shot 2020-10-28 at 7 49 56 PM

### DIFF
--- a/packages/website/src/common/Chart/Tooltip/index.tsx
+++ b/packages/website/src/common/Chart/Tooltip/index.tsx
@@ -31,13 +31,11 @@ const TemperatureMetric = (
   color: string,
   gridClassName: string | undefined
 ) => (
-  <Grid item xs={12} className={gridClassName}>
-    <Grid container justify="flex-start" item>
-      <Circle color={color} />
-      <Typography variant="caption">
-        {title} {`${formatNumber(temperature, 1)} °C`}
-      </Typography>
-    </Grid>
+  <Grid container justify="flex-start" item className={gridClassName}>
+    <Circle color={color} />
+    <Typography variant="caption">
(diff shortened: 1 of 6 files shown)


packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -28,19 +28,31 @@ const Tooltip = ({
     minute: spotterSurfaceTemp ? "2-digit" : undefined,
   });
 
+  const Circle = (backgroundColor: string, size: number) => (
```
SimplyBallistic, line 31: TODO make a component

packages/website/src/common/Chart/index.tsx
```diff
@@ -241,7 +240,7 @@ function Chart({
         spotterData?.surfaceTemperature || [],
         tempWithSurvey,
         surfaceTemperatureData,
-        !spotterData ? bottomTemperatureData : [],
```
SimplyBallistic, line 244: TODO reintroduce bottom temp as bool (possibly, investigate)

packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -9,8 +9,38 @@ import {
...
+  })
+);
+
+const TemperatureMetric = (
+  temperature: number,
+  title: string,
+  color: string,
+  gridClassName: string | undefined
+) => (
+  <Grid item xs={12} className={gridClassName}>
```
SimplyBallistic, line 34: from testing, this second grid wrap seems redundant, the class could be moved to the parent Grid, and `xs` could be removed. Completed Grid header: `<Grid container justify="flex-start" item className={gridClassName}>`. Think this is a good idea?

Issue #342, opened 2020-10-28
Add a new home page


ericboucher:
Link to Zeplin: https://app.zeplin.io/project/5ec6b346f536202b32a4011e/screen/5f99975e8a33301fc85ad1d5

Preview (refer to Zeplin for the most up to date version):
![newhome1](https://user-images.githubusercontent.com/16843267/97503521-7af9ee80-1975-11eb-8619-2f87748b93c5.jpg)

PR #344, opened 2020-10-30
Add landing page

tleraj:
The purpose of this PR is to implement the new designs for the landing page and close https://github.com/aqualinkorg/aqualink-app/issues/342.


packages/website/src/common/NavBar/index.test.tsx
```diff
@@ -28,7 +28,7 @@ describe("NavBar", () => {
 
     element = render(
       <Provider store={store}>
-        <HomePageNavBar searchLocation={false} />
+        <HomePageNavBar routeButtons={false} searchLocation={false} />
```
SimplyBallistic, line 31: Route buttons could be omitted since its default, not too concerning

packages/website/src/routes/Landing/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+import Card from "./Card";
+import landingPageImage from "../../assets/img/landing-page/header.jpg";
+import { cardTitles } from "./titles";
+
+const LandingPage = ({ classes }: LandingPageProps) => {
+  return (
+    <>
+      <NavBar routeButtons searchLocation={false} />
+      <div className={classes.landingImage}>
+        <div className={classes.overlay}>
```
SimplyBallistic, line 26: Instead of using divs could we use boxes here? For example, Box might have props to flex or align center

SimplyBallistic (approved):
Small comments, otherwise looks good!

PR #348, opened 2020-10-31
Add shipped status

ericboucher:
Add `shipped` to `status` enum in the reef table


avalmas-programize:
In order to add a value to an enum we can use the syntax:
```sql
ALTER TYPE "public"."reef_status_enum" ADD VALUE 'shipped'
```
and skip all the dropping and creating a new enum.

Also the reverting function was not working if a reef had 'shipped' status. So I added an update query that changes all shipped statuses to 'in_review'.

avalmas-programize (approved):

PR #351, opened 2020-11-03
Add error boundary

tleraj:
The purpose of this PR is to create an error boundary in order to display a fallback UI instead of component tree that crashed due to an error. The fallback UI design looks like this:

Screenshot from 2020-11-03 14-15-21


packages/website/src/layout/App/errorBoundary.tsx
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable react/prefer-stateless-function */
+import React, { Component, PropsWithChildren } from "react";
+import ErrorPage from "../../common/ErrorPage";
+
+interface ErrorBoundaryState {
+  hasError: boolean;
+}
+
+class ErrorBoundary extends Component<
```
SimplyBallistic, line 9: Why is this a class component compared to function?

SimplyBallistic:
Still face various random issues when I try clicking populated areas and switch quickly like you described:
image
image

SimplyBallistic:
If this issue is easy to replicate, it might be more worthwhile if our component 'restarted' the map (render nothing then render map) since this would remove user intervention. It would be confusing for some users as the map might disappear for a second, but it will remove the need for reloading. We can do this until the issue you mentioned on leaflet is solved.

packages/website/src/common/ErrorPage/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+              style={{ margin: "1rem" }}
+              color="primary"
+              variant="contained"
+              onClick={() => window.location.reload()}
+            >
+              Try Again
+            </Button>
+            <Button
+              style={{ margin: "1rem", color: "white" }}
+              component={Link}
```
SimplyBallistic, line 41: Don't think this will work because the `ErrorBoundary` component is placed outside of a router

packages/website/src/common/ErrorPage/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <Typography variant="h2">Sorry, something went wrong...</Typography>
+          </Box>
+          <Grid container direction="row" justify="center">
+            <Button
+              style={{ margin: "1rem" }}
+              color="primary"
+              variant="contained"
+              onClick={() => window.location.reload()}
+            >
+              Try Again
```
SimplyBallistic, line 37: I think renaming this to `Refresh` is better, since `Try Again` implies we are fixing the issue on the spot. Just my opinion however :)

packages/website/src/common/ErrorPage/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+            color="primary"
+            variant="contained"
+            onClick={() => window.location.reload()}
+          >
+            Try Again
+          </Button>
+          <Button
+            style={{ margin: "1rem", color: "white" }}
+            component={Link}
+            to="/map"
```
tleraj, line 39: If the app has crashed, the `Back to Map` button click has no effect.

ericboucher (approved):

PR #356, opened 2020-11-05
Remove `ActiveReefListener` component

tleraj:
This PR is scheduled to fix the homepage map crashing issue. It basically removes the `ActiveReefListener` component, which used `leaflet`'s `"moveend"`, which seemed to cause the issue. This approach passes a `ref` to the selected reef's `Popup` and opens it via `popup.setLatLng([lat, lng]).openOn(map)`.


tleraj:
Perhaps we should only wrap the `HomeMap` with the `ErrorBoundary` rather than the whole app. Any suggestions?

Issue #345, opened 2020-10-31
Survey Page Improvements


ericboucher:
- [ ] Allow editing the name of survey points (add a pen logo next to the trash can for deleting)

- [ ] Filtering on survey points should change the image too and take the first available image for this survey point.

- [ ] When hovering on a survey point tooltip, add “View Survey”  button linking to the survey page

- [ ] On the Survey page, color the selected survey point in blue

PR #360, opened 2020-11-06
Surveys improvements

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/345. Specifically:
 - [x] Allow editing the name of survey points (add a pen logo next to the trash can for deleting)
 - [x] Filtering on survey points should change the image too and take the first available image for this survey point
 - [x] When hovering on a survey point tooltip, add “View Survey” button linking to the survey page
 - [x] On the Survey page, color the selected survey point in blue

Designs implemented look like this:

| Initial layout | Edit button clicked | New Tooltip | Selected survey point |
|---|---|---|---|
| Screenshot from 2020-11-11 16-52-12 | Screenshot from 2020-11-11 16-51-14 | Screenshot from 2020-11-10 15-30-45 | Screenshot from 2020-11-11 13-44-22 |

### DIFF
--- a/packages/api/src/surveys/surveys.service.ts
+++ b/packages/api/src/surveys/surveys.service.ts
@@ -139,7 +139,11 @@ export class SurveysService {
         { reefId },
       )
       .groupBy('surveyMedia.surveyId, surveyMedia.poiId')
-      .select(['surveyMedia.surveyId', 'surveyMedia.poiId'])
+      .select([
+        'surveyMedia.surveyId',
+        'surveyMedia.poiId',
+        'array_agg(surveyMedia.url) poi_images',
+      ])
       .getRawMany();
 
     const observationsGroupedBySurveyId = this.groupBySurveyId(
@@ -148,6 +152,12 @@ export class SurveysService {
     );
     const poiIdGroupedBySurveyId = this.groupBySurveyId(
       surveyPointsQuery,
+      'poi_id',
+    );
+
+    const surveyImageGroupedByPoiId = this.groupBySurveyId(
+      surveyPointsQuery,
+      'poi_images',
       'poi_id',
     );
 
@@ -168,6 +178,7 @@ export class SurveysService {
         featuredSurveyMedia: survey.featuredSurveyMedia,
         observations: observationsGroupedBySurveyId[survey.id] || [],
         surveyPoints: poiIdGroupedBySurveyId[survey.id] || [],
+        surveyPointImage: surveyImageGroupedByPoiId[survey.id] || [],
       };
     });
   }
@@ -359,11 +370,13 @@ export class SurveysService {
     return trimmedComments === '' ? undefined : trimmedComments;
   }
 
-  private groupBySurveyId(object: any[], key: string) {
+  private groupBySurveyId(object: any[], key: string, extra?: string) {
     return object.reduce((rv, x) => {
       return {
         ...rv,
-        [x.survey_id]: [...(rv[x.survey_id] || []), x[key]],
+        [x.survey_id]: extra
+          ? { ...rv[x.survey_id], [x[extra]]: x[key] }
+          : [...(rv[x.survey_id] || []), x[key]],
       };
     }, {});
   }
--- a/packages/api/src/surveys/surveys.service.ts
+++ b/packages/api/src/surveys/surveys.service.ts
@@ -370,13 +370,20 @@ export class SurveysService {
     return trimmedComments === '' ? undefined : trimmedComments;
   }
 
-  private groupBySurveyId(object: any[], key: string, extra?: string) {
-    return object.reduce((rv, x) => {
+  /**
+   * Group the values of the provided object array by the survey id and optionally by a secondary key
+   *
+   * @param object The object of arrays to perform the group (must have a survey id in each record)
+   * @param key The key of the value you want to group
+   * @param secondary The optional secondary key to perform a deeper group
+   */
+  private groupBySurveyId(object: any[], key: string, secondary?: string) {
+    return object.reduce((result, current) => {
       return {
-        ...rv,
-        [x.survey_id]: extra
-          ? { ...rv[x.survey_id], [x[extra]]: x[key] }
-          : [...(rv[x.survey_id] || []), x[key]],
+        ...result,
+        [current.survey_id]: secondary
+          ? { ...result[current.survey_id], [current[secondary]]: current[key] }
+          : [...(result[current.survey_id] || []), current[key]],
       };
     }, {});
   }
--- a/packages/website/src/common/Chart/utils.ts
+++ b/packages/website/src/common/Chart/utils.ts
@@ -37,33 +37,6 @@ export const filterDailyData = (
   return ret;
 };
 
-export const findSurveyFromDate = (
-  inputDate: string,
-  surveys: SurveyListItem[]
-): number | null | undefined => {
-  const inputDateTime = new Date(inputDate);
-  const formatInputDate = `${inputDateTime.getFullYear()}-${
-    inputDateTime.getMonth() + 1
-  }-${inputDateTime.getDate()}`;
-  const surveyDates = surveys.map((survey) => {
-    if (survey.diveDate) {
-      const date = new Date(survey.diveDate);
-      return `${date.getFullYear()}-${date.getMonth() + 1}-${date.getDate()}`;
-    }
-    return null;
-  });
-
-  const survey = surveys.find(
-    (_survey, index) => surveyDates[index] === formatInputDate
-  );
-
-  if (survey) {
-    return survey.id;
-  }
-
-  return null;
-};
-
 const getSurveyDates = (surveys: SurveyListItem[]): (number | null)[] => {
   const dates = surveys.map((survey) => {
     if (survey.diveDate) {
@@ -82,6 +55,17 @@ export const sameDay = (
 
 const timeDiff = (incomingDate: string, date: Date) =>
   Math.abs(new Date(incomingDate).getTime() - date.getTime());
+
+export const findSurveyFromDate = (
+  inputDate: string,
+  surveys: SurveyListItem[]
+): number | null | undefined => {
+  return (
+    surveys.find(
+      (survey) => survey.diveDate && sameDay(survey.diveDate, inputDate)
+    )?.id || null
+  );
+};
 
 export function getDailyDataClosestToDate(dailyData: DailyData[], date: Date) {
   return dailyData.reduce((prevClosest, nextPoint) =>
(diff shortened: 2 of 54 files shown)


packages/api/src/surveys/surveys.service.ts
```diff
@@ -359,11 +370,13 @@ export class SurveysService {
...
 
-  private groupBySurveyId(object: any[], key: string) {
+  private groupBySurveyId(object: any[], key: string, extra?: string) {
     return object.reduce((rv, x) => {
       return {
         ...rv,
-        [x.survey_id]: [...(rv[x.survey_id] || []), x[key]],
+        [x.survey_id]: extra
+          ? { ...rv[x.survey_id], [x[extra]]: x[key] }
+          : [...(rv[x.survey_id] || []), x[key]],
```
SimplyBallistic, line 379: Is the ` || []` discrepancy between these two new lines normal? From my knowledge if you spread `undefined` JS does nothing. Otherwise if there's a deeper reason here, disregard this comment :)

packages/website/src/routes/ReefRoutes/Reef/Surveys/SurveyCard/index.tsx
```diff
@@ -17,17 +17,30 @@ import DeleteButton from "../DeleteButton";
...
-            <CardMedia
-              className={classes.cardImage}
-              image={survey.featuredSurveyMedia.url}
-            />
-          )}
+          {point === -1
+            ? survey.featuredSurveyMedia && (
+                <CardMedia
+                  className={classes.cardImage}
+                  image={survey.featuredSurveyMedia.url}
```
SimplyBallistic, line 35: If the only thing that's different here is `image` it might be possible to move the conditional to `image` itself

packages/website/src/routes/ReefRoutes/Reef/Surveys/index.tsx
```diff
@@ -50,8 +58,16 @@ const Surveys = ({ reefId, classes }: SurveysProps) => {
     const source = Axios.CancelToken.source();
     reefServices
       .getReefPois(`${reefId}`, source.token)
-      .then((response) => {
-        setPointOptions(response.data);
+      .then(({ data }) => {
+        setPointOptions(data);
+        if (data.length > 0) {
+          setEditPoiNameEnabled(
+            data.reduce((acc, poi) => ({ ...acc, [poi.id]: false }), {})
```
SimplyBallistic, line 65: I feel like there's a type missing here, since from my experience reduce normally needs something given to it. If it's implicit, maybe we can add a cast or type generic?

packages/website/src/routes/ReefRoutes/Reef/Surveys/index.tsx
```diff
@@ -75,8 +95,14 @@ const Surveys = ({ reefId, classes }: SurveysProps) => {
...
   }, [onResize]);
 
+  const onDeletePoiButtonClick = useCallback((id: number) => {
+    setDeletePoiDialogOpen(true);
+    setPoiToDelete(id);
+  }, []);
+
   const handlePointChange = (event: React.ChangeEvent<{ value: unknown }>) => {
-    setPoint(event.target.value as string);
+    const poiName = event.target.value as string;
```
SimplyBallistic, line 104: Fine if this was done for cleanliness, but it might be residue since I don't see the constant being used anywhere, unless I missed something haha. Maybe a comment instead would suffice?

packages/website/src/routes/ReefRoutes/Reef/Surveys/index.tsx
```diff
@@ -113,6 +139,70 @@ const Surveys = ({ reefId, classes }: SurveysProps) => {
...
+        setEditPoiNameEnabled({ ...editPoiNameEnabled, [key]: enabled });
+        // Reset Poi name draft on close
+        const poiName = pointOptions.find((item) => item.id === key)?.name;
+        if (poiName && !enabled) {
+          setEditPoiNameDraft({ ...editPoiNameDraft, [key]: poiName });
+        }
+      } else {
+        // If no key provided then change all
+        setEditPoiNameEnabled(
+          pointOptions.reduce((acc, poi) => ({ ...acc, [poi.id]: enabled }), {})
```
SimplyBallistic, line 155: Same as last comment on `reduce`, another on line 160

packages/website/src/routes/ReefRoutes/Reef/Surveys/index.tsx
```diff
@@ -113,6 +139,70 @@ const Surveys = ({ reefId, classes }: SurveysProps) => {
...
+        ...editPoiNameDraft,
+        [key]: event.target.value,
+      }),
+    [editPoiNameDraft]
+  );
+
+  const submitPoiNameUpdate = useCallback(
+    (key: number) => {
+      const newName = editPoiNameDraft[key];
+      if (newName !== "" && user?.token) {
```
SimplyBallistic, line 185: I think an empty string is falsely. But this is fine too :)

SimplyBallistic (commented):
The UI screenshots look really good! These are just some quality things that could be fixed :+1:

SimplyBallistic:
Do the Reds on the trashcan and cross match from the screenshots? Might be good to make them the same or make the trashcan a colour already present in the application.

packages/api/src/surveys/surveys.service.ts
ericboucher: @tleraj thoughts on this?

SimplyBallistic: Ahh I just realised one is an array, thought they were two objects at first glance. I think this is good then, can't think of a cleaner way to do this

packages/website/src/common/Chart/utils.ts
```diff
@@ -37,6 +37,33 @@ export const filterDailyData = (
...
+
+  const survey = surveys.find(
+    (_survey, index) => surveyDates[index] === formatInputDate
+  );
+
+  if (survey) {
+    return survey.id;
+  }
+
+  return null;
```
ericboucher, line 64: This can probaby be a one liner.

eg. 

return surveys.find(
    (_, index) => surveyDates[index] === formatInputDate
  )?.id || null

packages/website/src/common/Chart/utils.ts
```diff
@@ -262,7 +290,28 @@ export const createChartData = (
         data: tempWithSurvey,
         pointRadius: 5,
         backgroundColor: "#ffffff",
-        pointBackgroundColor: "#ffff",
+        pointBackgroundColor: (context) => {
+          if (surveyDate) {
+            const surveyDateFormatted = `${surveyDate.getFullYear()}-${
+              surveyDate.getMonth() + 1
+            }-${surveyDate.getDate()}`;
```
ericboucher, line 297: Big NO on this @tleraj 
This is VERY error-prone and created a bug last time because getMonth returns a number not a "01" padded string

packages/website/src/common/Chart/utils.ts
```diff
@@ -262,7 +290,28 @@ export const createChartData = (
...
+
+              if (chartDateFormatted === surveyDateFormatted) {
+                return "#6bc1e1";
+              }
+              return "#ffffff";
+            }
+            return "#ffffff";
+          }
+          return "#ffffff";
+        },
```
ericboucher, line 314: This seems quite inefficient

packages/website/src/common/Chart/utils.ts
```diff
@@ -262,7 +290,28 @@ export const createChartData = (
...
+            }-${surveyDate.getDate()}`;
+            const index = context.dataIndex;
+            if (context.dataset?.data && typeof index === "number") {
+              const chartPoint = context.dataset.data[index] as ChartPoint;
+              const chartDate = new Date(chartPoint.x as string);
+              const chartDateFormatted = `${chartDate.getFullYear()}-${
+                chartDate.getMonth() + 1
+              }-${chartDate.getDate()}`;
+
+              if (chartDateFormatted === surveyDateFormatted) {
```
ericboucher, line 306: Following up on the comment above, I think you can just use the utils function `sameDay`

packages/api/src/surveys/surveys.service.ts
```diff
@@ -359,11 +370,13 @@ export class SurveysService {
     return trimmedComments === '' ? undefined : trimmedComments;
   }
 
-  private groupBySurveyId(object: any[], key: string) {
+  private groupBySurveyId(object: any[], key: string, extra?: string) {
```
ericboucher, line 373: Can we use better names than key, x, rv here maybe? :)

packages/website/src/common/Chart/utils.ts
```diff
@@ -262,7 +274,19 @@ export const createChartData = (
...
         backgroundColor: "#ffffff",
-        pointBackgroundColor: "#ffff",
+        pointBackgroundColor: (context) => {
+          if (
+            surveyDate &&
+            context.dataset?.data &&
+            typeof context.dataIndex === "number"
+          ) {
+            const index = context.dataIndex;
+            const chartPoint = context.dataset.data[index] as ChartPoint;
```
ericboucher, line 284: no need to define index here, just use `context.dataIndex;` directly

packages/website/src/routes/ReefRoutes/Reef/Surveys/PointSelector/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+      color: theme.palette.primary.main,
+    },
+    menuItem: {
+      color: theme.palette.primary.main,
+    },
+    editPoiTextField: {
+      color: "black",
+      height: "2.5rem",
+      alignItems: "center",
+      "&:hover .MuiOutlinedInput-root .MuiOutlinedInput-notchedOutline": {
```
ericboucher, line 199: you should not need to call the Mui names explicitly. You probably want to do something like this - https://material-ui.com/customization/components/#overriding-styles-with-classes

packages/website/src/routes/ReefRoutes/Reef/Surveys/PointSelector/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+        borderColor: "rgba(0, 0, 0, 0.23)",
+      },
+      "& .MuiOutlinedInput-root.Mui-focused .MuiOutlinedInput-notchedOutline": {
```
ericboucher, line 202: ditto

packages/api/src/surveys/surveys.service.ts
avalmas-programize: Done.
Also I have added a doc string to the function

ericboucher (approved):

PR #361, opened 2020-11-08
Hide spotter chart when site has no spotter

ericboucher:
We noticed that when visiting a site with no spotter after visiting a site with a spotter, the spotter chart would sometimes stay on with data from the previous site.

This PR is adding an `hasSpotter` check before displaying spotter charts.

We could also look into making sure that `spotterData` get's removed properly when a new site is selected.

@tleraj let me know what you think

### DIFF
--- a/packages/website/src/routes/ReefRoutes/Reef/index.tsx
+++ b/packages/website/src/routes/ReefRoutes/Reef/index.tsx
@@ -25,6 +25,7 @@ import {
   reefRequest,
   reefSpotterDataRequest,
   reefSpotterDataSelector,
+  setReefSpotterData,
 } from "../../../store/Reefs/selectedReefSlice";
 import {
   surveysRequest,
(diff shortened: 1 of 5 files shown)


packages/website/src/store/Reefs/selectedReefSlice.ts
```diff
@@ -99,6 +99,13 @@ const selectedReefSlice = createSlice({
       }
       return state;
     },
+    setReefSpotterData: (
```
SimplyBallistic, line 102: Could just leave as it is, but if the only action this is doing is clearing, we could rename to `clearReefSpotterData` so you only have a choice to fetch from api or clear.

SimplyBallistic (approved):
Overall looks good :+1:

tleraj (approved):
Looks good!!

packages/website/src/routes/ReefRoutes/Reef/index.tsx
```diff
@@ -144,7 +144,7 @@ const Reef = ({ match, classes }: ReefProps) => {
       );
     } else {
       // Clear possible spotter data from previously selected reef
-      dispatch(setReefSpotterData(null));
+      dispatch(clearReefSpotterData(null));
```
ericboucher, line 147: @tleraj you could even go one step further and have `clearReefSpotterData()` instead and move the `set null` behavior to the slice.

PR #362, opened 2020-11-09
Add ping cloud function and healthCheck endpoint

avalmas-programize:
Add ping cloud function (schedule ping for every 5 minutes)
New cloud function env variable is needed (backend_base_url => The base url of the backend api)
Add healthCheck module and controller (used as a simple ping endpoint):
- GET /health-check: Return { status: 200 }

### DIFF
--- a/packages/api/cloud-functions/index.ts
+++ b/packages/api/cloud-functions/index.ts
@@ -1,3 +1,4 @@
+import Axios from 'axios';
 import * as functions from 'firebase-functions';
 import { createConnection } from 'typeorm';
 import { runDailyUpdate } from '../src/workers/dailyData';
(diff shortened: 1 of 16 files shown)


packages/api/cloud-functions/index.ts
```diff
@@ -80,3 +81,14 @@ exports.scheduledDailyUpdate = functions
       conn.close();
     }
   });
+
+exports.pingService = functions
+  .runWith({ timeoutSeconds: 540 })
```
ericboucher, line 86: I don't think we need a 9min timeout for this

packages/api/cloud-functions/index.ts
```diff
@@ -80,3 +81,14 @@ exports.scheduledDailyUpdate = functions
...
+  .pubsub.schedule('*/5 * * * *')
+  .timeZone('America/Los_Angeles')
```
ericboucher, line 88: no need for a timezone either

ericboucher:
@avalmas-programize let's make sure you update the README command with the new variable:


export $(grep -v '^#' .env.staging | xargs) && firebase functions:config:set database.url=$DATABASE_URL sofar_api.token=$SOFAR_API_TOKEN ...

ericboucher (changes requested):
See inline comments + add env_variable

packages/api/cloud-functions/index.ts
avalmas-programize: Ok

packages/api/cloud-functions/index.ts
avalmas-programize: Ok

ericboucher (approved):

PR #363, opened 2020-11-10
Use exclusion dates for live spotter data

avalmas-programize:


packages/website/src/routes/ReefRoutes/Reef/index.tsx
```diff
@@ -116,7 +116,7 @@ const Reef = ({ match, classes }: ReefProps) => {
 
   const { liveData, dailyData } = reefDetails || {};
 
-  const hasSpotter = Boolean(liveData?.surfaceTemperature);
+  const hasSpotter = Boolean(reefDetails?.spotterId);
```
ericboucher, line 119: I think this boolean was good. But maybe rename it to `hasSpotterData`?

We can have a spotterId set in the database but it might not send any data if it's not in the water for example

ericboucher: @tleraj

packages/api/src/reefs/reefs.service.ts
```diff
@@ -252,4 +293,20 @@ export class ReefsService {
...
   }
+
+  private filterSpotterDate(
+    spotterDate: SofarValue[],
+    exclusionDates: ExclusionDates[],
+  ) {
+    return spotterDate.filter(({ timestamp }) => {
+      const excluded = isNil(
+        exclusionDates.find(({ startDate: start, endDate: end }) => {
+          const dataDate = new Date(timestamp);
```
ericboucher, line 304: Maybe we could use something like this instead?

```
export const sameDay = (
  date1: string | number | Date,
  date2: string | number | Date
) => new Date(date1).toDateString() === new Date(date2).toDateString();
```

packages/api/src/reefs/reefs.service.ts
```diff
@@ -252,4 +293,20 @@ export class ReefsService {
       .of(reef)
       .addAndRemove(admins, reef.admins);
   }
+
+  private filterSpotterDate(
```
ericboucher, line 297: Should this be `filterSpotterDataByDate`?

packages/api/src/reefs/reefs.service.ts
avalmas-programize: We are dealing with ranges, not just dates. The dateString would not support the '>' very well.

packages/api/src/reefs/reefs.service.ts
ericboucher: @avalmas-programize

packages/api/src/reefs/reefs.service.ts
```diff
@@ -225,15 +246,35 @@ export class ReefsService {
...
+      .createQueryBuilder('exclusion')
+      .where('exclusion.spotter_id = :spotterId', {
+        spotterId: reef.spotterId,
+      })
+      .andWhere('DATE(exclusion.startDate) <= DATE(:endDate)', {
+        endDate,
+      })
+      .andWhere('DATE(exclusion.endDate) >= DATE(:startDate)', {
+        startDate,
+      })
```
ericboucher, line 259: Let's just get all the dates here. We won't have many exclusion dates anyways

packages/api/src/reefs/reefs.service.ts
avalmas-programize: Ok

ericboucher (approved):

PR #364, opened 2020-11-13
Add autoselect and geolocation to search bar

ericboucher:
This PR:
- adds the region name to the searchable text and display
- autoselects the first row so that clicking enter zooms on it
- if no site is found, propose to use geolocation on `enter`

Screen Shot 2020-11-13 at 3 17 08 PM
Screen Shot 2020-11-13 at 3 17 20 PM

### DIFF
--- a/packages/website/.env.example
+++ b/packages/website/.env.example
@@ -15,3 +15,7 @@ REACT_APP_CARTO_API_KEY=
 
 # Mapbox token to access custom styles
 REACT_APP_MAPBOX_ACCESS_TOKEN=
+
+# Mapbox API Base URL
+# Proposed value: "https://api.mapbox.com/geocoding/v5/mapbox.places/"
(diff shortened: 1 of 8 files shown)


SimplyBallistic (approved):
Overall looks good :+1:

tleraj:
Noticed the following behavior: if I select an item from the dropdown or  submit a search with an existing reef and afterwards searched for a country, then the submit had no effect. That happened because the `searchedReef` was never reset.

tleraj (approved):
Looks good!!

thanasislerai:
Noticed the following bug on mobile: if a location search has taken place and then a reef was selected from the table, the map didn't fly to that reef. Fixed!

tleraj:
Noticed the following bug on mobile: if a location search has taken place and then a reef was selected from the table, the map didn't fly to that reef. Fixed!

Issue #346, opened 2020-10-31
Use "site" local timezone


ericboucher:
Whenever possible, we should use the "site / reef" local timezone to display and select dates in.

- chart labels and tooltips
- create a survey
- survey dates / last surveyed

PR #365, opened 2020-11-17
Use site's local time zone

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/346.


packages/website/src/helpers/dates.ts
```diff
@@ -39,3 +40,60 @@ export const findChartPeriod = (range: Range) => {
       return "day";
   }
 };
+
+// Converts a given date to a specified time zone
+export const convertToLocalTime = (
+  utcTime?: string | null,
+  timeZone?: string | null
+): string | null =>
+  // eslint-disable-next-line no-nested-ternary
```
ericboucher, line 49: These linting rules exist for a reason ;)

packages/website/src/helpers/dates.ts
```diff
@@ -39,3 +40,60 @@ export const findChartPeriod = (range: Range) => {
...
+    ? timeZone
+      ? new Date(utcTime).toLocaleString("en-US", { timeZone })
+      : utcTime
+    : null;
+
+// Returns the same date but for a different time zone
+export const setTimeZone = (date: Date | null, timeZone?: string | null) => {
+  if (date) {
+    if (timeZone) {
+      const localTime = new Date(date.toLocaleString("en-US", { timeZone }));
```
ericboucher, line 60: This function would be much simpler to use if written like this:

```
if (date && timeZome) {
  const ....
  return ...
}

return date
```

ericboucher: Also, do we actually need to change all the dates? What we really want is just a way to display dates differently, but it's ok if they are stored in UTC along the way, no?

tleraj: We need to take into consideration the following scenario: the reef manager creates a survey from a timezone different than the site's timezone. Then the time he sets in the survey form needs to correspond to the site's local time. For example, if he sets dive date `2020-11-19 08:15 AM`, then this time should correspond to the site's local time, and not the manager's local time. That's why `setTimeZone` is used for. If this is an overkill, and assume that most of the times the manager's timezone will be equal to the site's timezone, then we can remove this.

packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -86,7 +86,6 @@ const Tooltip = ({
           title={
             <Typography color="textPrimary" variant="caption">
               {dateString}
-              {spotterSurfaceTemp && " UTC"}
```
ericboucher, line 89: Maybe replace by `Local`?

packages/website/src/helpers/dates.ts
```diff
@@ -39,3 +40,56 @@ export const findChartPeriod = (range: Range) => {
       return "day";
   }
 };
+
+// Converts a given date to a specified time zone
+export const convertToLocalTime = (
```
ericboucher, line 45: maybe this could take a format option as well?

packages/website/src/common/SurveyForm/index.tsx
```diff
@@ -49,13 +55,15 @@ const SurveyForm = ({ reefId, onSubmit, classes }: SurveyFormProps) => {
   const nativeSubmit = useCallback(
     (data: any) => {
```
SimplyBallistic, line 56: If the only variable you use is `comments` here you could maybe construct an incomplete type just for this function?
For example, we can see `comments: string` from the PropType so we could just do `data: {comments: string}`

packages/website/src/common/SurveyForm/index.tsx
```diff
@@ -289,6 +298,10 @@ interface SurveyFormIncomingProps {
   ) => void;
 }
 
+SurveyForm.defaultProps = {
+  timeZone: null,
```
SimplyBallistic, line 302: @ericboucher Ben let me know its not good practice to use defaultProps in favor of object default values (set on line 31 here), especially in TS anymore as its planned to be phased out, and there is no typing on `defaultProps`. We could do the same here?
We would need to disable the eslint rule, then move the default value to line 31 instead as `{ ..., timeZone = null, ...}`
Other default props could be moved incrementally.
Otherwise, this is fine as well. What do you both think?

Ben's source: https://twitter.com/hswolff/status/1133759319571345408

packages/website/src/helpers/dates.ts
```diff
@@ -39,3 +40,57 @@ export const findChartPeriod = (range: Range) => {
...
+  utcTime?: string | null,
+  timeZone?: string | null,
+  options?: Intl.DateTimeFormatOptions
+): string | null | undefined => {
```
SimplyBallistic, line 49: maybe returning just null or undefined would be nice

SimplyBallistic (commented):
Looks good overall! Just a few small suggestions :+1:

packages/website/src/helpers/dates.ts
tleraj: The return type should be the same as the `utcTime` variable type, and is usually a survey dive date, which is defined as `string | null | undefined` (type `SurveyState` in `src/store/Survey/types`).

packages/website/src/common/SurveyForm/index.tsx
tleraj: I guess object default values are OK, but it doesn't look bad as it is now to me. It's your call @ericboucher :wink:

packages/website/src/helpers/dates.ts
ericboucher: Maybe we could create an intermediate Type to make it clearer then?

ericboucher (approved):

Issue #354, opened 2020-11-05
Combine Survey View and Reef Details Chart Components to one component


Ry-DS:
After #337 is merged, the charts will look like this:

| Reef Details | Surveys |
| ------------- | ------------- |
| ![image](https://user-images.githubusercontent.com/25571542/98193552-9d8d9800-1f71-11eb-84fd-3538d1bef99f.png) | ![image](https://user-images.githubusercontent.com/25571542/98193567-a67e6980-1f71-11eb-9b3b-721d16ce4d85.png) |

They are slightly inconsistent and could share code with eachother. A new component to do this would  be great.

PR #368, opened 2020-11-23
Unify charts

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/354.


packages/website/src/common/CombinedCharts/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+  onDateChange,
+  pickerDate,
+  spotterData,
+  timeZone,
+  startDate,
+  endDate,
+  chartPeriod,
+  classes,
+}: CombinedChartsProps) => {
+  const [open, setOpen] = useState<boolean>(false);
```
SimplyBallistic, line 43: Type could be implicit, but this is fine too.

packages/website/src/common/CombinedCharts/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+            display="flex"
+            alignItems="center"
+            justifyContent="center"
+            textAlign="center"
+            p={4}
+          >
+            <CircularProgress size="6rem" thickness={1} />
+          </Box>
+        ) : (
+          (spotterData && spotterData.bottomTemperature.length > 1 && (
```
SimplyBallistic, line 87: Optional chaining might work here?

packages/website/src/common/CombinedCharts/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+            open={open}
+            onClose={() => setOpen(false)}
+            onOpen={() => setOpen(true)}
+            value={range}
+            onRangeChange={onRangeChange}
+          />
+          <DatePicker value={pickerDate} onChange={onDateChange} />
+        </Grid>
+      )}
+      {hasSpotterData &&
```
SimplyBallistic, line 74: This is a little complex, maybe splitting these 3 conditional components to 3 booleans at the top would simplify things? Not too big of a concern, but might be worthwhile since not much is happening anyway in this function.

packages/website/src/common/CombinedCharts/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+      >
+        <Typography className={classes.graphTitle} variant="h6">
+          DAILY WATER TEMPERATURE (°C)
+        </Typography>
+      </ChartWithTooltip>
+      {hasSpotterData && (
+        <Grid container alignItems="baseline" spacing={3}>
+          <SelectRange
+            open={open}
+            onClose={() => setOpen(false)}
```
SimplyBallistic, line 66: its unusual to me why there isn't a `setOpen` instead of `onClose` and `onOpen` but this could be left as is.

SimplyBallistic (commented):
Looks good overall :+1: Maybe these suggestions might help improve this PR a little...

SimplyBallistic:
Maybe this component could be moved to `common/Charts`?

packages/website/src/common/CombinedCharts/index.tsx
tleraj: I thought so too, but `spotterData?.bottomTemperature.length` gives an `Object is possibly 'undefined'` error.

packages/website/src/common/Chart/CombinedCharts.tsx
```diff
@@ -0,0 +1,169 @@
...
+  timeZone,
+  startDate,
+  endDate,
+  chartPeriod,
+  classes,
+}: CombinedChartsProps) => {
+  const [open, setOpen] = useState<boolean>(false);
+  const spotterDataLoading = useSelector(reefSpotterDataLoadingSelector);
+
+  const spotterComponent = useCallback(() => {
```
SimplyBallistic, line 44: When I mentioned simplifying the logic here, I was moreso thinking of having 3 booleans with simple, expressive names, so we know the 3 states this conditional can be in just by reading them. 
For example:
`const doShowLoading=....`
`const doShowSpotterData=...`
`const doShow...=....`
then when rendering:
`{doShowLoading &&<.../>}`
`{doShow... && <.../>}`
etc. 

So we don't have to do the logic in our heads when we read this haha. 

I'm not sure about this callback however, as when I see this hook I think its a utility function, not a component.
If you think a component is best, maybe moving it outside and making it a proper component with its own props and everything would be nice. However if you think this would be a mess (there are quite a lot of props you'd have to pass down) my approach might work well too, you'll just have to think of some good logic for the booleans.

SimplyBallistic: For the boolean approach, since the components are mutually exclusive, you might need to add extra logic to those 3 booleans to make sure only one ever will be true. Otherwise, if thats too complicated, you could just use the logic you have right now for the 3 booleans then do: 
`doShow... ? <.../> : doShow... ? <.../> : <../>` Its a nested ternary, but since the boolean names make it easy to read, I think it won't be an issue :) Eric can weigh in on this.

packages/website/src/common/Chart/CombinedCharts.tsx
```diff
@@ -0,0 +1,169 @@
...
+        temperatureThreshold={temperatureThreshold}
+        maxMonthlyMean={maxMonthlyMean}
+        background
+        className={classes.chart}
+      >
+        <Typography className={classes.graphTitle} variant="h6">
+          DAILY WATER TEMPERATURE (°C)
+        </Typography>
+      </ChartWithTooltip>
+      {hasSpotterData && (
```
SimplyBallistic, line 119: this component and the one below it both render on the same condition. You could combine them to a Fragment.

packages/website/src/common/Chart/CombinedCharts.tsx
```diff
@@ -0,0 +1,169 @@
...
+        <Grid container alignItems="baseline" spacing={3}>
+          <SelectRange
+            open={open}
+            onClose={() => setOpen(false)}
```
SimplyBallistic, line 123: we could replace `onClose` and `onOpen` with `setOpen`, so we don't need to pass two functions but just `setOpen`?

packages/website/src/common/Chart/CombinedCharts.tsx
SimplyBallistic: After a second look, I think we can just remove the `useCallback` and just rename `spotterComponent` to something more descriptive and in `PascalCase` so we can wrap this PR up :+1:

`useCallback` would be perfect here if we were passing this component as a dependancy, but in its current state `useCallback` would likely actually worsen performance than improve it.

tleraj: @SimplyBallistic So I ended up using the boolean logic you mentioned above. I used the names `spotterDataLoading`, `spotterDataSuccess` and `spotterDataError`. Let me know if they sound good :slightly_smiling_face:

Ry-DS: Looks good! I would just add a prefix to the boolean names like `*is*SpotterDataLoading`

packages/website/src/common/CombinedCharts/index.tsx
Ry-DS: Ah I see, its because you cant compare `undefined` which could be returned by an optional chain with a number according to TS. We could leave this then :+1:

Ry-DS (approved):
Looks great! I did a quick commit to rename the booleans and just pass `setOpen` without a new function. Let me know if that's good with you :+1:

Issue #352, opened 2020-11-03
Improve responsiveness through mui theme configuration


ericboucher:
To make our UI more responsive and look better on Mobile with less style tweaks, we should configure our Typography and Box Padding defaults.

Eg for Typography: https://material-ui.com/customization/typography/#responsive-font-sizes

PR #372, opened 2020-11-25
Improve responsiveness through MUI theme configuration

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/352.

UI Changes:

| Before | After |
|---|---|
| Screenshot_20201201_144022 | Screenshot_20201201_144103 |
| Screenshot_20201201_171203 | Screenshot_20201216_140626 |


packages/website/src/layout/App/theme.ts
```diff
@@ -192,4 +195,24 @@ const theme: any = createMuiTheme({
   },
 });
 
+// eslint-disable-next-line fp/no-mutation
```
Ry-DS, line 198: A better way to do this might be to create an empty theme `const breakpointTheme = createMuiTheme({})` and just use that in the above theme whenever you need breakpoints.

packages/website/src/routes/Landing/index.tsx
```diff
@@ -113,6 +113,9 @@ const styles = (theme: Theme) =>
       backgroundSize: "cover",
       left: "10rem",
       height: "54rem",
+      [theme.breakpoints.down("xs")]: {
+        height: "36rem",
```
Ry-DS, line 117: Not too sure if this is a big concern, but wouldn't it be better to use a unit besides `rem` or `em`? Since these units are for font sizes, don't see why the image should be relative to those. Maybe a `%` or a set `px` would work better so the image height is more predictable.

packages/website/src/routes/Surveys/View/SurveyDetails.tsx
```diff
@@ -23,7 +23,13 @@ const SurveyDetails = ({ reef, survey, classes }: SurveyDetailsProps) => {
   const nImages = getNumberOfImages(survey?.surveyMedia || []);
   const { region: regionName } = getReefNameAndRegion(reef);
   return (
-    <Grid style={{ marginTop: "1rem" }} container item xs={12} direction="row">
+    <Grid
+      style={{ marginTop: "1rem" }}
```
Ry-DS, line 27: Unless the grid is closely related to text (there's text above it maybe) it might be best to do a set value so the value is more predictable. If the global font size changes, this will change too.

Ry-DS (changes requested):
Just a few suggestions. @ericboucher feel free to weigh in on these if you think they aren't too concerning!

packages/website/src/layout/App/theme.ts
Ry-DS: Found a better solution to mine on stack overflow: https://stackoverflow.com/a/56370022/5279269

Ry-DS:
have this deployed? Since there's a global mui style change, would be nice to make sure everything still looks good on mobile :+1:
Or if you can confirm this yourself that would be great too :)

ericboucher:
@Ry-DS just deployed on staging - https://aqualink-app-staging.web.app/

Ry-DS:
> @Ry-DS just deployed on staging - https://aqualink-app-staging.web.app/. It might take a hard reload to update.

Looks good overall! Here are some small fixes that could be done on the `/drones` page:
| 1 | 2 |
| - | - |
| image | image |

The blue cards could be centered maybe?

packages/website/src/routes/Landing/index.tsx
```diff
@@ -9,19 +9,61 @@ import {
...
   return (
     <>
       <NavBar routeButtons searchLocation={false} />
+      {scrollPosition === 0 && isMobile && (
+        <Box
+          width="100%"
+          display="flex"
+          justifyContent="flex-end"
+          position="absolute"
+          top="calc(100vh - 64px)"
```
Ry-DS, line 59: Is it a coincidence that this and line 158 match exactly? If not, introducing a variable to store this value or maybe just a small comment explaining why they are the same would be great!

packages/website/src/routes/Landing/index.tsx
```diff
@@ -9,19 +9,61 @@ import {
...
 const LandingPage = ({ classes }: LandingPageProps) => {
+  const [scrollPosition, setScrollPosition] = useState(0);
+  const windowSize = useWindowSize();
+  const isMobile = windowSize && windowSize.width < MOBILE_SIZE;
+
+  const handleScroll = useCallback(() => {
+    setScrollPosition(window.pageYOffset);
+  }, []);
+
+  const seeMore = useCallback(() => {
```
Ry-DS, line 35: `useCallback` should *never* be used unless you're passing the function to a dependency array; it may actually decrease performance if its not used with `useEffect` or anything else that uses dependency arrays :)

packages/website/src/routes/Landing/index.tsx
```diff
@@ -9,19 +9,61 @@ import {
...
 import landingPageImage from "../../assets/img/landing-page/header.jpg";
 import { cardTitles } from "./titles";
+import { useWindowSize } from "../../helpers/useWindowSize";
+
+const MOBILE_SIZE = 600;
 
 const LandingPage = ({ classes }: LandingPageProps) => {
+  const [scrollPosition, setScrollPosition] = useState(0);
+  const windowSize = useWindowSize();
+  const isMobile = windowSize && windowSize.width < MOBILE_SIZE;
```
Ry-DS, line 29: Optional Chaining could be used here, but you might be replacing this anyway from the above comment...

packages/website/src/routes/Landing/index.tsx
```diff
@@ -9,19 +9,61 @@ import {
...
     <>
       <NavBar routeButtons searchLocation={false} />
+      {scrollPosition === 0 && isMobile && (
+        <Box
+          width="100%"
+          display="flex"
+          justifyContent="flex-end"
+          position="absolute"
+          top="calc(100vh - 64px)"
+          padding="0 10px 0 10px"
```
Ry-DS, line 60: Would `padding: 0 10px` work here?

packages/website/src/routes/Landing/index.tsx
```diff
@@ -9,19 +9,61 @@ import {
...
 import { Link } from "react-router-dom";
 
 import NavBar from "../../common/NavBar";
 import Footer from "../../common/Footer";
 import Card from "./Card";
 import landingPageImage from "../../assets/img/landing-page/header.jpg";
 import { cardTitles } from "./titles";
+import { useWindowSize } from "../../helpers/useWindowSize";
+
+const MOBILE_SIZE = 600;
```
Ry-DS, line 24: You should use this: https://material-ui.com/components/use-media-query/
instead of a hard coded mobile size.

packages/website/src/routes/Landing/index.tsx
```diff
@@ -9,19 +9,61 @@ import {
...
 
 const LandingPage = ({ classes }: LandingPageProps) => {
+  const [scrollPosition, setScrollPosition] = useState(0);
+  const windowSize = useWindowSize();
+  const isMobile = windowSize && windowSize.width < MOBILE_SIZE;
+
+  const handleScroll = useCallback(() => {
```
Ry-DS, line 31: You could just define this inside the `useEffect` itself, since (I think) its not used anywhere else, and remove `useCallback`. Or just remove `useCallback` since the performance impact would be negligible.

Ry-DS (changes requested):
A few suggestions :+1:

packages/website/src/routes/Landing/index.tsx
tleraj: It just so happens here that the button and the navbar have roughly the same height. Here I place the button at the bottom of the screen (so `100vh` minus the button's height which is `56px` but I add a little more so that it doesn't touch the bottom edge of the screen). In line 158, I want the image's height to be `100vh` minus the height of the navbar. I added comments on both just to be clear :slightly_smiling_face:

packages/website/src/routes/Surveys/View/SurveyDetails.tsx
```diff
@@ -23,7 +23,13 @@ const SurveyDetails = ({ reef, survey, classes }: SurveyDetailsProps) => {
   const nImages = getNumberOfImages(survey?.surveyMedia || []);
   const { region: regionName } = getReefNameAndRegion(reef);
   return (
-    <Grid style={{ marginTop: "1rem" }} container item xs={12} direction="row">
+    <Grid
+      style={{ marginTop: 16 }}
```
Ry-DS, line 27: Can't remember off the top of my head if `Grid` supports prop spacing like `Box` does. If possible, replacing this with `<Grid marginTop={X}\>` would be great. Otherwise, this is fine too :+1:

packages/website/src/routes/Landing/index.tsx
```diff
@@ -19,9 +23,46 @@ import landingPageImage from "../../assets/img/landing-page/header.jpg";
...
+      {scrollPosition === 0 && isMobile && (
+        <Box
+          width="100%"
+          display="flex"
+          justifyContent="flex-end"
+          position="fixed"
+          bottom="10px"
+          padding="0 10px"
+        >
+          <Fab onClick={seeMore} size="large">
```
Ry-DS, line 61: Out of curiosity, how do I trigger this button? I tried the link Eric sent on mobile but couldn't find this button. Is the link outdated? Otherwise if it works fine on your end all is good :+1:

Ry-DS (approved):
Looks good!

packages/website/src/routes/Landing/index.tsx
tleraj: I noticed this too, that's why I changed position from `absolute` to `fixed` and set it's `bottom` to `10px`. So the content in the link is outdated. We need a re-deployment :+1:

ericboucher: Deployed on staging, I still don't see it though

tleraj: I'm not sure why it doesn't work for you. It appears to work for me:
| My device | Browser responsive |
|---|---|
| Screenshot_2020-12-22-15-28-38-159_com android chrome  | Screenshot_20201222_153308 |

tleraj: I added a `zIndex` of 100 so now it should be working

ericboucher (approved):

PR #373, opened 2020-11-25
Refresh firebase token on expiration only

tleraj:
The purpose of this PR is to handle a possible 401 infinite loop. With these changes, a request for a new token is made only if the previous one has expired. Otherwise, we get a 401 error without infinitely trying to acquire a new token.


packages/website/src/index.tsx
```diff
@@ -18,18 +19,24 @@ app.auth().onAuthStateChanged((user) => {
...
+          const decoded = jwt.decode(oldToken) as { exp: number };
+          const now = new Date().getTime();
+          if (config && status === 401 && decoded.exp < now) {
+            // Unauthorized due to token expiration
+            const newToken = await user.getIdToken();
+            store.dispatch(setToken(newToken));
+            const newConfig = {
+              ...config,
+              headers: {
+                ...config.headers,
```
ericboucher, line 33: It does the trick for now but I am wondering if the API should be returning two different error codes? Thoughts @tleraj, @avalmas-programize ?

ericboucher (approved):

Issue #370, opened 2020-11-24
Firebase API Key doesn't appear to be optional


spiderkeys:
In the website package's README.md, it mentions that the REACT_APP_FIREBASE_API_KEY is optional. However, after running the app with `yarn start` with only the REACT_APP_API_BASE_URL variable specified, the page throws an uncaught exception when opened in the browser to the effect of:

```
{code: "auth/invalid-api-key", message: "Your API key is invalid, please check you have copied it correctly.", a: null}
```
from:
```
Provider.getImmediate | @ | provider.ts:108
FirebaseAppImpl._getService | @ | firebaseApp.ts:127
```
After setting up Firebase, creating an API key, and setting the variable, the page correctly loaded in the browser.

ericboucher:
Nice catch @spiderkeys! I think we should aim for the development to work without `REACT_APP_FIREBASE_API_KEY` to make it as easy as possible to interact with the app.

We could also specify a default `REACT_APP_API_BASE_URL` directly in the README or the sample .env. @spiderkeys which BASE_URL did you end up using?

spiderkeys:
@ericboucher I just used localhost as the BASE_URL, as suggested in the current README:
```
Required
   REACT_APP_API_BASE_URL="http://localhost:8080/"
```
This seems like a sane default to me, as many people are going to want to test locally. 

I agree that being able to run the site without a Firebase backend available is a good idea. For reference, my steps right now to bring up the site in a functional manner locally are:
```
cd packages/website
REACT_APP_API_BASE_URL="http://localhost:8080/" REACT_APP_FIREBASE_API_KEY="${FIREBASE_API_KEY}" yarn start
```

PR #377, opened 2020-12-02
Optional Firebase API Key

tleraj:
This PR is scheduled to fix https://github.com/aqualinkorg/aqualink-app/issues/370.


packages/website/src/firebase.ts
```diff
@@ -1,8 +1,12 @@
 import * as firebase from "firebase/app";
 import "firebase/auth";
 
-const app = firebase.initializeApp({
-  apiKey: process.env.REACT_APP_FIREBASE_API_KEY,
-});
+const { REACT_APP_FIREBASE_API_KEY: apiKey } = process.env;
+
+const app = apiKey
```
ericboucher, line 6: Let's add a console warning in here if the apiKey is missing. Maybe something like this:

`No Firebase API Key provided. Please set up Firebase to use login and admin functionalities.`

ericboucher (approved):

PR #379, opened 2020-12-03
Fix timezone shortname behavior

ericboucher:
Javascript was defaulting to the users's local shortname for the timezone and passing the full date with timezone included was breaking the charts.

Let me know if you guys can think of a better approach @SimplyBallistic @tleraj


packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -60,14 +61,20 @@ const Tooltip = ({
...
     month: "2-digit",
     year: "2-digit",
     ...(spotterSurfaceTemp ? hourlyOptions : {}),
   });
+  const timeZoneName = new Date(date)
+    .toLocaleString("en", {
+      timeZone: timeZone || "UTC",
+      timeZoneName: "short",
+    })
+    .split(" ")[3];
```
SimplyBallistic, line 76: I don't like the idea of using this method since it seems kinda hacky

If you don't mind adding a new library, this solution is elegant: 
https://momentjs.com/timezone/
then:
`moment().tz('America/Los_Angeles').format('[GMT]Z');`
Otherwise, maybe we could use:
```
const dateString = moment(new Date().toLocaleString("en", {
    // example
    timeZone: 'Australia/Yancowinna',

})).format('[GMT]Z');
```
this is assuming the `timeZone` variable is in this format: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -90,6 +97,7 @@ const Tooltip = ({
           title={
             <Typography color="textPrimary" variant="caption">
               {dateString}
+              {spotterSurfaceTemp && ` ${timeZoneName}`}
```
SimplyBallistic, line 100: I'm not sure if a 0 temperature is likely, but when working with numbers, checking existance like this isn't too great. Maybe using a lodash function like `isNull` might be good? Otherwise if `0` will never happen this is fine too.

packages/website/src/routes/Surveys/View/index.tsx
```diff
@@ -270,6 +271,7 @@ const SurveyViewPage = ({ reef, surveyId, classes }: SurveyViewPageProps) => {
                           chartPeriod={findChartPeriod(range)}
                           surveys={[]}
                           depth={reef.depth}
```
SimplyBallistic, line 273: Ignore if this isn't practical or out of scope, but as chart accepts more variables which are just an extension of `reef`, it might be worthwhile to just pass reef as a whole if we don't use this component without one.

packages/website/src/common/Chart/Tooltip/index.tsx
ericboucher: I am not a huge fan either, but the problem with the approach ou propose is that we actually don' want to have GMT all the time since some timezones are "AST, PST, ..." and I didn't find a clever way to get just the timezone part yet.

tleraj (approved):

packages/website/src/common/Chart/Tooltip/index.tsx
ericboucher: @tleraj any better idea here?

SimplyBallistic: After playing around a bit moment timezone does seem to support the abbreviations quite well:
image
Then if we have one that doesn't have an abbreviation...
image
we just do a simple `const needsGMT=str.includes('+') || s.includes('-')` and prefix it if there's a plus or minus. What do you think @ericboucher?

packages/website/package.json
```diff
@@ -42,6 +42,7 @@
     "lodash": "^4.17.15",
     "material-table": "^1.65.0",
     "moment": "^2.27.0",
```
SimplyBallistic, line 44: Think we can remove moment and replace uses of it with `moment-timezone`? I think `moment-timezone` uses `moment` under the hood, so they should theoretically be the same? 
Honestly however the impact including both is likely minimal and we could just keep both :+1:

packages/website/src/helpers/dates.ts
```diff
@@ -96,3 +100,10 @@ export const convertSurveysToLocalTime = (
     ...item,
     diveDate: convertToLocalTime(item.diveDate, timeZone) || item.diveDate,
   }));
+
+export const getTimeZoneName = (timeZone: string): string => {
+  const rawTimeZoneName = moment().tz(timeZone).format("z");
+  const needsGMT =
+    rawTimeZoneName.includes("+") || rawTimeZoneName.includes("-");
```
SimplyBallistic, line 107: We both know what this does, but think its obvious enough for an outsider without a comment? If so, we can leave as is :)

packages/website/src/helpers/dates.ts
```diff
@@ -96,3 +100,10 @@ export const convertSurveysToLocalTime = (
...
+  return `${needsGMT ? "GMT" : ""}${rawTimeZoneName}`;
```
SimplyBallistic, line 108: Looks good without a space? I haven't seen it rendered but if it doesn't need a space all good

SimplyBallistic (approved):

Issue #389, opened 2020-12-16
Filter table and map to sites with a spotter only


ericboucher:
Add a simple toggle for super admins between the details box and the table.

`has spotter` to filter down the table and map to only display sites with a linked spotter.

PR #395, opened 2020-12-22
Filter sites with spotter

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/389.
 - [x] Filter on desktop
 - [x] Filter on mobile

Designs:

WEB
| off | on |
|---|---|
| Screenshot_20201222_194521 | Screenshot_20201222_194541 |

MOBILE
| off | on |
|---|---|
| screencapture-localhost-3000-map-2020-12-23-13_18_24 | screencapture-localhost-3000-map-2020-12-23-13_17_58 |


ericboucher:
Let's just use a switch - https://material-ui.com/components/switches/
And say `spotters only`
You can use the same for mobile since it only shows for super admins anyways

packages/website/src/routes/HomeMap/ReefTable/index.tsx
```diff
@@ -77,17 +80,19 @@ const ReefTable = ({ openDrawer, classes }: ReefTableProps) => {
...
-            justifyContent="flex-end"
-          >
-            <Switch onChange={toggleSwitch} color="primary" />
-            <Typography color="textSecondary" variant="h6">
-              spotters only
-            </Typography>
-          </Box>
+          {isSuperAdmin(user) && (
+            <Box
+              padding="0 40px"
```
ericboucher, line 85: let's use class styling, no?

packages/website/src/mocks/mockUser.ts
```diff
@@ -1,7 +1,7 @@
 import { User } from "../store/User/types";
 
 export const mockUser: User = {
-  adminLevel: "default",
+  adminLevel: "super_admin",
```
ericboucher, line 4: let's keep the default user with a default adminLevel. Our preference is to test what normal users will see.

packages/website/src/routes/HomeMap/ReefTable/index.tsx
tleraj: I think it's not a problem, since `Box` accepts these values as props. Either way is fine.

ericboucher:
Let's make sure it's really right-aligned, i.e. the `y` of only is aligned with the right of the table

packages/website/src/routes/HomeMap/ReefTable/index.tsx
```diff
@@ -12,23 +12,36 @@ import {
...
+import {
+  reefsListLoadingSelector,
+  filterReefsWithSpotter,
+} from "../../../store/Reefs/reefsListSlice";
 import EnhancedTableHead from "./tableHead";
 import { useWindowSize } from "../../../helpers/useWindowSize";
+import { userInfoSelector } from "../../../store/User/userSlice";
+import { isSuperAdmin } from "../../../helpers/user";
+import {
+  onlyWithSpotterSelector,
```
ericboucher, line 33: let's rename all these to -> withSpotterOnly instead of onlyWithSpotter

packages/website/src/routes/HomeMap/ReefTable/index.tsx
```diff
@@ -42,6 +55,20 @@ const ReefTable = ({ openDrawer, classes }: ReefTableProps) => {
...
 
+  const toggleSwitch = (event: ChangeEvent<HTMLInputElement>) => {
+    const {
+      target: { checked },
+    } = event;
+    dispatch(filterReefsWithSpotter(checked));
+    dispatch(setOnlyWithSpotter(checked));
+  };
+
+  const onSwitchClick = (
```
ericboucher, line 66: I don't think we actually need to set the onClick here

packages/website/src/store/Reefs/reefsListSlice.ts
```diff
@@ -32,14 +33,22 @@ export const reefsRequest = createAsyncThunk<
 const reefsListSlice = createSlice({
   name: "reefsList",
   initialState: reefsListInitialState,
-  reducers: {},
+  reducers: {
+    filterReefsWithSpotter: (state, action: PayloadAction<boolean>) => ({
```
ericboucher, line 37: are you sure we need this new reducer? Maybe we could just use the new `withSpotterOnly` and have a util function `filterReefWithSpotters` instead of a reducer that gets called on the reefsList for the markers and for the table. What do you think?

tleraj: This was my initial approach, to use a `filterReefsWithSpotters` util, but it would require to be called twice (once for the map markers and once for the reef table), which would be time consuming for a large amount of reefs.

packages/website/src/routes/HomeMap/ReefTable/index.tsx
tleraj: Actually we need this on mobile. On mobile, the whole drawer is clickable, and clicking anywhere on the drawer (including the switch) would cause the drawer to close.If we click on the switch, the drawer will close without changing the state of the switch (from `true` to `false` and vice versa). That's why we need this `event.stopPropagation()`, in order to prevent the drawer close.

ericboucher: Ok, let's add a comment then to explain this behavior

ericboucher (approved):

Issue #390, opened 2020-12-16
Add update info to metric cards 


ericboucher:
We would like to make it clearer to users when the data was last updated.

For satellite data and DHW we should inform the user about the update frequency
For spotter, we should display the last data received time (eg. `data received 5min ago`)

  // Sofar variables used by Aqualink
  
  https://polar.ncep.noaa.gov/waves/
  SofarModels.NOAAOperationalWaveModel: {
    significantWaveHeight: 'significantWaveHeight',
    meanDirectionWindWaves: 'meanDirectionWindWaves',
    peakPeriod: 'peakPeriod',
  }
  
  SofarModels.NOAACoralReefWatch: {
    degreeHeatingWeek: 'degreeHeatingWeek',
    analysedSeaSurfaceTemperature: 'analysedSeaSurfaceTemperature',
  }

  SofarModels.GFS: {
    magnitude10MeterWind: 'magnitude10MeterWind',
    direction10MeterWind: 'direction10MeterWind',
  }
  
  
Designs to come soon - @drewjgray

drewjgray:
@ericboucher What is the update frequency currently for the three data sources?

drewjgray:
Here are the designs. The update frequencies are placeholders, we should double-check the data apis. 

<img width="576" alt="Screen Shot 2020-12-17 at 9 25 08 AM" src="https://user-images.githubusercontent.com/6283065/102512259-71c10000-404f-11eb-828f-8514da7ee90e.png">
<img width="576" alt="Screen Shot 2020-12-17 at 9 25 16 AM" src="https://user-images.githubusercontent.com/6283065/102512262-72599680-404f-11eb-926a-6548094b5e1a.png">
<img width="572" alt="Screen Shot 2020-12-17 at 9 25 24 AM" src="https://user-images.githubusercontent.com/6283065/102512265-72f22d00-404f-11eb-9e45-8bd0fc0513d6.png">
<img width="405" alt="Screen Shot 2020-12-17 at 9 36 36 AM" src="https://user-images.githubusercontent.com/6283065/102512266-72f22d00-404f-11eb-9069-b61f9f7b59c4.png">

drewjgray:
@ericboucher notice I deleted the icons in the top right of the sat and buoy card and made the titles on one line.

PR #398, opened 2020-12-23
Add update info

tleraj:
The purpose of this PR is to implement the designs and close https://github.com/aqualinkorg/aqualink-app/issues/390.
 - [x] Satellite observation card
 - [x] Sensor observation card
 - [x] Heat stress card
 - [x] Wind/Waves card
 - [x]  Add links

Designs:

With spotter
|---|---|---|
| Screenshot_20210111_171524 | Screenshot_20210111_171548 | Screenshot_20210111_171644 |

Without spotter
|---|---|---|
| Screenshot_20210112_124724 | Screenshot_20210112_124749 | Screenshot_20210112_124850 |


packages/website/src/helpers/dates.ts
```diff
@@ -115,3 +115,31 @@ export const getTimeZoneName = (timeZone: string): string => {
     rawTimeZoneName.includes("+") || rawTimeZoneName.includes("-");
   return `${needsGMT ? "GMT" : ""}${rawTimeZoneName}`;
 };
+
+export const timeAgo = (timestamp?: string) => {
```
ericboucher, line 119: Something like this `DateTime.fromMillis(lastActivity).toRelative()` should be enough for our use case I think

ericboucher:
@tleraj for the last card (wind waves), we either have a spotter and use the "live" tooltip, or we don't have a spotter and use the GFSS one. We should never need both at the same time.

packages/website/src/helpers/dates.ts
tleraj: Are you suggesting this https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html?

ericboucher:
@tleraj - for the links let's use this for now:

NOAA SST -> https://coralreefwatch.noaa.gov/

NOAA CRW -> https://coralreefwatch.noaa.gov/product/5km/index_5km_baa_max_r07d.php

GFS -> https://www.ncdc.noaa.gov/data-access/model-data/model-datasets/global-forcast-system-gfs

packages/website/src/routes/ReefRoutes/Reef/Sensor/__snapshots__/index.test.tsx.snap
```diff
@@ -99,6 +90,70 @@ exports[`Sensor Card should render with given state from Redux store 1`] = `
...
+              <mock-box
+                display="flex"
+                flexdirection="column"
+              >
+                <mock-typography
+                  variant="caption"
+                >
+                  Last data received
+                   
+                  4490 days ago
```
ericboucher, line 128: You need to make sure these numbers are not gonna change every day

ericboucher:
- Let's improve the spacing in and around the grey buttons
- Let's make sure the button always stays on the right side, I think you should make a row grid with the text on the left and the button on the right.
- The clock icon should be more to the right
- Let's avoid having 3 lines (I think moving the clock icon should fix this, otherwise make the text a bit smaller)
 
Screen Shot 2021-01-05 at 10 53 25 AM
Screen Shot 2021-01-05 at 10 49 08 AM

ericboucher:
When there is no spotter, let's move the yellow button into the same new info box at the bottom, a bit like this:

Screen Shot 2021-01-11 at 5 40 18 PM

packages/website/src/common/UpdateInfo/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+            )}
+          </Link>
+        )}
+      </Grid>
+    </Grid>
+  </Grid>
+);
+
+const styles = (theme: Theme) => {
+  const maxWidth = 1080;
```
ericboucher, line 82: What are we trying to achieve with this? This looks a bit too specific to me to be using `between("md", maxWidth)`

packages/website/src/routes/ReefRoutes/Reef/Sensor/index.tsx
```diff
@@ -8,60 +8,57 @@ import {
...
     default:
-      return "Not Installed Yet";
+      return ["Not Installed Yet", false];
   }
 };
 
 const Sensor = ({ reef, classes }: SensorProps) => {
   const { surfaceTemperature, bottomTemperature } = reef.liveData;
 
+  const ago = timeAgo(
```
ericboucher, line 58: let's use the term `relativeDate` instead of `ago` and rename `timeAgo` to `toRelativeDate`

ericboucher (approved):

Issue #386, opened 2020-12-16
Use latest data for temperature on map page table, detail box and popup


ericboucher:
To avoid differences in the data being displayed in the details box and on the site page, we would like to use the live data on the map page for sensor temperature in site details box and popup.

<img width="908" alt="Screen Shot 2020-12-16 at 10 59 36 AM" src="https://user-images.githubusercontent.com/16843267/102393886-d156d700-3f8d-11eb-9e5a-16bcefb7a430.png">

ericboucher:
Now that we have the "live data" in our DB, we can use the latestData view to achieve this.

ericboucher:
And for SST in the table and the box as well

PR #402, opened 2021-01-05
[HOLD] Use live data on popup and selected reef card

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/386.


ericboucher:
This will be replaced by an implementation using the "latest_data" view of the timeseries table

PR #413, opened 2021-01-12
Improve Register Page

khattakdev:
@ericboucher  Please review the design and let me know if you want any changes
design review


ericboucher:
Looks good @khattakdev, can we have the `Sign up` prompt only appear when the user is not logged in?

ericboucher:
We could also have the same message as a tooltip on the greyed-out `submit` button, what do you think?

packages/website/src/routes/RegisterSite/index.tsx
```diff
@@ -205,6 +213,16 @@ const Apply = ({ classes }: ApplyProps) => {
                     <Typography variant="h4" gutterBottom>
                       Site Information
                     </Typography>
+                    <Typography variant="h6" gutterBottom>
```
tleraj, line 216: Maybe we should include this `<Typography>` into a `{ !user && ( ... ) }`, so that it shows up only when we have a not signed-in user

packages/website/src/routes/RegisterSite/index.tsx
```diff
@@ -161,7 +169,7 @@ const Apply = ({ classes }: ApplyProps) => {
   return (
     <>
       <NavBar searchLocation={false} />
-      <Box height="100%" pt={4}>
+      <Box className={classNames(classes.boxBar)} height="100%" pt={4}>
```
tleraj, line 172: Why not just `className={classes.boxBar}`?

tleraj (dismissed):
Looks good!! Just a fiew suggestions.

packages/website/src/routes/RegisterSite/index.tsx
khattakdev: Great idea! Let me change that

khattakdev:
> We could also have the same message as a tooltip on the greyed-out `submit` button, what do you think?

Do you mean to have the message below Site Information and on tooltip both?

packages/website/src/routes/RegisterSite/index.tsx
khattakdev: Make sense, let me update that

ericboucher:
@@khattakdev or @tleraj could you have a quick look at my comments and make the necessary changes? Thanks :)

khattakdev:
I've implemented those changes in the last commit.

ericboucher:
Awesome, I hadn't seen sorry!

packages/website/src/routes/RegisterSite/index.tsx
```diff
@@ -205,6 +212,18 @@ const Apply = ({ classes }: ApplyProps) => {
...
+                    {!user && (
+                      <Typography variant="h6" gutterBottom>
+                        Please
+                        <Button
+                          color="primary"
+                          onClick={() => handleRegisterDialog(true)}
+                        >
+                          Sign up
+                        </Button>
+                        before registering a new site
```
ericboucher, line 224: Let's do please sign up (registration button) or log in (sign in button) before registering a new site

ericboucher (approved):
Looks great, let's add a sign in link as well as I mention in the comment and then we're good to go!

khattakdev:
@ericboucher  I guess it's ready to merge.

UI

Test Suite

Issue #404, opened 2021-01-06
Add organization name on site detail page


ericboucher:
Add `Managed by OrganizationName` as a subtitle below the site name on the site detail page.

PR #415, opened 2021-01-13
Add organization name

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/404.
Design:

Screenshot_20210113_170452

Let me know if the font size is OK.


tleraj:
I noticed that `reef.admins` is an array of users. Is there a case that a reef is managed by multiple users, therefore multiple organizations? If yes, what should we show? For now I've used `reef.admins[0]`.

ericboucher (approved):

PR #416, opened 2021-01-13
Handle timezone

ericboucher:


ericboucher:
@tleraj the current implementation was not working properly with sites with recent surveys

tleraj:
@ericboucher Did you notice the issue on any specific timezone? Because until now I haven't been able to reproduce any issue.

ericboucher:
The issue we are facing is that the browser sometimes has problems converting back to datetime in `sortByDate` when doing the reordering. Hence we get a maxDate < minDate which results in an empty chart.

My hypothesis is that it only appears as an issue when there is surveys because otherwise chartJs is "smart enough" to detect that you mixed up max and min and is ok displaying it when you only have one line.

In any case, I don't think this is the right approach overall. Let's keep a normal datestring as long as we can and instead create a smart util function to localize the dates and times before we display them.

So instead, let's create one/two functions to use at the LAST moment to display dates, eg: displayTimeInTimezone and displayDateInTimezone following this logic:

const dateString = new Date(date).toLocaleString("en", {
    day: "2-digit",
    month: "2-digit",
    year: "2-digit",
    ...(hourlyData ? hourlyOptions : {}),
  });

  const timeZoneName = getTimeZoneName(timeZone || "UTC");

  return `${dateString}${hourlyData && timeZoneName && ` ${timeZoneName}`}
  
@tleraj I am going to close this PR and re-open the issue so that we can have a start fresh.

PR #418, opened 2021-01-14
Add deployed status to sites

ericboucher:
- Add deployed status flag to sites
- When filtering by spotters only, only display sites marked as deployed
- On the site details page, do not load sensor data for non-admins if the buoy is not deployed, we achieve this by only sending spotter data through the API if the buoy is deployed.


packages/api/src/utils/liveData.ts
```diff
@@ -87,21 +87,22 @@ export const getLiveData = async (
...
-        wavePeakPeriod: getLatestData(spotterRawData.wavePeakPeriod),
-        waveMeanDirection: getLatestData(spotterRawData.waveMeanDirection),
-        longitude:
-          spotterRawData.longitude && getLatestData(spotterRawData.longitude),
-        latitude:
-          spotterRawData.latitude && getLatestData(spotterRawData.latitude),
-      }
-    : {};
+  const spotterData =
+    spotterRawData && reef.status === 'deployed'
```
Ry-DS, line 91: Shouldn't we use the enum here to compare? Like `reef.status === ReefStatus.Deployed`

packages/website/src/routes/ReefRoutes/Reef/Sensor/index.tsx
```diff
@@ -47,6 +47,10 @@ const applicationTag = (
```
Ry-DS, line 47: While we are here, maybe its worthwhile to give this function a better name like `getApplicationTag` or `findApplicationTag` along with a docs comment? Not sure what the last boolean here represents at a glance.

Ry-DS (approved):
Overall looks good! Just a few small comments to improve readability

packages/api/src/utils/liveData.ts
```diff
@@ -87,21 +87,22 @@ export const getLiveData = async (
...
+      ? {
+          surfaceTemperature: getLatestData(spotterRawData.surfaceTemperature),
```
Ry-DS, line 93: I would personally use lodash's `mapValues` util function; it looks like we are just taking every field from `spotterRawData` and running it through `getLatestData`. What do you think?

ericboucher: You're thinking of something like this?

```
const spotterData =
    spotterRawData && reef.status === ReefStatus.Deployed
      ? mapValues(spotterRawData, (values) => values && getLatestData(values))
      : {};
```

Ry-DS: This works for me: 
```ts
const spotterData: { [k in keyof SpotterData]?: SofarValue } =
    spotterRawData && reef.status === ReefStatus.Deployed
      ? mapValues(spotterRawData, (values) => values && getLatestData(values))
      : {};
```

packages/api/src/reefs/reefs.service.ts
```diff
@@ -210,24 +210,26 @@ export class ReefsService {
...
-      );
+        reef.status === ReefStatus.Deployed &&
+        isNil(
+          await this.exclusionDatesRepository
+            .createQueryBuilder('exclusion')
+            .where('exclusion.spotter_id = :spotterId', {
+              spotterId: reef.spotterId,
+            })
+            .andWhere('DATE(exclusion.startDate) <= DATE(:date)', {
+              date: new Date(),
```
Ry-DS, line 223: `new Date()` is used 3 times, might be worthwhile to move it to `const now=new Date()`?

ericboucher: I think these functions will disappear very soon (we're changing how we handle exclusion dates)

Ry-DS: Sounds good :+1:

tleraj (approved):
Everything looks good!!

tleraj:
Noticed a text break on the sensor card on`960px` width.
Screenshot_20210118_160600

PR #420, opened 2021-01-14
Update header image

ericboucher:
Before
Screen Shot 2021-01-14 at 11 05 06 AM

After
Screen Shot 2021-01-14 at 11 04 54 AM


drewjgray:
@ericboucher Looks great.

PR #421, opened 2021-01-15
Get wind data from spotter when available

ericboucher:


drewjgray (approved):
Looks good.

packages/website/src/routes/ReefRoutes/Reef/Waves/index.tsx
```diff
@@ -68,7 +68,7 @@ const Waves = ({ liveData, classes }: WavesProps) => {
                   color="textSecondary"
                   variant="h3"
                 >
-                  {formatNumber(windSpeed?.value, 1)}
+                  {formatNumber(windSpeed?.value && windSpeed.value * 3.6, 1)}
```
Ry-DS, line 71: Maybe a comment to clarify what 3.6 means here? Or moving it to a named constant.

packages/api/src/utils/liveData.ts
```diff
@@ -94,8 +84,10 @@ export const getLiveData = async (
         significantWaveHeight: getLatestData(
           spotterRawData.significantWaveHeight,
         ),
-        wavePeakPeriod: getLatestData(spotterRawData.wavePeakPeriod),
-        waveMeanDirection: getLatestData(spotterRawData.waveMeanDirection),
+        wavePeriod: getLatestData(spotterRawData.wavePeakPeriod),
+        waveDirection: getLatestData(spotterRawData.waveMeanDirection),
```
Ry-DS, line 88: Note for the future: Merging #418 will cause a merge conflict, as we will likely use `mapValues` here.

packages/api/src/workers/dailyData.ts
```diff
@@ -195,9 +201,17 @@ export async function getDailyData(
   const wavePeriod =
     peakPeriodWindWaves && getAverage(peakPeriodWindWaves, true);
 
-  const minWindSpeed = windVelocities && getMin(windVelocities);
-  const maxWindSpeed = windVelocities && getMax(windVelocities);
-  const avgWindSpeed = windVelocities && getAverage(windVelocities);
+  // Get wind data if unavailable through a spotter
+  const windSpeeds =
+    spotterData.windSpeed.length > 0 ? spotterData.windSpeed : windSpeedsRaw;
```
Ry-DS, line 206: I would use lodash's `isEmpty` just because its a little neater, and also works well with `undefined` or `null` if that ever gets passed. Its very minor however and could be left as is.

Ry-DS (approved):
Looks good!

PR #422, opened 2021-01-18
Add unauthorized message for reef application

tleraj:
The purpose of this PR is to add a `Not authorized` message in the application page, for non reef admins who visit the `/reefs/{id}/apply` page. The designs implemented look like this:

| signed out | non reef admin | reef admin |
|---|---|---|
| Screenshot_20210126_150642 | Screenshot_20210126_150703 | Screenshot_20210126_150725 |


ericboucher:
@tleraj let's have two messages:

1. If the user is not signed up: `You need to sign up to access this page` and the buttons should be `Log in / sign up`
2. `You are not authorized to access this page. If this is an error, contact info@aqualink.org` and the button should be `Back to site`

tleraj:
I noticed that the reef application is accessible by any signed in user. If this is something that makes sense, then there's no point in displaying the `You are not authorized to access this page. If this is an error, contact info@aqualink.org`. This message would only make sense if a signed in user who is not a reef admin tries to submit an application (in that case I tweaked the message to `You are not authorized to execute this action. If this is an error, contact info@aqualink.org`). I think that it doesn't make sense to allow him this operation, so the submit button should be disabled and do not display an unauthorized message at all. Or, we might change the backend to restrict non reef admins from accessing the reef application. @ericboucher Let me know what you think.

ericboucher:
TODO
- [ ] hide the application to logged-in users who are not admins

ericboucher (approved):

Issue #387, opened 2020-12-16
Add exclusion date section for reef admins


ericboucher:
If there is a spotter, add a time selector on the right side with the following notice:
`Spotter data before this date will not be displayed`

<img width="1467" alt="Screen Shot 2020-12-16 at 9 43 22 AM" src="https://user-images.githubusercontent.com/16843267/102394179-38748b80-3f8e-11eb-8dfc-381ba60534a4.png">

- The time is selected in the site timezone (same as dive time)
- This should add an exclusion period of 1-month ending with this date

ericboucher:
@avalmas-programize let's update the exclusion "date" paradigm to take times instead of just dates.

avalmas-programize:
I have some questions:
- Does this mean that a spotter has a unique set of (start, end) dates?
- Should we add a relationship between the exclusionDates table and the reef table (we can keep the spotterId field)
- Should we check for datetimes now and not dates in all data? (live data, historical data etc)

ericboucher:
- no it will still have multiple ones but the goal of this issue is to create the "initial" exclusion date and avoid getting data from installation


- I don't think we need a relationship between these two tables, we can just get the spotterId currently associated with the reef

- yes, as mentioned in my comment above we should check for datetime instead of dates

ericboucher:
Implemented as a popup

PR #423, opened 2021-01-18
Add deploy and exclusion date functionality

avalmas-programize:
This PR is based on https://github.com/aqualinkorg/aqualink-app/pull/418 so it should be merged after it.

### Added three new endpoints
- GET /:reef_id/exclusion_dates Get exclusion dates of the reef's spotter
- POST /:reef_id/deplοy Deploy the reef's spotter
- POST /:reef_id/maintain Add maintainance period

### Change exclusion dates to exclusion datetimes
- Add migration to drop the null constraint on `start_date` making it optional
- Fix implementation wherever the transition from date to datetime causes issues
- Take into consideration that `start_date` might be NULL

Related issue: https://github.com/aqualinkorg/aqualink-app/issues/387


ericboucher:
@avalmas-programize I think it looks good overall but it might be cleaner to simply get all the exclusion dates for a reef and create javascript utility functions for filtering or returning isExcluded(date)

avalmas-programize:
I have added a new exclusion date util function to reduce the code repetition.
Letting the database do the filtering is much more efficient than js code.

tleraj:
The designs implemented look like this:

### Deploy
 | button | dialog |
 |---|---|
 | Screenshot_20210119_183834 | Screenshot_20210119_183854 |

### Maintenance
 | button | dialog |
 |---|---|
 | Screenshot_20210119_183914 | Screenshot_20210119_183930 |

ericboucher:
> I have added a new exclusion date util function to reduce the code repetition.
> Letting the database do the filtering is much more efficient than js code.

I don't think we should be worried about efficiency here. Clarity and simplicity is more important in our case

packages/api/src/reefs/reefs.service.ts
```diff
@@ -280,6 +274,90 @@ export class ReefsService {
...
+
+    const reef = await this.reefsRepository.findOne({
+      id,
+      spotterId: Not(IsNull()),
+      status: Not(ReefStatus.Deployed),
+    });
+
+    if (!reef) {
+      throw new NotFoundException(`Reef with ID ${id} not found`);
+    }
```
ericboucher, line 288: This is super confusing. Let's only get the reef by ID before throwing `throw new NotFoundException(`Reef with ID ${id} not found`);` if we can't find it.

THEN check of there is actually a spotter and the status is not already deployed and set a clearer message.

packages/api/src/reefs/reefs.service.ts
```diff
@@ -280,6 +274,90 @@ export class ReefsService {
...
+
+    const reef = await this.reefsRepository.findOne({
+      id,
+      spotterId: Not(IsNull()),
+      status: ReefStatus.Deployed,
+    });
+
+    if (!reef) {
+      throw new NotFoundException(`Reef with ID ${id} not found`);
+    }
```
ericboucher, line 313: Same comment as above

packages/api/src/reefs/reefs.service.ts
```diff
@@ -280,6 +274,90 @@ export class ReefsService {
...
+
+    const dateConflict = await getConflictingExclusionDate(
+      this.exclusionDatesRepository,
+      reef.spotterId,
+      startDate,
+      endDate,
+    );
+
+    this.logger.debug(dateConflict);
```
ericboucher, line 328: 👣

packages/api/src/reefs/reefs.service.ts
```diff
@@ -280,6 +274,90 @@ export class ReefsService {
...
+  async getExclusionDates(id: number) {
+    const reef = await this.reefsRepository.findOne({
+      id,
+      spotterId: Not(IsNull()),
+      status: ReefStatus.Deployed,
+    });
+
+    if (!reef) {
+      throw new NotFoundException(`Reef with ID ${id} not found`);
+    }
```
ericboucher, line 352: Same comment as above

packages/api/src/utils/reef.utils.ts
```diff
@@ -101,3 +102,52 @@ export const handleDuplicateReef = (err) => {
   logger.error('An unexpected error occurred', err);
   throw new InternalServerErrorException('An unexpected error occurred');
 };
+
```
ericboucher, line 105: Let's transition these to javascript and drop the conflicting exclusion dates as discussed

packages/api/src/reefs/reefs.service.ts
antoniosValmas: Added two separate error message:
1. Reef is already deployed
2. Reef has no spotter

packages/api/src/reefs/reefs.service.ts
antoniosValmas: Same

packages/api/src/reefs/reefs.service.ts
antoniosValmas: Removed

packages/api/src/reefs/reefs.service.ts
antoniosValmas: Same

packages/api/src/utils/reef.utils.ts
antoniosValmas: Done

packages/api/migration/1610959829016-MakeStartExclusionDateOptional.ts
```diff
@@ -0,0 +1,30 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "exclusion_dates" DROP CONSTRAINT "no_duplicates_start_date"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "exclusion_dates" ALTER COLUMN "start_date" DROP NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "exclusion_dates" ADD CONSTRAINT "no_duplicates_start_date" UNIQUE ("spotter_id", "start_date")`,
+    );
```
ericboucher, line 16: @antoniosValmas do we still want this?

packages/api/src/reefs/reefs.controller.ts
```diff
@@ -79,4 +82,30 @@ export class ReefsController {
...
+  }
+
+  @UseGuards(IsReefAdminGuard)
+  @Post(':reef_id/maintain')
+  maintainSpotter(
+    @Param('reef_id', ParseIntPipe) id: number,
+    @Body() maintainSpotterDto: MaintainSpotterDto,
+  ): Promise<void> {
+    return this.reefsService.maintainSpotter(id, maintainSpotterDto);
+  }
```
ericboucher, line 102: Let's rename this to `addExclusionDates`

packages/api/src/reefs/reefs.service.ts
```diff
@@ -269,17 +263,98 @@ export class ReefsService {
...
+
+    const reef = await this.reefsRepository.findOne(id);
+
+    if (!reef) {
+      throw new NotFoundException(`Reef with ID ${id} not found`);
+    }
+
+    if (!reef.spotterId) {
+      throw new BadRequestException(`Reef with ID ${id} has no spotter`);
+    }
```
ericboucher, line 317: Maybe we can have a small util function for these type of errors that we are using a lot to avoid duplication?

packages/api/src/reefs/reefs.service.ts
```diff
@@ -269,17 +263,98 @@ export class ReefsService {
...
+
+    if (reef.status !== ReefStatus.Deployed) {
+      throw new BadRequestException(`Reef with ID ${id} is not deployed`);
+    }
```
ericboucher, line 321: I don't think we should have this check here

packages/api/src/reefs/reefs.service.ts
```diff
@@ -269,17 +263,98 @@ export class ReefsService {
...
```
ericboucher, line 349: I don't think we should have this check here

packages/api/src/workers/dailyData.ts
```diff
@@ -139,16 +144,38 @@ export async function getDailyData(
   const spotterData = spotterRawData
     ? {
         surfaceTemperature: extractSofarValues(
-          spotterRawData.surfaceTemperature,
+          filterSpotterDataByDate(
+            spotterRawData.surfaceTemperature,
```
ericboucher, line 148: Let's use mapping for this function since there is no variable name change

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
```diff
@@ -97,6 +105,24 @@ const ReefNavBar = ({
 
   return (
     <>
+      {user?.token && reef.timezone && (
+        <>
+          <DeployDialog
+            onClose={() => setDeployDialogOpen(false)}
```
ericboucher, line 111: You can also use `isManager` here, no?

Needing a check on `reef.timezone` seems weird to me

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
```diff
@@ -174,15 +200,51 @@ const ReefNavBar = ({
...
+                    {reef.status === "deployed" && (
+                      <Grid item>
+                        <Button
+                          className={classes.button}
+                          onClick={() => setMaintainDialogOpen(true)}
+                          size="small"
+                          color="primary"
+                          variant="outlined"
+                        >
+                          ADD MAINTENANCE
```
ericboucher, line 244: Let's replace MAINTENACE by `EXCLUSION DATES` for now

tleraj: How should the header of this dialog look like? `Exclusion Dates Period`? Or is it good as it is now (`Maintenance Period`)?

ericboucher: Maybe - `Data Exclusion Dates`?

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/MaintainDialog/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+    },
+    dateTitle: {
+      marginBottom: 10,
+    },
+    startDateContainer: {
+      marginBottom: 20,
+    },
+    textField: {
+      color: "black",
+      "&:hover .MuiOutlinedInput-root .MuiOutlinedInput-notchedOutline": {
```
Ry-DS, line 274: I'd have to play around with it a little to better understand, but shouldn't we share this code with DeployDialog? We are already being a little unorthodox by referencing the Mui classes directly, it would be nice to have this hacky code in only one place if its the only way.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/MaintainDialog/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+interface MaintainDialogIncomingProps {
+  open: boolean;
+  token: string;
+  timeZone?: string | null;
+  reefId: number;
+  onClose: () => void;
+}
+
+MaintainDialog.defaultProps = {
```
Ry-DS, line 291: From a holistic view, I feel we could have a shared component between these two dialogs no? Looking at the screenshots I see the header and message is shared, then the fields are what differ. Maybe have the text fields as children?

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
```diff
@@ -202,6 +265,9 @@ const styles = () =>
     managerInfo: {
       marginRight: "0.5rem",
     },
+    button: {
+      width: 180,
```
Ry-DS, line 269: Maybe minWidth is better here so the button can grow by convention? This is just from intuition, feel free to keep this if in reality a set width works better.

Ry-DS (commented):
Overall looks good! I attatched a few smaller fixes to help with readability. Feel free to cherrypick which ones you think are worthwhile.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/MaintainDialog/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+    setEndPickerError("");
+    onClose();
+  };
+
+  const onMaintainAdd = () => {
+    const localStartDate = setTimeZone(maintainStartDateTime, timeZone);
+    const localEndDate = setTimeZone(maintainEndDateTime, timeZone);
+
+    if (!localStartDate) {
+      setStartPickerError("Cannot be empty");
```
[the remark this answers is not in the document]
tleraj: These if statements are use to set different parts of the state (one for `startPickerError` and one for `endPickerError`), so I'm not sure if we could combine them

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
tleraj: The reason I used a fixed width is because I want the `EDIT DETAILS` and `ADD EXCLUSION DATES` buttons to have the same width, because they are vertically aligned.  

Screenshot_20210125_135931

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/MaintainDialog/index.tsx
tleraj: Combined to a common component :wink:

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/MaintainDialog/index.tsx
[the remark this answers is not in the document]
Ry-DS: ahh I see, didn't notice. All good then :+1:

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
Ry-DS: ahh I see, thanks for the clarification :)

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/MaintainDialog/index.tsx
tleraj: The whole purpose of this definition, is to fix the textfield hover border color issue. Without this, when we hover on any textfield, it would look like this:  

Screenshot_20210125_153434

I haven't yet found any way to achieve this only with using jss. Maybe we could add this class on `App.css`, so that it's defined only once.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/DeployDialog/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+const DeployDialog = ({
+  open,
+  token,
+  timeZone,
+  reefId,
+  onClose,
+  classes,
+}: DeployDialogProps) => {
+  const dispatch = useDispatch();
+  const [deployDateTime, setDeployDateTime] = useState<Date | null>(null);
```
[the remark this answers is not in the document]
tleraj: Not quite sure if we could benefit from something like this, because again we would have to define actions for every state variable change, e.g. `setDeployDateTime` should become an action like `() => dispatch({ type: "setDeployDateTime", payload: date })`. Also, we haven't use this approach in any other component, so this might raise a consistency argument.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/MaintainDialog/index.tsx
Ry-DS: Hmm, if this issue is for all text fields it could go into the theme. But if you believe the issue is isolated to just these two use cases we can leave this code as is :+1:

packages/api/migration/1610959829016-MakeStartExclusionDateOptional.ts
antoniosValmas: Not really. Removing

packages/api/src/reefs/reefs.controller.ts
antoniosValmas: Done

packages/api/src/reefs/reefs.service.ts
antoniosValmas: ok

packages/api/src/reefs/reefs.service.ts
antoniosValmas: ok

packages/api/src/workers/dailyData.ts
antoniosValmas: Yeah this seems cleaner.

packages/api/src/utils/reef.utils.ts
```diff
@@ -101,3 +104,40 @@ export const handleDuplicateReef = (err) => {
...
+) => {
+  const allDates = await exclusionDatesRepository
+    .createQueryBuilder('exclusion')
+    .where('exclusion.spotter_id = :spotterId', {
+      spotterId,
+    })
+    .getMany();
+
+  return allDates.filter((exclusionDate) => {
+    return (
```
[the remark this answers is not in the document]
antoniosValmas: Removing

packages/api/src/utils/reef.utils.ts
```diff
@@ -101,3 +104,40 @@ export const handleDuplicateReef = (err) => {
...
+    );
+  });
+};
+
+export const filterSpotterDataByDate = (
+  spotterData: SofarValue[],
+  exclusionDates: ExclusionDates[],
+) => {
+  const data = spotterData.filter(({ timestamp }) => {
+    const excluded = isNil(
```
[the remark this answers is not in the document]
antoniosValmas: ok

packages/api/src/workers/dailyData.ts
```diff
@@ -293,27 +320,22 @@ export async function getReefsDailyData(
...
-            .andWhere('DATE(exclusion.startDate) <= DATE(:endOfDate)', {
-              endOfDate,
-            })
-            .andWhere('DATE(exclusion.endDate) >= DATE(:endOfDate)', {
-              endOfDate,
-            })
-            .getOne(),
-        );
+      const startOfDate = moment(endOfDate);
+      startOfDate.hours(0).minutes(0).seconds(0).milliseconds(0);
```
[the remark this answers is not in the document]
antoniosValmas: Hmm I thought those function did not return the result.
Ok I will chain them then.

packages/api/src/workers/dailyData.ts
```diff
@@ -293,27 +320,22 @@ export async function getReefsDailyData(
...
+        exclusionDatesRepository,
+        reef.spotterId,
+        startOfDate.toDate(),
+        endOfDate,
+      );
 
       const dailyDataInput = await getDailyData(
         reef,
         endOfDate,
         Boolean(includeSpotterData),
```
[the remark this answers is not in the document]
antoniosValmas: Yeah I should move it in the variable declaration. I prefer the `Boolean(...)` syntax though.

packages/api/src/reefs/reefs.service.ts
antoniosValmas: Hmm although this being a quite nice idea, it doesn't work.
I have created a new function:
```js
const shouldHaveReef = (Id, reef) => {
    if (!reef) {
      throw new NotFoundException(`Reef with ID ${id} not found`);
    }
}
```
I have replaced the code in the service functions with the above function and typescript doesn't understands that all this function does, is a type check to the reef variable (to remove the undefined type). So it raises an error saying reef is possible undefined in the subsequent lines in the service method.
So I think It's too much work to make it function and also keep it clean.

ericboucher: Maybe this function could return reef? But it's not a big deal as is, it would be nice to ave more uniform error handling though

packages/api/src/reefs/reefs.controller.ts
```diff
@@ -79,4 +82,30 @@ export class ReefsController {
...
+  @Post(':reef_id/deploy')
+  deploySpotter(
+    @Param('reef_id', ParseIntPipe) id: number,
+    @Body() deploySpotterDto: DeploySpotterDto,
+  ): Promise<void> {
+    return this.reefsService.deploySpotter(id, deploySpotterDto);
+  }
+
+  @UseGuards(IsReefAdminGuard)
+  @Post(':reef_id/maintain')
```
ericboucher, line 96: we could probably rename MaintainSpotterDto and `/maintain` to exclude_dates and excludeSpotterDatesDto

packages/api/src/workers/dailyData.ts
```diff
@@ -293,27 +288,25 @@ export async function getReefsDailyData(
...
-            })
-            .andWhere('DATE(exclusion.startDate) <= DATE(:endOfDate)', {
-              endOfDate,
-            })
-            .andWhere('DATE(exclusion.endDate) >= DATE(:endOfDate)', {
-              endOfDate,
-            })
-            .getOne(),
-        );
+      const startOfDate = moment(endOfDate)
```
ericboucher, line 291: I don't really like introducing this date logic here. It means we have to keep track of our weird date (start/end) logic in one more place

Maybe we could just have an `getExclusionDates` function and pass it down to the dailyData as is.

ericboucher (approved):

PR #427, opened 2021-01-20
Fix popup closing issue

### LINKED ISSUE #426: Reopen reef popup if marker changes due to offset
Steps to replicate:
1. Open the aqualink map
2. search `Dux`
3. Click the first option (takes you to a reef near the left edge of USA)
4. Once the pan animation ends, the popup closes unexpectedly. 

-- find a way to stop the popup from closing pre-maturely.

Ry-DS:
Closes #426
Disables `Popup`'s auto opening `useEffect` if the marker its applied on isn't the original marker (its an offset)

test: http://aqua-popup.surge.sh

### DIFF
--- a/packages/website/src/routes/HomeMap/Map/Markers/index.tsx
+++ b/packages/website/src/routes/HomeMap/Map/Markers/index.tsx
@@ -2,7 +2,7 @@ import { useDispatch, useSelector } from "react-redux";
 import { LayerGroup, Marker, useLeaflet } from "react-leaflet";
 import MarkerClusterGroup from "react-leaflet-markercluster";
 import React, { useCallback, useEffect } from "react";
-import L from "leaflet";
+import L, { DivIcon, Icon } from "leaflet";
 import { reefsToDisplayListSelector } from "../../../../store/Reefs/reefsListSlice";
 import { Reef } from "../../../../store/Reefs/types";
 import {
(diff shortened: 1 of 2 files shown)


tleraj:
I am a little suspicious because using Leaflet's `"moveend"` event was causing problems in the past. I'm not 100% sure that it causes problems now, but seeing it again brings me a little anxiety :stuck_out_tongue:

packages/website/src/routes/HomeMap/Map/Popup/index.tsx
```diff
@@ -37,7 +38,18 @@ const Popup = ({ reef, classes }: PopupProps) => {
...
+      const moveFunc = () => {
+        const point: LatLngTuple = [lat, lng];
+        popup.setLatLng(point).openOn(map);
+        map.off("moveend", moveFunc);
+      };
+      moveFunc();
+      // If the map is zoomed out (we can possibly see repeat markers) then we should reopen the popup.
+      // There's a good chance it might close on moveend (pan animation finishes) - a small quirk of Leaflet
+      // when we have markers outside of the normal map range.
+      // Closest instance I could find in the wild: https://stackoverflow.com/a/30135133/5279269
```
ericboucher, line 51: The other option is to wait until the end of the zoom event before opening the popup. This way you only have to open it once. No?

packages/website/src/routes/HomeMap/Map/Markers/index.tsx
```diff
@@ -44,11 +44,40 @@ const clusterIcon = (cluster: any) => {
   });
 };
 
+function MarkerWithPopup({
```
ericboucher, line 47: this doesn't feel super necessary to have as a standalone function

ericboucher (approved):

Issue #424, opened 2021-01-19
Dynamic map center on first load


ericboucher:
Let's explore having the map center on the site with the highest alert level / DHW instead of a fixed center.

PR #428, opened 2021-01-20
Center map to reef with max dhd

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/424.

### DIFF
--- a/packages/website/src/helpers/reefUtils.ts
+++ b/packages/website/src/helpers/reefUtils.ts
@@ -1,0 +1,34 @@
+import { Reef, Position } from "../store/Reefs/types";
+
+export const findMaxDhwReefPosition = (reefs: Reef[]): Position | null => {
+  const dhws = reefs.map((reef) => reef.latestDailyData.degreeHeatingDays);
+  const maxIndex = dhws.indexOf(Math.max(...dhws));
+  const maxDhwReef = reefs[maxIndex];
+
+  // If the polygon type is a Point, return its coordinates
+  if (maxDhwReef?.polygon.type === "Point") {
+    return [
+      maxDhwReef.polygon.coordinates[1],
+      maxDhwReef.polygon.coordinates[0],
+    ];
+  }
+
+  // If the polygon type is a Polygon, return the coordinates of its centroid
+  if (maxDhwReef?.polygon.type === "Polygon") {
+    const points = maxDhwReef.polygon.coordinates[0].length;
+    const centroidLat =
+      maxDhwReef.polygon.coordinates[0].reduce(
+        (acum, curr) => acum + curr[1],
+        0
+      ) / points;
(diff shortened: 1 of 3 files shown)


tleraj:
Noticed that after searching for a location that is not a reef (e.g. `France`) and then selecting an item from the search dropwdown, the map did not zoom to the selected reef from the dropdown. Fixed with the above commit.

ericboucher:
Position and LatLng have the OPPOSITE order, we have to be careful about this so I explicitly returned a LatLng object.

For the centroid and future more complicated functions, we could probably start using Turf.js

packages/website/src/helpers/reefUtils.ts
```diff
@@ -0,0 +1,34 @@
+import { LatLng } from "leaflet";
+import { Reef } from "../store/Reefs/types";
+
+export const findMaxDhwReefPosition = (reefs: Reef[]): LatLng | null => {
+  const dhws = reefs.map((reef) => reef.latestDailyData.degreeHeatingDays);
+  const maxDhwReef = reefs[dhws.indexOf(Math.max(...dhws))];
```
Ry-DS, line 6: Reduce might look cleaner? You could also use lodash's `maxBy` here.

packages/website/src/helpers/reefUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+      maxDhwReef.polygon.coordinates[0]
+    );
+  }
+
+  // If the polygon type is a Polygon, return the coordinates of its centroid
+  if (maxDhwReef?.polygon.type === "Polygon") {
+    const points = maxDhwReef.polygon.coordinates[0].length;
+    const centroidLat =
+      maxDhwReef.polygon.coordinates[0].reduce(
+        (acum, curr) => acum + curr[1],
```
Ry-DS, line 21: You could use lodash's meanBy for both of these

Ry-DS (commented):
Looks good, but we could do most of these operations with lodash to keep things more readable.

tleraj:
> > Noticed that after searching for a location that is not a reef (e.g. `France`) and then selecting an item from the search dropdown, the map did not zoom to the selected reef from the dropdown. Fixed with the above commit.
> 
> This fix does not work @tleraj, if I search for `France`, I now have to press enter twice in order for the map to zoom to the correct spot... Maybe that's not the issue? Things seem to work better on prod - aqualink.org, so maybe it has something to do with the name matching?
> 
> When testing things like that, I highly recommend writing down a list of use-cases and making sure that everything still works as well or better after your code changes.

I reverted this commit so that we can have a closer look on https://github.com/aqualinkorg/aqualink-app/issues/431.

ericboucher (approved):

PR #434, opened 2021-01-23
Fix chart tooltip spacing

ericboucher:
Screen Shot 2021-01-22 at 4 08 14 PM

### DIFF
--- a/packages/website/src/common/Chart/Tooltip/index.tsx
+++ b/packages/website/src/common/Chart/Tooltip/index.tsx
@@ -167,7 +167,7 @@ const styles = () =>
     },
     tooltipHeader: {
       flex: "0 1 auto",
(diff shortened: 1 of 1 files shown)


Ry-DS (approved):
Looks good! As long as it works on your end :+1:

packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -167,7 +167,7 @@ const styles = () =>
     },
     tooltipHeader: {
       flex: "0 1 auto",
-      padding: "0.5rem 1rem 0 1rem",
+      padding: "0.5rem 1rem 1rem 1rem",
```
Ry-DS, line 170: I think you can do `padding: 0.5rem 1rem 1rem` (top - right and left - bottom)

Issue #430, opened 2021-01-21
Bug with lat/lon input in site detail edit mode


ericboucher:
In Safari, the lat/lon user inputs of the site details page is acting up.

- using the delete button on `171.6` results in `171` directly, removing the `.` which is quite confusing
- having an empty `latitude` or `longitude` during the edit (eg. if you want to spell it out from the beginning) crashes the page -> `Something went wrong`.

On Safari specifically:
- adding a number in the middle does not work or happens it at the end
- saving sometimes does not work and the previous value of the pin seems to be used

<img width="918" alt="Screen Shot 2021-01-20 at 4 50 01 PM" src="https://user-images.githubusercontent.com/16843267/105258944-abe65b80-5b3f-11eb-8f98-698df4f69b11.png">

PR #438, opened 2021-01-26
Improve reef location editing to avoid crashes and cursor jumps

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/430.

 - [x] fix empty values for `latitude` and `longitude`
 - [x] fix delete button on `171.6` delete button resulting in `171` and not `171.`
 - [x] fix cursor moving to the end when entering values to the middle on safari
 - [x] update reef.name maxLength to be `100` instead of `50` and add a validation in the frontend


tleraj:
Noticed an issue on register site page:  when a user submitted the reef data, there was no redirect on the reef's page. That was happening because the register component was wrapped inside a `Router`, which is not necessary, because the whole app is wrapped inside a `Router` in `src/layout/index.tsx`. Fixed with the previous commit. Also added a loading state variable, so that the `Submit` button gets disabled and displaying `Saving...`.

tleraj:
So the issue was that we were using `react-hook-form` and we had an `onChange` prop on some of our textfields (longitude and latitude), which in general is not highly recommended (an in our case causing a bug on Safari). I removed the `react-hook-form`, because we need the catch the change event on those two textfields in order to change the marker position. So now it's a simple textfield with `value` and `onChange` implementation.

packages/website/src/layout/App/App.css
```diff
@@ -112,3 +112,17 @@ body {
...
+input::-webkit-outer-spin-button,
+input::-webkit-inner-spin-button {
+  -webkit-appearance: none;
+  margin: 0;
+}
+
+/* Firefox */
+input[type=number] {
+  -moz-appearance: textfield;
+}
```
ericboucher, line 128: newline?

Are these changes still necessary?

tleraj: Newline added. Yes we need it to hide the increase/decrease arrows for the type number textfields https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp.

ericboucher:
> So the issue was that we were using `react-hook-form` and we had an `onChange` prop on some of our textfields (longitude and latitude), which in general is not highly recommended (an in our case causing a bug on Safari). I removed the `react-hook-form`, because we need the catch the change event on those two textfields in order to change the marker position. So now it's a simple textfield with `value` and `onChange` implementation.

@tleraj can you check in the code if we're not doing that in other places just in case?

tleraj:
> > So the issue was that we were using `react-hook-form` and we had an `onChange` prop on some of our textfields (longitude and latitude), which in general is not highly recommended (an in our case causing a bug on Safari). I removed the `react-hook-form`, because we need the catch the change event on those two textfields in order to change the marker position. So now it's a simple textfield with `value` and `onChange` implementation.
> 
> @tleraj can you check in the code if we're not doing that in other places just in case?

@ericboucher So I checked if we do this in other places as well. But they are all necessary as the involve date/time pickers, or dropdown item selection (e.g. weather conditions on survey form). I tested all of them just in case, and they work fine.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -20,64 +20,137 @@ import {
...
 
-      if (
-        draftReef?.coordinates &&
-        (field === "latitude" || field === "longitude")
-      ) {
+  const onFieldChange = (
+    event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
+  ) => {
+    const { name: field, value: newValue } = event.target;
+    switch (field) {
```
Ry-DS, line 88: You can cut down on some duplication here if these fields were in an object instead, then you can use the field name to update the object key, and group numeric fields together, like:
```ts
switch(field) {
   case "numeric1":
   case "numeric2":
        updateState({...state, [field]: /*handle numeric here*/}}
    default:
        /* Update state normally for all other fields */
       updateState({...state, [field]: value}}
```
This is possible with `useState` or `useReducer` which would store state as an object natively. Let me know what you think of this approach :+1:

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -20,64 +20,137 @@ import {
...
+        break;
+      default:
+        break;
     }
-  }, [draftLatitude, draftLongitude, setValue]);
+  };
 
   return (
-    <form onSubmit={handleSubmit(formSubmit)}>
+    <form onSubmit={(e) => formSubmit(e)}>
```
Ry-DS, line 153: Might be possible to just pass the function normally?
`onSubmit={formSubmit}`

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -130,68 +196,67 @@ const EditForm = ({ reef, onClose, onSubmit, classes }: EditFormProps) => {
...
               Cancel
             </Button>
           </Grid>
           <Grid item>
             <Button
-              className={classes.button}
               type="submit"
+              disabled={
+                loading ||
+                !!reefName.error ||
```
Ry-DS, line 250: This would also be easier if all this state was in an object (loop through all entries to check if nothing has errors)

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -130,68 +196,67 @@ const EditForm = ({ reef, onClose, onSubmit, classes }: EditFormProps) => {
...
-              inputRef={register({
-                required: "Required",
-                pattern: {
-                  value: /^[+-]?([0-9]*[.])?[0-9]+$/,
-                  message: "Invalid input",
-                },
-              })}
-              error={!!errors.longitude}
-              helperText={errors?.longitude?.message || ""}
+              error={Boolean(reefLongitude.error)}
```
Ry-DS, line 229: If all these fields are similar enough, maybe you could make a new component for text fields? Where there's a boolean stating if the textfield is numeric or not.

Ry-DS (commented):
Few suggestions to improve quality :+1:

tleraj:
So I created a custom `useFormField` hook in order to handle the form field changes and created a new component for textfields. I added some comments to clarify what exactly this hook does. @Ry-DS Let me know what you think :slightly_smiling_face:

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
tleraj: In general I prefer my state to be stored in multiple variables rather than an object. If this raises a performance argument, then we can change this.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
tleraj: This now became

```typescript
!!find(
  [reefName, reefDepth, reefLongitude, reefLatitude],
  (field) => field.error
)
```
which I think is much cleaner. Let me know what you think :slightly_smiling_face:

tleraj:
So I created a custom `useFormField` hook in order to handle the form field changes and created a new component for textfields. I added some comments to clarify what exactly this hook does. @Ry-DS Let me know what you think :slightly_smiling_face:

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
Ry-DS: Sounds fair, would be nice to reduce duplication here somehow. Maybe a function? Otherwise, I guess we could leave as is :+1:

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -1,121 +1,143 @@
...
 import { Reef, ReefUpdateParams } from "../../../../../store/Reefs/types";
 import { getReefNameAndRegion } from "../../../../../store/Reefs/helpers";
 import {
   reefDraftSelector,
   setReefDraft,
 } from "../../../../../store/Reefs/selectedReefSlice";
+import { useFormField } from "./useFormField";
 
-const EditForm = ({ reef, onClose, onSubmit, classes }: EditFormProps) => {
+const STEP = 1 / 10 ** 15;
```
Ry-DS, line 23: A comment here, or a more descriptive name would be nice :)

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -1,121 +1,143 @@
...
+  ]);
+
+  const [reefLatitude, setReefLatitude] = useFormField(
+    location ? location.coordinates[1].toString() : "",
+    ["required", "isNumeric"],
+    draftLatitude?.toString(),
+    setDraftReefCoordinates("latitude")
+  );
+
+  const [reefLongitude, setReefLongitude] = useFormField(
```
Ry-DS, line 70: Shouldn't the useFormField hook automatically default to `""` if it detects an empty input?

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -1,121 +1,143 @@
...
+        setReefName(newValue);
+        break;
+      case "depth":
+        setReefDepth(newValue);
+        break;
+      case "latitude":
+        setReefLatitude(newValue, true);
+        break;
+      case "longitude":
+        setReefLongitude(newValue, true);
```
Ry-DS, line 112: I like what you did here, overriding the set method in `useFormField`!

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
```diff
@@ -1,121 +1,143 @@
...
       <Grid container alignItems="flex-end" spacing={3}>
         <Grid container item sm={12} md={6} spacing={2}>
           <Grid item sm={8} xs={12}>
-            <TextField
-              className={classes.textField}
-              variant="outlined"
-              inputProps={{ className: classes.textField }}
-              fullWidth
-              defaultValue={reefName}
+            <Textfield
```
Ry-DS, line 124: Textfield->TextField?

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/useFormField.ts
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => {
+    if (draftValue) {
+      setField({
+        value: draftValue,
+        error: checks
+          .map((check) => validations[check](draftValue))
+          .filter((error) => error)[0],
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Ry-DS, line 46: Why is this disabled? We should find a way to omit this, but a comment for now explaining why would be nice. Maybe then we can try think of a way to write this effect where we don't need this eslint rule.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/useFormField.ts
```diff
@@ -0,0 +1,60 @@
...
+  }, [draftValue]);
+
+  const handleFieldChange = (value: string, runExtraHandler?: boolean) => {
```
Ry-DS, line 49: maybe default runExtraHandler to false? `runExtraHandler = false` Its falsy either way, but its neater ;)

packages/website/src/routes/RegisterSite/index.tsx
```diff
@@ -167,184 +166,180 @@ const Apply = ({ classes }: ApplyProps) => {
...
+            </Grid>
+          </Container>
         </Box>
-        {newReefId && <Redirect to={`/reefs/${newReefId}`} />}
-        <Snackbar
-          anchorOrigin={{ vertical: "bottom", horizontal: "right" }}
-          open={snackbarOpenFromCarto && snackbarOpenFromDatabase}
-          autoHideDuration={4000}
+      </Box>
+      {newReefId && <Redirect to={`/reefs/${newReefId}`} />}
```
Ry-DS, line 308: This looks quite important, should it be this far down in the tree? Looks like something that should belong up top.

Ry-DS (commented):
Looking good! A few final comments :)

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/index.tsx
Ry-DS: Looks great!

packages/website/src/routes/RegisterSite/index.tsx
tleraj: We need the newly created reef's id. Not sure how we can move this at an upper level.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/useFormField.ts
tleraj: Found a way to overcome this by using `useReducer` instead of `useState`. Now the value of `checks` is being read by the `reducer` and is not needed in the `useEffect`'s dependency array :wink:

packages/website/src/routes/RegisterSite/index.tsx
ericboucher: @tleraj I think he meant in this component but up in the page to make it more visible

tleraj: @ericboucher Done :wink:

ericboucher (approved):

PR #443, opened 2021-01-28
Add a hack to fix anti-aliasing issue with leaflet (thin white grid)

### LINKED ISSUE #439: Upgrade Leaflet version
The goal is to fix tile separation lines from appearing on map tiles

This issue appeared in an old version Leaflet, but was supposed to be fixed. Maybe update Leaflet version? 

Screen Shot 2021-01-27 at 8 52 11 AM

ericboucher:
The hack consists in re-scaling images to prevent the 1px gap between them from appearing.

Fixes #439


packages/website/src/routes/ReefRoutes/Reef/Map/plugins/leaflet-tilelayer-subpixel-fix.ts
```diff
@@ -0,0 +1,21 @@
...
+/*
+ * Workaround for 1px lines appearing in some browsers due to fractional transforms
+ * and resulting anti-aliasing.
+ * https://github.com/Leaflet/Leaflet/issues/3575
+ */
+
+(function () {
+  if (!L || !L.GridLayer || !L.GridLayer.prototype) return;
+
+  const originalInitTile = L.GridLayer.prototype._initTile;
```
ericboucher, line 12: @Ry-DS `_initTile` is not exported by the types we use so the app is complaining on build. Do we have an easy way to augment it (beside a patch?)

PR #444, opened 2021-01-28
[Application] Sunset application folder

ericboucher:
Now that the registration, site creation, and spotter application pages have been added into the Aqualink web app, we are sunsetting the standalone application pages.


packages/website/src/store/User/userSlice.ts
```diff
@@ -5,7 +5,7 @@ import {
   ActionReducerMapBuilder,
   AsyncThunk,
 } from "@reduxjs/toolkit";
-import { FirebaseError } from "firebase";
+import { FirebaseError } from "firebase/app";
```
Ry-DS, line 8: Looks out of place, but all good if its for cleanup :+1:

Ry-DS (approved):
Looks good!

Issue #432, opened 2021-01-21
Activate search bar on site and survey pages


ericboucher:
To ease up navigation in the app, activate the search bar on site and survey pages.

The only difference is that the search should only use sites and not zoom to `France` or other random input.

When the user selects a site in the dropdown, we should open the selected site page

PR #446, opened 2021-02-01
Add search bar to site and survey details pages

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/432. When the user is on site or survey details page and searches for a reef in the search bar, he gets redirected on the searched site's detail page.
Designs:

| desktop | mobile |
|---|---|
| Screenshot_20210201_152216 | Screenshot_20210201_152202 |

### DIFF
--- a/packages/website/src/common/NavBar/index.tsx
+++ b/packages/website/src/common/NavBar/index.tsx
@@ -98,7 +98,7 @@ const NavBar = ({
             {searchLocation && (
               <Hidden xsDown>
                 <Grid item sm={3}>
-                  <Search geolocationEnabled={Boolean(geolocationEnabled)} />
+                  <Search geolocationEnabled={geolocationEnabled} />
                 </Grid>
               </Hidden>
             )}
@@ -178,7 +178,7 @@ const NavBar = ({
             {searchLocation && (
               <Hidden smUp>
                 <Grid item xs={12} style={{ margin: 0, paddingTop: 0 }}>
-                  <Search geolocationEnabled={Boolean(geolocationEnabled)} />
+                  <Search geolocationEnabled={geolocationEnabled} />
                 </Grid>
               </Hidden>
             )}
@@ -248,7 +248,7 @@ interface NavBarIncomingProps {
 }
 
 NavBar.defaultProps = {
-  geolocationEnabled: true,
+  geolocationEnabled: false,
   routeButtons: false,
 };
 
--- a/packages/website/src/common/NavBar/index.tsx
+++ b/packages/website/src/common/NavBar/index.tsx
@@ -29,7 +29,12 @@ import RouteButtons from "../RouteButtons";
 import MenuDrawer from "../MenuDrawer";
 import { userInfoSelector, signOutUser } from "../../store/User/userSlice";
 
-const NavBar = ({ searchLocation, routeButtons, classes }: NavBarProps) => {
+const NavBar = ({
+  searchLocation,
+  geolocationEnabled,
+  routeButtons,
+  classes,
+}: NavBarProps) => {
   const user = useSelector(userInfoSelector);
   const dispatch = useDispatch();
   const [registerDialogOpen, setRegisterDialogOpen] = useState<boolean>(false);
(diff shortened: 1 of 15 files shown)


packages/website/src/routes/Surveys/index.tsx
```diff
@@ -21,15 +21,27 @@ import NavBar from "../../common/NavBar";
...
+  const reefsList = useSelector(reefsListSelector);
   const loading = useSelector(reefLoadingSelector);
   const error = useSelector(reefErrorSelector);
   const dispatch = useDispatch();
   const reefId = match.params.id;
   const surveyId = match.params.sid;
 
+  // Fetch reefs for the search bar
+  useEffect(() => {
+    if (reefsList.length === 0) {
```
ericboucher, line 40: What would happen the API returns an empty list for some reason? Or an error?

packages/website/src/common/NavBar/index.tsx
```diff
@@ -238,10 +243,12 @@ const styles = (theme: Theme) =>
 
 interface NavBarIncomingProps {
   searchLocation: boolean;
+  geolocationEnabled?: boolean;
   routeButtons?: boolean;
 }
 
 NavBar.defaultProps = {
+  geolocationEnabled: true,
```
ericboucher, line 251: The default should probably be `false`

packages/website/src/common/NavBar/index.tsx
```diff
@@ -93,7 +98,7 @@ const NavBar = ({ searchLocation, routeButtons, classes }: NavBarProps) => {
             {searchLocation && (
               <Hidden xsDown>
                 <Grid item sm={3}>
-                  <Search />
+                  <Search geolocationEnabled={Boolean(geolocationEnabled)} />
```
ericboucher, line 101: Since we have a default and are expecting a boolean, I don't think we need this conversion

packages/website/src/routes/Surveys/index.tsx
```diff
@@ -21,15 +21,27 @@ import NavBar from "../../common/NavBar";
...
+      dispatch(reefsRequest());
```
Ry-DS, line 41: Same here, would it be possible to move this useEffect to `Search` (the component that uses this data)?

packages/website/src/common/Search/index.tsx
```diff
@@ -154,6 +167,10 @@ const styles = () =>
     },
   });
 
-type SearchProps = WithStyles<typeof styles>;
+interface SearchIncomingProps {
+  geolocationEnabled: boolean;
```
Ry-DS, line 171: Why not make this optional, like what was done in NavBar? This will simplify NavBar a little and maybe even remove the unneeded cast Eric pointed out.

packages/website/src/common/Search/index.tsx
```diff
@@ -27,8 +28,9 @@ const reefAugmentedName = (reef: Reef) => {
   return name || region || "";
 };
 
-const Search = ({ classes }: SearchProps) => {
+const Search = ({ geolocationEnabled, classes }: SearchProps) => {
   const [searchedReef, setSearchedReef] = useState<Reef | null>(null);
+  const [searchedReefId, setSearchedReefId] = useState<number | null>(null);
```
Ry-DS, line 33: Little confused why this extra state exists - seems to mimic `reefOnMap`? If there's a deeper meaning a comment might help :+1:

Ry-DS (commented):
Some pointers on what I think could be improved :+1:

Ry-DS:
I just noticed this useEffect doesn't have a dependancy array: https://github.com/aqualinkorg/aqualink-app/blob/master/packages/website/src/routes/HomeMap/Map/index.tsx#L106 , maybe you could add one in this PR?

packages/website/src/routes/Surveys/index.tsx
tleraj: The idea behind this is to make a request for the reefs list only when that is necessary, e.g. user refreshes a site/survey details page. The approach I followed in case the API returns `[]` was causing `useEffect` to run endlessly. I fixed this by setting the initial value for `reefsList.list` to `undefined`, and whenever we want this value from the redux store, I do:

```typescript
const reefs = useSelector(reefsListSelector) || [];
```

Now the condition for this `useEffect` can become `if(!reefs)` insead of `if(reefsList.length === 0)`. If the backend fetches `[]` then `reefsList.list` becomes `[]`, without letting this `useEffect` to run endlessly.

packages/website/src/common/Search/index.tsx
tleraj: Nice catch. It really does the exact same logic of `reefOnMap` :wink:

ericboucher (approved):

PR #447, opened 2021-02-01
Introduce Scrollable Handle Drawer

### LINKED ISSUE #436: Implement mobile "drag up" instead of clicking to expand bottom bottom card.
This issue will deprecate issue #435. 

The user should be able to drag the grey bar on the top of the hidden bottom card to see All Reefs. To close the expanded card the user should be able to drag down to get back to the map. (Similar to how Google Maps works on mobile). Designs attached. 

Screen Shot 2021-01-25 at 8 38 30 AM

Ry-DS:
test: http://aqua-drawer.surge.sh
Closes #436
If testing on desktop, ensure you're emulating a mobile!
image
Possible points of discussion: 
- The performance isn't great on dev, but seems manageable once deployed. Might be worthwhile to investigate performance on real phones to ensure its up to standard.
- [x] Explore if any extra work is needed - drawer isn't controlled by state anymore and some systems may rely on being able to control the open state of the drawer


Ry-DS:
TODO: Drawer doesn't close anymore when a row is clicked - find a way to fix this.

.eslintrc
```diff
@@ -92,7 +92,7 @@
       }
     ],
     "no-unused-vars": [
-      "error",
+      "warn",
```
ericboucher, line 95: why this change?

Issue #384, opened 2020-12-09
Improve timezone handling


ericboucher:
We display dates in the sites local timezone.

The dates are stored in the database in UTC. To achieve our goal, we currently convert all dates to string using their respective timezone but the timezone info gets "removed in the process.
Eg. `2020-12-06 12:00 UTC` becomes (in PST) `2020-12-06 04:00` without any timezone info.

Hence, we lose crucial information in the process. For example. running `new Date(transformedDate).toLocaleString(timezone)` does not work and shifts the time again as it wrongly assumes that the given date is in UTC.

Instead, we would like to keep the complete date object and simply shift the timezone it is displayed in.

Eg. `2020-12-06 12:00 UTC` should become (assuming the destination timezone is PST) `2020-12-06 04:00 PST`, as a date object (preferred), or a string.

ericboucher:
Let's keep a normal datestring in UTC as long as we can and instead create a smart util function to localize the dates and times before we display them.

So instead, let's create one/two functions to use at the LAST moment to display dates, eg: displayTimeInTimezone and displayDateInTimezone following this logic:

const dateString = new Date(date).toLocaleString("en", {
    day: "2-digit",
    month: "2-digit",
    year: "2-digit",
    ...(hourlyData ? hourlyOptions : {}),
  });

  const timeZoneName = getTimeZoneName(timeZone || "UTC");

  return `${dateString}${hourlyData && timeZoneName && ` ${timeZoneName}`}

tleraj:
@ericboucher  So the basic idea is that we hold the dates just as they come from the backend (in ISO format), and the whole timezone information gets attached right before we display them (e.g on chart tooltip, survey dates, ...).

tleraj:
Also, I'm pretty sure that the warning in https://github.com/aqualinkorg/aqualink-app/issues/445 is caused by the fact that we use `moment` to display dates at any format we want. If we could make the convention that we would display dates in either of these formars:
 - `MM/DD/YY(YY)`
 - `MM/DD/YY(YY) HH:MM (AM or PM) {timeZoneName}`,

then we could use the function bellow for time display and give the the necessary arguments:

```typescript
export const displayTimeInLocalTimezone = (
  hourlyData: boolean,
  date: DateString,
  timeZone?: string | null,
  options?: Intl.DateTimeFormatOptions
) => {
  if (date) {
    const timeZoneName = getTimeZoneName(timeZone || "UTC");
    const dateString = new Date(date)
[...]

Issue #445, opened 2021-02-01
Investigate browser console warning "provided is not in a recognized RFC2822"


ericboucher:
`
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: 10/15/2020, 10:30:00 AM, _f: undefined, _strict: undefined, _locale: [object Object]
`

To reproduce, go to the home page and load a reef, or go to any reef page, eg. https://aqualink.org/reefs/954

PR #451, opened 2021-02-02
Handle timezones

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/384 and https://github.com/aqualinkorg/aqualink-app/issues/445.


tleraj:
So the thing that caused the deprecation warning was that we were using `moment(convertToLocalTime(...)).format(...)`, but moment accepts only ISO formatted strings and `convertToLocalTime(...)` didn't return such a string. I removed all `convertTo...` utils and changed `displayTimeInLocalTimezone` so that it accepts a desired format which is then passed to `moment`. Now all the dates are in ISO format, but when they are being displayed, they are displayed in the reef's local time thanks to `displayTimeInLocalTimezone`. @ericboucher @Ry-DS Let me know what you think :slightly_smiling_face:

packages/website/src/helpers/dates.ts
```diff
@@ -69,45 +61,6 @@ export const setTimeZone = (date: Date | null, timeZone?: string | null) => {
   return date;
 };
 
-export const convertDailyDataToLocalTime = (
```
ericboucher, line 72: @tleraj We still need a way to convert data before passing it to chartjs no? Otherwise the tooltip data and chartjs axis won't line up. Or do we have other ways to shift the timezone that chartjs is using to display the charts?

tleraj: I found out that `moment(isoDate).tz(timeZone).format()` returns an ISOish date. For example, `2021-01-31T23:59:59.999Z` becomes `2021-01-31T18:59:59-05:00` with an `America/New_York` timezone. So having this in mind, I recreated the `convert...` methods and tried to convert the data before passing them to the charts. But for some reason, chart js does not understand this new format, and displays data in the initial ISO format.

packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -65,7 +66,8 @@ const Tooltip = ({
     isoDate: date,
     format: `MM/DD/YY${hourlyData ? " hh:mm A" : ""}`,
     displayTimezone: hourlyData,
-    timeZone,
+    timeZone: userTimeZone,
```
ericboucher, line 69: why do we need this change in the tooltip? The tooltip was working well, no?

packages/website/src/common/Chart/Tooltip/index.tsx
```diff
@@ -51,29 +51,25 @@ const TemperatureMetric = ({
...
-  const hourlyOptions = {
-    hour: "2-digit",
-    minute: "2-digit",
-  };
-  const dateString = new Date(date).toLocaleString("en", {
-    day: "2-digit",
-    month: "2-digit",
-    year: "2-digit",
-    ...(hourlyData ? hourlyOptions : {}),
+  // const dateString = hourlyData ? date : moment(date).format("MM/DD/YY");
```
ericboucher, line 64: 👣

tleraj:
Made some final changes so that the picker's date corresponds to site's local time.

packages/website/src/common/Chart/Tooltip/index.tsx
tleraj: The chart data correspond to user's local time zone. So the first parameter (`timeZone`) is the time zone we want to process the date with, and the second parameter (`timeZoneToDisplay`) is the time zone we want to be finally displayed (the site's time zone).

Ry-DS (commented):
Suggestions to improve readability :+1:

ericboucher (approved):

ericboucher (changes requested):
What we see now is different than what we had an I don't understand why.

Can we add tests to the different time shift functions maybe?

Currently on prod:
Screen Shot 2021-02-16 at 5 21 56 PM

With this version:
Screen Shot 2021-02-16 at 5 21 40 PM

ericboucher (approved):

PR #453, opened 2021-02-05
Remove chart slice under tooltips

ericboucher:
Screen Shot 2021-02-04 at 4 51 44 PM


tleraj (approved):
Looks good!! Just a small thing to take into consideration.

ericboucher:
Good catch!

PR #456, opened 2021-02-09
Add location coordinates validation

tleraj:
The purpose of this PR is to ensure that:
 - `|latitude| <= 90`
 - `|longitude| <= 180`

### DIFF
--- a/packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/useFormField.ts
+++ b/packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/useFormField.ts
@@ -1,5 +1,4 @@
 import { useEffect, useReducer } from "react";
-import isInt from "validator/lib/isInt";
 
 import validators from "../../../../../helpers/validators";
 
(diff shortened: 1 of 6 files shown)


packages/website/src/routes/RegisterSite/index.tsx
```diff
@@ -78,10 +77,17 @@ const Apply = ({ classes }: ApplyProps) => {
     {
       id: "lat",
       label: "Latitude",
-      validator: (lat: string) => isLatLong(`${lat},${formModel.get("lng")}`),
-      errorMessage: "Enter a valid lat/lng",
+      validator: (lat: string) =>
+        isNumeric(lat) && Math.abs(parseFloat(lat)) <= 90,
+      errorMessage: "Enter a valid latitude",
```
ericboucher, line 82: Maybe we could say, `Enter a valid latitude between -90 and 90.`

Same for longitude?

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/EditForm/useFormField.ts
```diff
@@ -15,6 +15,14 @@ const validations = {
   isInt: (value: string) => (!isInt(value) ? "Must be an integer" : undefined),
   isNumeric: (value: string) =>
     !isNumeric(value) ? "Must be numeric" : undefined,
+  isLat: (value: string) =>
+    !isNumeric(value) || Math.abs(parseFloat(value)) > 90
+      ? "Invalid latitude value"
```
ericboucher, line 20: Let's update the message here as well.


Should we extract the validators like isLat, isLong, isNumeric since we are using them in multiple places?

packages/website/src/routes/RegisterSite/index.tsx
```diff
@@ -77,21 +77,21 @@ const Apply = ({ classes }: ApplyProps) => {
     {
       id: "lat",
       label: "Latitude",
-      validator: validators.isLat,
-      errorMessage: "Enter a valid latitude between -90 and 90",
+      validator: (lat: string) => !validators.isLat(lat),
+      errorMessage: validators.isLat(`${formModel.get("lat")}`),
```
ericboucher, line 81: This is super weird. You should reformat how you get the error message instead of from the validator instead of calling the function again in here

tleraj: I realised that we don't need these validators at all. We just have to set the `errorMessage` based on the form's current field value according to the imported validator. I think this makes more sense.

ericboucher (approved):

PR #457, opened 2021-02-09
Restrict access to reef applications

avalmas-programize:
Delete old fetch many application endpoint
Create new endpoint: GET /reef-applications/reefs/:reef_id
Changed ManyToOne relationship to OneToOne in reef application with reefs

Frontend changes are needed: The frontend should use the new endpoint


ericboucher:
@tleraj can you make the necessary frontend changes on this PR?

packages/api/src/reef-applications/reef-applications.entity.ts
```diff
@@ -45,7 +47,8 @@ export class ReefApplication {
   @UpdateDateColumn()
   updatedAt: Date;
 
-  @ManyToOne(() => Reef, { onDelete: 'CASCADE' })
+  @OneToOne(() => Reef, { onDelete: 'CASCADE' })
```
ericboucher, line 50: Should this trigger a migration?

avalmas-programize: Just added it.

packages/website/src/routes/ReefRoutes/ReefApplication/index.tsx
```diff
@@ -82,12 +82,8 @@ const Apply = ({ match, classes }: ApplyProps) => {
...
-          if (data.length > 0) {
-            setReefApplication(data[0]);
-            setMessage(null);
-          } else {
-            setMessage("No application found");
-          }
+          setReefApplication(data);
+          setMessage(null);
         })
         .catch(() => setMessage("There was an error getting the application"))
```
ericboucher, line 88: Maybe differentiate between 403 and other errors?

packages/website/src/routes/ReefRoutes/ReefApplication/index.tsx
```diff
@@ -82,14 +83,20 @@ const Apply = ({ match, classes }: ApplyProps) => {
...
-          if (data.length > 0) {
-            setReefApplication(data[0]);
-            setMessage(null);
+          setReefApplication(data);
+          setMessage(null);
+        })
+        .catch(({ response }: AxiosError) => {
+          if (response?.data.statusCode === 403) {
+            setMessage(
+              "You are not authorized to access this page. If this is an error, contact"
```
ericboucher, line 92: contact who @tleraj ?

ericboucher (approved):

packages/api/migration/1612891721900-Extra.ts
```diff
@@ -0,0 +1,26 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class Extra1612891721900 implements MigrationInterface {
+  name = 'Extra1612891721900';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "reef_application" DROP CONSTRAINT "FK_77d33d9b9602120cd1529312e77"`,
+    );
```
ericboucher, line 9: I don't think you need to drop and recreate these constraints, just having one action in up and one action in dow should be enough, no? @avalmas-programize

avalmas-programize: Usually when altering a column that is also a foreign key it doesn't allow you to until you drop the foreign key constraint.
Apparently that is not true on adding new indexes on the column.
Fixed. (Also changed the name of the file, it was too simple)

Issue #452, opened 2021-02-04
Selected Site Featured Image should forward to the corresponding survey page


ericboucher:
Clicking on the image should redirect to the survey the image is coming from

<img width="954" alt="Screen Shot 2021-02-03 at 10 39 56 AM" src="https://user-images.githubusercontent.com/16843267/106825503-df45e180-6639-11eb-884a-1490b55cafd5.png">

PR #458, opened 2021-02-09
Link featured images to surveys

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/452.

### DIFF
--- a/packages/website/src/common/Search/index.tsx
+++ b/packages/website/src/common/Search/index.tsx
@@ -12,7 +12,6 @@ import Autocomplete from "@material-ui/lab/Autocomplete";
 import { Redirect } from "react-router-dom";
 
 import {
-  reefOnMapSelector,
   setReefOnMap,
   setSearchResult,
 } from "../../store/Homepage/homepageSlice";
@@ -35,9 +34,10 @@ const reefAugmentedName = (reef: Reef) => {
 const Search = ({ geolocationEnabled, classes }: SearchProps) => {
   const [searchedReef, setSearchedReef] = useState<Reef | null>(null);
   const [searchValue, setSearchValue] = useState("");
+  // Variable that listens to reef id changes in order to redirect in case geolocation is disabled
+  const [newReefId, setNewReefId] = useState<number>();
   const dispatch = useDispatch();
   const reefs = useSelector(reefsListSelector);
-  const reefOnMap = useSelector(reefOnMapSelector);
   // eslint-disable-next-line fp/no-mutating-methods
   const filteredReefs = (reefs || [])
     .filter((reef) => reefAugmentedName(reef))
@@ -74,12 +74,14 @@ const Search = ({ geolocationEnabled, classes }: SearchProps) => {
     if (value) {
       setSearchedReef(null);
       dispatch(setReefOnMap(value));
+      setNewReefId(value.id);
     }
   };
 
   const onSearchSubmit = () => {
     if (searchedReef) {
       dispatch(setReefOnMap(searchedReef));
+      setNewReefId(searchedReef.id);
       setSearchedReef(null);
     } else if (searchValue && geolocationEnabled) {
       mapServices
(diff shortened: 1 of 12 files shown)


packages/website/src/common/Search/index.tsx
```diff
@@ -35,9 +34,10 @@ const reefAugmentedName = (reef: Reef) => {
 const Search = ({ geolocationEnabled, classes }: SearchProps) => {
   const [searchedReef, setSearchedReef] = useState<Reef | null>(null);
   const [searchValue, setSearchValue] = useState("");
+  // Variable that listens to reef id changes in order to redirect in case geolocation is disabled
+  const [newReefId, setNewReefId] = useState<number>();
```
ericboucher, line 38: I'm not sure I understand why we need this new id holder

packages/website/src/common/Search/index.tsx
```diff
@@ -97,8 +102,9 @@ const Search = ({ geolocationEnabled, classes }: SearchProps) => {
 
   return (
     <>
+      {/* Redirect to searched reef's details page for components that do not use geolocation */}
       {!geolocationEnabled && reefOnMap?.id && (
```
Ry-DS, line 106: when we say `geolocation` are we using the user's location? A better name for this to avoid confusion could be `geocodingEnabled`, where geocoding is the process of turning a location string into lat and long.

packages/website/src/routes/HomeMap/ReefTable/SelectedReefCard/index.tsx
```diff
@@ -160,10 +165,15 @@ const SelectedReefContent = ({ reef, url }: SelectedReefContentProps) => {
       justify="space-between"
       spacing={1}
     >
-      {url && (
+      {url && surveyId && (
```
Ry-DS, line 168: Changing url to `imageUrl` or `imgSrc` would be nice :+1: just `url` implies that it could be a hyperlink imo

packages/website/src/common/Search/index.tsx
```diff
@@ -97,8 +102,9 @@ const Search = ({ geolocationEnabled, classes }: SearchProps) => {
...
-        <Redirect to={`/reefs/${reefOnMap?.id}`} />
+        <Redirect to={`/reefs/${reefOnMap.id}`} />
```
Ry-DS, line 107: I don't think doing a hard `Redirect` like this is a good idea - we could be changing `setReefOnMap` anywhere in the codebase and this would give quite unpredictable behavior.

Instead, maybe we could list when exactly we need to redirect, and perform the redirect there with `history.push`?
For example, the only two places I see are:
- onDropdownSelect
- onSearchSubmit
So, lets just put `history.push` there instead? Then, we narrow the scope of this to only run in those cases listed instead of as it is now (whenever reefOnMap changes). This should allow the surveyId link you added plus any future redirects to not be affected down the line.

If I'm missing something, or you think we really do need `<Redirect/>` let me know :+1:

tleraj: In general I have been avoiding the use of `history.push` as it triggers an `eslint(fp/no-mutating-methods)` warning, and used `Redirect` instead. If we are OK with that, then I can use `history.push` instead :slightly_smiling_face:

packages/website/src/common/Search/index.tsx
```diff
@@ -74,14 +73,20 @@ const Search = ({ geolocationEnabled, classes }: SearchProps) => {
...
       dispatch(setReefOnMap(value));
+      if (!geocodingEnabled) {
+        browserHistory.push(`/reefs/${value.id}`);
+      }
     }
   };
 
   const onSearchSubmit = () => {
     if (searchedReef) {
+      if (!geocodingEnabled) {
```
ericboucher, line 84: Taking a step back, maybe we should change this flag to be something like `searchOnMapEnabled` or the opposite: `directLink`

And rename the `reefOnMap` to something like `selectedReef`, what do you think?

tleraj:
So to sum up:
 - featured/selected site image on homepage --> site details
 - featured image on site details --> survey details
 - timeline images on site details --> survey details

ericboucher (approved):

packages/website/src/common/Search/index.tsx
Ry-DS: Definitely worthwhile compared to redirect here, since it gives the behaviour we need :)

Issue #459, opened 2021-02-09
Visual bug for site tooltips on large screens


ericboucher:
On larger screens, the map tooltip does not scale properly and the base tooltip is larger than the content box:

<img width="2560" alt="Screen Shot 2021-02-09 at 3 35 01 PM" src="https://user-images.githubusercontent.com/16843267/107442967-c2ebee00-6aec-11eb-9604-d102e4db69f1.png">

PR #465, opened 2021-02-17
Scale map tooltip on large screens

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/459.
Designs;

| 3840x2160 |
|---|
| screencapture-localhost-3000-map-2021-02-17-18_46_25 |

### DIFF
--- a/packages/website/src/routes/HomeMap/Map/Popup/index.tsx
+++ b/packages/website/src/routes/HomeMap/Map/Popup/index.tsx
@@ -176,6 +176,9 @@ const styles = (theme: Theme) =>
     popup: {
       width: "12vw",
       minWidth: 200,
+      [theme.breakpoints.up("xl")]: {
+        width: "7vw",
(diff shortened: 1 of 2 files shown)


packages/website/src/routes/HomeMap/Map/Popup/index.tsx
```diff
@@ -176,6 +176,9 @@ const styles = (theme: Theme) =>
     popup: {
       width: "12vw",
       minWidth: 200,
+      [theme.breakpoints.up("xl")]: {
+        width: "7vw",
```
ericboucher, line 180: is this useful you think?

tleraj: Without it it the tooltip looks a little stretched horizontally

ericboucher (approved):

PR #468, opened 2021-02-19
Import hobo data

avalmas-programize:
A new endpoint has been created (to take advantage of the dependency injection of nest).
This is not the final solution as the functionality will not be public at first.

Currently:
The script parses all xlsx files successfully
Creates the new reef, the new reef points of interest and the time series data.
The images are being parsed and their metadata are being read.


avalmas-programize:
Images have been parsed
Surveys and Survey media are created
A script has been added in the script folder (upload-hobo-data.ts) and a new command in the package.json has been added.
The script takes in average 100-110 seconds to run.

packages/api/src/google-cloud/google-cloud.service.ts
```diff
@@ -31,6 +32,37 @@ export class GoogleCloudService {
...
+        'GCS_BUCKET or STORAGE_FOLDER variable has not been initialized',
+      );
+      throw new InternalServerErrorException();
+    }
+    const ext = path.extname(filePath);
+    const basename = path.basename(filePath).replace(ext, '');
+    const randomString = Array(16)
+      .fill(null)
+      .map(() => Math.round(Math.random() * 15).toString(16))
+      .join('');
```
ericboucher, line 47: Let's avoid duplicating code, we already have most of these functions lying around, eg. `assignName`

packages/api/src/time-series/time-series.controller.ts
```diff
@@ -1,4 +1,25 @@
...
 @Controller('time-series')
-export class TimeSeriesController {}
+export class TimeSeriesController {
+  constructor(private timeSeriesService: TimeSeriesService) {}
+
+  @Post('upload')
+  @UseInterceptors(FileInterceptor('file'))
+  uploadHoboData(
+    @Body('aliases') aliases: string,
+    @Body('email') email: string,
```
ericboucher, line 19: why do we need the email or alias in here?

packages/api/src/time-series/time-series.service.ts
```diff
@@ -1,4 +1,331 @@
...
+      };
+
+      return {
+        name: aliasesMap[reefId],
+        polygon: point,
+      };
+    });
+
+    this.logger.log('Saving reefs');
+    const reefEntities = await this.reefRepository.save(reefs);
```
ericboucher, line 154: what happens in case of duplicates here?

packages/api/src/time-series/time-series.service.ts
```diff
@@ -1,4 +1,331 @@
...
+        }));
+      })
+      .flat();
+
+    const bottomTemperatureData = parsedData.map((data) => ({
+      timestamp: data.timestamp,
+      value: data.bottomTemperature,
+      reef: data.reef,
+      poi: data.poi,
+      metric: metricToMetrics[Metric.BOTTOM_TEMPERATURE],
```
ericboucher, line 224: add new source "HOBO"

packages/api/src/time-series/time-series.service.ts
```diff
@@ -1,4 +1,331 @@
...
+        });
+      })
+      .flat();
+
+    const surveys = imageData.map((image) => ({
+      reef: image.reef,
+      userId: user,
+      diveDate: image.createdDate,
+      // Add default values to other required fields
+      weatherConditions: WeatherConditions.Calm,
```
ericboucher, line 273: Let's add a "no data" option as discussed, I think it will be more accurate/scientific than defaulting to "calm"

packages/api/src/time-series/time-series.service.ts
```diff
@@ -1,4 +1,331 @@
...
+      .map((image) => ({
+        url: image.url,
+        featured: true,
+        hidden: false,
+        type: MediaType.Image,
+        poiId: image.poi,
+        survey: image.survey,
+        metadata: JSON.stringify({}),
+        // Add default values to other required fields
+        observations: Observations.Healthy,
```
ericboucher, line 323: DITTO, let's add a "no data" option

packages/api/package.json
```diff
@@ -52,11 +53,13 @@
     "@nestjs/schedule": "^0.4.1",
     "@nestjs/typeorm": "^7.0.0",
     "@types/bluebird": "^3.5.32",
+    "@types/unzipper": "^0.10.3",
     "axios": "^0.19.2",
     "axios-retry": "^3.1.8",
     "bluebird": "^3.7.2",
     "class-transformer": "^0.3.1",
     "class-validator": "^0.12.2",
+    "csv-parse": "^4.15.1",
```
ericboucher, line 62: let's make sure to put all we can in `devDependencies`

antoniosValmas: Moved a `@types` dependency to dev

packages/api/src/time-series/time-series.service.ts
avalmas-programize: Done

packages/api/src/time-series/time-series.service.ts
avalmas-programize: Done

packages/api/src/time-series/time-series.service.ts
avalmas-programize: Done

packages/api/src/google-cloud/google-cloud.service.ts
avalmas-programize: I have created a new helper function (getRandomName)

packages/api/src/time-series/time-series.service.ts
antoniosValmas: It gets the existing reef and continues

packages/api/src/time-series/time-series.controller.ts
antoniosValmas: Do you have any objection on the parameters of the script?
Email: The user email
Aliases: A stringified array of tuples of size 2, on the first index the index of the reef is expected and in the second the name of the reef

ericboucher: I was mostly surprised that this was in the time-series controller and services. Maybe we could keep the data upload functions separate?

Also, I am not sure we need Aliases, especially if we have reef deduplication. The name is a just "Palau Patch Reef X"

ericboucher:
Let's make sure that the data is augmented upon reef creation (adding maxMonthlyMean and max monthly values, region, ...)

packages/api/src/time-series/time-series.controller.ts
antoniosValmas: Ok I will create a separate file
Is this going to be the case for all reef imported? I just didn't want to have hard-coded string that were going to change in the future.

packages/api/src/reef-pois/reef-pois.service.ts
```diff
@@ -5,22 +5,43 @@ import {
...
+      throw new NotFoundException(`Reef with id ${reef} was not found`);
+    }
+
+    const polygon: GeoJSON =
+      longitude && latitude
+        ? {
+            type: 'Point',
+            coordinates: [longitude, latitude],
+          }
+        : reefEntity.polygon;
```
ericboucher, line 42: @tleraj does that work for you or is it easier if we keep it `null` until it is set by a script or by the user as something different than the reef.polygon?

ericboucher:
@antoniosValmas is this PR ready for a full review?

avalmas-programize:
> @antoniosValmas is this PR ready for a full review?

Yes this PR is ready for full review

packages/api/src/reef-pois/reef-pois.service.ts
tleraj: It would be easier to check if coordinates are just `null` rather than `[reef.lng, reef.lat] === [point.lng, point.lat]`.

packages/api/src/reef-pois/reef-pois.service.ts
```diff
@@ -5,22 +5,43 @@ import {
...
+    if (!reefEntity) {
+      throw new NotFoundException(`Reef with id ${reef} was not found`);
+    }
+
+    const polygon: GeoJSON | undefined =
+      longitude !== undefined && latitude !== undefined
+        ? {
+            type: 'Point',
+            coordinates: [longitude, latitude],
+          }
```
ericboucher, line 41: It might be worth exporting this logic into a small util since we use it in a few places

packages/api/scripts/upload-hobo-data.ts
```diff
@@ -0,0 +1,81 @@
...
+import { User } from '../src/users/users.entity';
+import { Survey } from '../src/surveys/surveys.entity';
+import { SurveyMedia } from '../src/surveys/survey-media.entity';
+import { GoogleCloudService } from '../src/google-cloud/google-cloud.service';
+import { Sources } from '../src/reefs/sources.entity';
+import { uploadHoboData } from '../src/utils/import-hobo-data';
+import { Region } from '../src/regions/regions.entity';
+import { MonthlyMax } from '../src/reefs/monthly-max.entity';
+
+const { argv } = yargs
```
ericboucher, line 18: Let's add a complete example command here for ease of use

packages/api/src/time-series/dto/poi-data.dto.ts
```diff
@@ -0,0 +1,14 @@
+import { IsNumberString, Validate } from 'class-validator';
+import { ReefPointOfInterest } from '../../reef-pois/reef-pois.entity';
+import { Reef } from '../../reefs/reefs.entity';
+import { EntityExists } from '../../validations/entity-exists.constraint';
+
+export class PoiDataDto {
```
ericboucher, line 6: data-range.dto.ts and reef-data.dto.ts seem to be identical? Is there a way to re-use / merge or is it cleaner this way?

ericboucher:
@antoniosValmas I'll let you resolve the conflicts, I started but I wasn't sure in all cases so probably easier for you :)

packages/api/src/reef-pois/reef-pois.service.ts
avalmas-programize: Done

packages/api/scripts/upload-hobo-data.ts
avalmas-programize: Done

packages/api/src/time-series/dto/poi-data.dto.ts
avalmas-programize: Yeah duplication bothers me as well.
However, this is more future-proof because we are assigning for each endpoint its dedicated dto file.
So if we need to add a new param we will need less changed to be made.
Also we will not need to come up with a clever name for the resulting grouped dto.

packages/api/src/utils/import-hobo-data.ts
```diff
@@ -0,0 +1,649 @@
...
+  reefRepository: Repository<Reef>;
+  poiRepository: Repository<ReefPointOfInterest>;
+  timeSeriesRepository: Repository<TimeSeries>;
+  userRepository: Repository<User>;
+  surveyRepository: Repository<Survey>;
+  surveyMediaRepository: Repository<SurveyMedia>;
+  sourcesRepository: Repository<Sources>;
+  regionRepository: Repository<Region>;
+  monthlyMaxRepository: Repository<MonthlyMax>;
+}
```
ericboucher, line 59: It seems simpler to meet to just pass in the connection from the script file and then initiate the repositories in the `uploadHoboData` function

packages/api/src/utils/import-hobo-data.ts
```diff
@@ -0,0 +1,649 @@
+import { chunk, Dictionary, groupBy, keyBy, minBy } from 'lodash';
```
ericboucher, line 1: Let's change the name of the file to `upload-hobo-data` to be more consistent maybe? Or something similar

avalmas-programize: Yeah `upload-hobo-data` seems better.

packages/api/src/utils/import-hobo-data.ts
avalmas-programize: Well apart from the repositories I also initialize a googleCloudService (which also needs a repository).
So I thought that having all the initialization away from my logic it would feel less crammed.
Also I think this is much more future-proof because this function can easily be called by a service, since you said that this might end-up in the future in our WEB API, by passing the necessary repositories and services acquired by dependency injection.

packages/api/scripts/upload-hobo-data.ts
```diff
@@ -0,0 +1,86 @@
...
+import { Sources } from '../src/reefs/sources.entity';
+import { uploadHoboData } from '../src/utils/upload-hobo-data';
+import { Region } from '../src/regions/regions.entity';
+import { MonthlyMax } from '../src/reefs/monthly-max.entity';
+
+const { argv } = yargs
+  .scriptName('upload-hobo-data')
+  .usage('$0 <cmd> [args]')
+  .example(
+    '$0 -p Aqualink.zip -u example@aqualink.com',
```
ericboucher, line 22: Can -p be a folder?

ericboucher (approved):
Looks good. Let's just have -p be a path to a folder instead.

PR #473, opened 2021-02-22
Fix Mobile Drawer scrolling issues

Ry-DS:
Test: http://aqua-drawer.surge.sh
This PR fixes a bug where the mobile drawer would scroll while closed causing a confusing overlay to show:
image

The mobile drawer now doesn't scroll at all, as any information the user needs would be at the top.
A bouncing pill animation was also added to encourage users to use the drawer after they selected a point on the map. The selected reef name is also shown at the bottom instead of 'all reefs'


tleraj (approved):
Looks good!! :slightly_smiling_face:

ericboucher (approved):

PR #477, opened 2021-03-01
Add table sort dropdown on mobile

### LINKED ISSUE #474: Add ordering options on mobile
@ewolfer67 could we look into adding ordering options in the main reef table on mobile? We dropped the table headers per the design but we might want to add a way to toggle between alphabetical ordering and alert-level ordering

Ry-DS:
Test: http://aqua-sorting.surge.sh
Closes #474
image
image


ericboucher:
Maybe find a way to order the deop down alphabetically @Ry-DS ?

ericboucher (approved):

Issue #481, opened 2021-03-05
Add screenshot to Aqualink README


ericboucher:
Take a nice screenshot of the application and add it to the project README below the description.

PR #482, opened 2021-03-07
Add screenshot to readme (#481)

sachinh19:
Added a screenshot of the landing page of the application below the description in the README file.


ericboucher:
Great start @sachinh19

A few thoughts:
- let's try and have a screenshot that gives more insight into the app. Maybe we could add multiple, but I think a map page and a reef page might be more impressive than the landing page. (https://aqualink.org/map or https://aqualink.org/reefs/2940)
- Keep the image size small so that it's fast to load for people on slower internet connections that come to our readme.

Hope this helps :)

sachinh19:
> Great start @sachinh19
> 
> A few thoughts:
> 
>     * let's try and have a screenshot that gives more insight into the app. Maybe we could add multiple, but I think a map page and a reef page might be more impressive than the landing page. (https://aqualink.org/map or https://aqualink.org/reefs/2940)
> 
>     * Keep the image size small so that it's fast to load for people on slower internet connections that come to our readme.
> 
> 
> Hope this helps :)

Thanks for the feedback. I have addressed the comments and amended the commit.

ericboucher:
Great, now I would suggest:
- Try to have the two images on one line
- For the map screenshot, chose a part of the world that looks a bit fancier :)
- For the reef details page, zoom out a bit to get the metrics card and the chart, no need to keep the blue header

sachinh19:
> Great, now I would suggest:
> 
>     * Try to have the two images appear on one line in the readme
> 
>     * For the map screenshot, chose a part of the world that looks a bit fancier :)
> 
>     * For the reef details page, zoom out a bit to get the metrics card and the chart, no need to keep the blue header

- I tried aligning the images beside each other, they appear smaller and maybe difficult to see all the details on the page for someone who is scrolling the README
- Can you clarify by what do you mean by fancier? Trying to get the complete details so I can try and avoid rework the next time over.
- Will change this in the next update

ericboucher:
The new map screenshot is great. Let's have only one image for now then, I think the map screenshot you propose here works well. Thanks for looking into this!

ericboucher (approved):
Nice! If you can reduce the size of the image a bit that would be best. Otherwise looks good to go. Thanks

PR #483, opened 2021-03-08
Survey point page enhancements

tleraj:
The purpose of this PR is to add the following enhancements:

### Reef detail page
- [x] Survey points, only show those that have a different position on the map
- [x] If HOBO or spotter associated to the reef, use that, otherwise use the closest survey point

### Survey Point Page
- [x] Make it possible to edit the survey point position on the survey point page by entering an edit mode (similar to what we have on the reef details page). Also make the orange pin draggable on the map.

### New Chart
- [x] Add View range picker with a default to 3 months (instead of 1 in the design),  1year, Max.  (see figma designs)
- [x] “Data available from minDate to maxDate”
- [x] Once the spotter data is in our time series database, use the same chart and box for the spotter data on the reef details page
- [x] In the stats box, display HISTORIC instead of surface and all the sensor data when available (hobo, spotter surface, spotter bottom)
- [x] Make it possible to enter new dates manually

### Survey Page
- [x] On survey page, make the survey point name clickable
- [x] In the temperature box, show data from the survey date only. If only hobo data is available, show that instead.

Designs:
### Edit dialog
| closed | open |
|---|---|
| Screenshot_20210316_221700 | Screenshot_20210316_221713 |

### New Chart
Screenshot_20210316_221932


packages/website/src/common/Chart/ChartWithCard/TempAnalysis.tsx
```diff
@@ -0,0 +1,303 @@
...
+        min: minSpotterBottom,
+      },
+      display: hasSpotterBottom,
+    },
+  ];
+
+  const formattedpickerStartDate = moment(pickerStartDate).format("MM/DD/YYYY");
+  const formattedpickerEndDate = moment(pickerEndDate).format("MM/DD/YYYY");
+
+  if (!showCard) {
```
ericboucher, line 126: if you're gonna return null on `!showCard`, let's do it right after defining `showCard`

packages/website/src/common/Chart/ChartWithCard/TempAnalysis.tsx
```diff
@@ -0,0 +1,303 @@
...
+      title: depth ? `BUOY ${depth}m` : "BUOY AT DEPTH",
+      titleClass: `${classes.buoyBottomText} ${!depth && classes.smallTitle}`,
+      rows: {
+        max: maxSpotterBottom,
+        mean: meanSpotterBottom,
+        min: minSpotterBottom,
+      },
+      display: hasSpotterBottom,
+    },
+  ];
```
ericboucher, line 121: All this code is super duplicative.

I think we should almost make it metric agnostic. That's also one situation where having the color as style is ok so that you can have all the definitions in one place.

As a first step you could simplify the cardColumns to be {title, color, rows: calculateMetrics(dataArray)}. For `display` let's just use `!isNil(rows.max)`

packages/website/src/common/Chart/ChartWithCard/TempAnalysis.tsx
```diff
@@ -0,0 +1,303 @@
...
+      lineHeight: "17px",
+    },
+    buoySurfaceText: {
+      color: "#46a5cf",
+      lineHeight: "17px",
+    },
+    hoboText: {
+      color: "#f78c21",
+      lineHeight: "17px",
+    },
```
ericboucher, line 284: If you move the color to the column definition you can simplify to just one lineHeight: 17px

packages/website/src/common/Chart/ChartWithCard/helpers.ts
```diff
@@ -0,0 +1,102 @@
...
+import moment from "moment";
+
+import {
+  MonthlyMaxData,
+  SofarValue,
+  SpotterData,
+} from "../../../store/Reefs/types";
+import { formatNumber } from "../../../helpers/numberUtils";
+
+export const calculateCardMetrics = (
```
ericboucher, line 11: This function is super duplicative again. With the approach I proposed in the other file, I would recommend creating a single standardize function to get the min.mean.max out of an array

packages/website/src/helpers/map.ts
```diff
@@ -0,0 +1,106 @@
...
+  const lngArray = coordArray.map((item) => item[0]);
+  const latArray = coordArray.map((item) => item[1]);
+
+  const lngMean = lngArray.reduce((a, b) => a + b) / len;
+  const latMean = latArray.reduce((a, b) => a + b) / len;
+
+  return [lngMean, latMean];
+};
+
+export const mapBounds = (polygon: Polygon): LatLngBounds => {
```
ericboucher, line 21: I think a lot of these functions already exist within leaflets or famous libraries, no?

tleraj: Not anything too fancy. I removed the `mapBounds` function and used `L.polygon(polygon).getBounds()` method where needed. I found a decent [library](http://makinacorpus.github.io/Leaflet.GeometryUtil/global.html#distance), but unfortunately it does not have type definitions.

ericboucher (approved):

Issue #476, opened 2021-02-25
Add SST Anomaly to reef details page


ericboucher:
### Calculation
We know have monthly max values that we can use to determine daily SST anomaly. Using the max monthly values and daily SST, we can calculate the anomaly as follows (using the SST timestamp as the date):

The daily global 5km SSTA product requires a daily climatology to calculate the daily SST anomalies. Daily climatologies (DC) are derived from the monthly mean (MM) climatology via linear interpolation. To achieve this, we assigned the MM value to the 15th day of each corresponding month, with the individual days between these dates being derived using linear interpolation.
We then calculate the SSTA product using:
SST_anomaly = SST - DC
where the SST is the value for the day in question, and DC is the corresponding daily climatology for that day of the year.

### Frontend - Card design:
![Screen Shot 2021-02-25 at 8 15 06 AM](https://user-images.githubusercontent.com/16843267/109220151-6edc3d00-776d-11eb-8eda-13bbb3c0b896.png)

PR #484, opened 2021-03-09
Add SST Anomaly

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/476.

- [x] Calculate value
- [x] Display value

Designs:

| desktop | iPad Pro | iPad | iPhone 6/7/8 |
|---|---|---|---|
| Screenshot_20210405_185300 | Screenshot_20210405_185715 | Screenshot_20210405_185733 | Screenshot_20210405_185754 |


ericboucher:
@tleraj @avalmas-programize this is a good start but this is overly complicated.

You simply need to figure out: what is the value before and how far was it (in number of days), what is the next value and how far is it? What is the spread between the two monthly max dates (deltaDays)

Then your new value is:

`previousValue * (deltaDays - previousDistance)/deltaDays + nextValue * (deltaDays - nextDistance)/deltaDays`

packages/api/src/utils/liveData.ts
```diff
@@ -146,32 +146,35 @@ export const findSstAnomaly = (
...
+    .clone()
+    .subtract(15, 'days')
+    .set('date', 15)
+    .startOf('day');
+
+  // The date of the next value. It must fall on the next month of the previous
+  // value.
+  const nextDate = previousDate.clone().add(1, 'month').startOf('day');
+
+  const previousValue = orderedMontlyMax[previousDate.get('month')].temperature;
```
ericboucher, line 169: Let's add a quick comment saying that this works because
`// 0 = January & 11 = December`

packages/api/src/utils/liveData.ts
```diff
@@ -146,32 +146,35 @@ export const findSstAnomaly = (
...
+  // the 15th day of this month.
+  const previousDate = now
+    .clone()
+    .subtract(15, 'days')
+    .set('date', 15)
+    .startOf('day');
+
+  // The date of the next value. It must fall on the next month of the previous
+  // value.
+  const nextDate = previousDate.clone().add(1, 'month').startOf('day');
```
ericboucher, line 167: I don't think you need the .startOf('day'); since you are cloning from `previousDate`

packages/api/src/utils/liveData.ts
```diff
@@ -146,32 +146,35 @@ export const findSstAnomaly = (
...
+
+  const previousValue = orderedMontlyMax[previousDate.get('month')].temperature;
+  const previousDistance = now.diff(previousDate, 'days');
```
ericboucher, line 170: Are we sure that this returns a float?
Otherwise we might want to have:

`deltaDays = previousDistance + nextDistance` instead of calculating another diff

ericboucher (commented):
Looking a lot cleaner, thanks!

avalmas-programize (approved):
Seems good to me aswell

ericboucher (approved):

packages/api/src/utils/liveData.ts
```diff
@@ -128,3 +130,53 @@ export const getLiveData = async (
...
+ * Daily climatologies (DC) are derived from the monthly mean (MM) climatology via linear interpolation.
+ * To achieve this, we assigned the MM value to the 15th day of each corresponding month, with the individual
+ * days between these dates being derived using linear interpolation. We then calculate the SSTA product using:
+ *
+ * ST_anomaly = SST - DC
+ *
+ * where the SST is the value for the day in question, and DC is the corresponding daily climatology for that
+ * day of the year.
+ * */
+export const findSstAnomaly = (
```
ericboucher, line 145: Not sure if "find" is the best word. Maybe just `get`?

PR #485, opened 2021-03-09
Timeseries spotter data

avalmas-programize:
- New back-fill script to add spotter data in the time-series table
- New cloud function to schedule daily updates of spotter data

### DIFF
--- a/packages/api/cloud-functions/index.ts
+++ b/packages/api/cloud-functions/index.ts
@@ -2,6 +2,7 @@ import Axios from 'axios';
 import * as functions from 'firebase-functions';
 import { createConnection } from 'typeorm';
 import { runDailyUpdate } from '../src/workers/dailyData';
+import { runSpotterTimeSeriesUpdate } from '../src/workers/spotterTimeSeries';
 
 // We have to manually import all required entities here, unfortunately - the globbing that is used in ormconfig.ts
 // doesn't work with Webpack. This declaration gets processed by a custom loader (`add-entities.js`) to add import
@@ -89,4 +90,28 @@ exports.pingService = functions.pubsub
     // eslint-disable-next-line no-console
     console.log('Pinging server');
     await Axios.get(`${backendBaseUrl}/health-check`);
+  });
+
+exports.scheduledSpotterTimeSeriesUpdate = functions
+  .runWith({ timeoutSeconds: 540 })
+  .pubsub.schedule('0 4 * * * *')
+  .timeZone('America/Los_Angeles')
+  .onRun(async () => {
+    const dbUrl = functions.config().database.url;
+    // eslint-disable-next-line fp/no-mutation
+    process.env.SOFAR_API_TOKEN = functions.config().sofar_api.token;
+    // eslint-disable-next-line no-undef
+    const entities = dbEntities.map(extractEntityDefinition);
+    const conn = await createConnection({
+      ...dbConfig,
+      url: dbUrl,
+      entities,
+    });
+    try {
+      await runSpotterTimeSeriesUpdate(conn);
+      // eslint-disable-next-line no-console
+      console.log(`Daily update on ${new Date()}`);
+    } finally {
+      conn.close();
+    }
   });
--- a/packages/api/cloud-functions/index.ts
+++ b/packages/api/cloud-functions/index.ts
@@ -94,7 +94,8 @@ exports.pingService = functions.pubsub
 
 exports.scheduledSpotterTimeSeriesUpdate = functions
   .runWith({ timeoutSeconds: 540 })
-  .pubsub.schedule('0 4 * * * *')
+  // Run spotter data update every hour
+  .pubsub.schedule('0 * * * * *')
   .timeZone('America/Los_Angeles')
   .onRun(async () => {
     const dbUrl = functions.config().database.url;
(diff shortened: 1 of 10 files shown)


ericboucher:
We will need to add a few metrics

packages/api/cloud-functions/index.ts
```diff
@@ -90,3 +91,27 @@ exports.pingService = functions.pubsub
     console.log('Pinging server');
     await Axios.get(`${backendBaseUrl}/health-check`);
   });
+
+exports.scheduledSpotterTimeSeriesUpdate = functions
+  .runWith({ timeoutSeconds: 540 })
+  .pubsub.schedule('0 4 * * * *')
```
ericboucher, line 97: Add a comment with the schedule in. plain text

ericboucher: Let's run these updates every hour or every two hours

packages/api/cloud-functions/index.ts
```diff
@@ -90,3 +91,27 @@ exports.pingService = functions.pubsub
...
+    const entities = dbEntities.map(extractEntityDefinition);
+    const conn = await createConnection({
+      ...dbConfig,
+      url: dbUrl,
+      entities,
+    });
+    try {
+      await runSpotterTimeSeriesUpdate(conn);
+      // eslint-disable-next-line no-console
+      console.log(`Daily update on ${new Date()}`);
```
ericboucher, line 113: Specify what update is running

packages/api/migration/1615299917998-AppendNewMetrics.ts
```diff
@@ -0,0 +1,76 @@
...
+      `CREATE TYPE "metrics_metric_enum_old" AS ENUM('alert', 'dhw', 'satellite_temperature', 'surface_temperature', 'bottom_temperature', 'sst_anomaly')`,
+    );
+
+    await queryRunner.query(`DROP VIEW "latest_data"`);
+    await queryRunner.query(
+      `ALTER TABLE "time_series" DROP CONSTRAINT "no_duplicate_data"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "time_series" DROP CONSTRAINT "no_duplicate_reef_data"`,
+    );
```
ericboucher, line 42: do we need to drop and recreate these constraints?

packages/api/cloud-functions/index.ts
avalmas-programize: Added comment and changed schedule to run every hour.

packages/api/cloud-functions/index.ts
avalmas-programize: Done

packages/api/migration/1615299917998-AppendNewMetrics.ts
avalmas-programize: These constraints are both include the `metric` field.
So since we need to change the type of the `metric` field we must delete all the constraints (rules) and views associated with it.

However on the `up` script this is not needed because we don't alter the type of field but altering the type contents which seems to be allowed.

packages/api/src/utils/spotter-time-series.ts
```diff
@@ -0,0 +1,148 @@
...
+                  reefToSource,
+                  label[1],
+                  repositories.timeSeriesRepository,
+                ),
+              ),
+            )
+            .flat(),
+          (props, i) => {
+            logger.log(
+              `Saved ${i + 1} out of ${2 * days} of daily spotter data`,
```
ericboucher, line 141: Why 2 *?

packages/api/src/utils/spotter-time-series.ts
```diff
@@ -0,0 +1,148 @@
...
+      ).then((spotterData) => {
+        const dataLabels: [keyof SpotterData, Metric][] = [
+          ['surfaceTemperature', Metric.SURFACE_TEMPERATURE],
+          ['bottomTemperature', Metric.BOTTOM_TEMPERATURE],
+          ['significantWaveHeight', Metric.SIGNIFICANT_WAVE_HEIGHT],
+          ['waveMeanDirection', Metric.WAVE_MEAN_DIRECTION],
+          ['wavePeakPeriod', Metric.WAVE_PEAK_PERIOD],
+          ['windDirection', Metric.WIND_DIRECTION],
+          ['windSpeed', Metric.WIND_SPEED],
+        ];
```
ericboucher, line 123: Maybe we could use a map or objects instead of tuples? This will be clearer in the function below.

Eg. [{spotterMetricName: y, timeseriesMetricName: x}]

packages/api/src/utils/spotter-time-series.ts
avalmas-programize: It was the number of metrics uploaded. I have changed it to the length of the array of metrics.

packages/api/src/utils/spotter-time-series.ts
avalmas-programize: I don't see it being clearer with this.

I need to have all the an data structure that will allow me to map some properties of the spotter data (longitude and latitude should be ignored) to timeseries objects
I don't really need the indexing of an object so I thought a simple tuple will be sufficient

Can you elaborate?

ericboucher (approved):

PR #487, opened 2021-03-11
Rename spotter params to use "top" instead of surface

avalmas-programize:
Renamed `surface_temperature` to `top_temperature`

**Migration issue**: since we might have data on both `metrics` and `time_series` table the type of the `metric` field cannot be easily updated. We can perform the change with the following steps:
- Create two migrations:
  - The first one to add an extra value to enum and 
  - The second one to update the values of the tables and change the enum to its final form. 

_Two migration are needed because enum changes must be committed before utilized._


ericboucher:
@avalmas-programize let's take care of the frontend changes as well

avalmas-programize:
the migration for this should be run with the extra flag `-t each`:
```sh
yarn migration:run -t each
```

tleraj (approved):
Looks good!!

ericboucher:
FYI - I fixed the tests in @tleraj's PR

PR #490, opened 2021-03-18
Daily data fix

avalmas-programize:
Add query params `start` and `end` to be able to select range of data.
If params are omitted last 90 data points are returned

### DIFF
--- a/packages/api/src/reefs/reefs.controller.ts
+++ b/packages/api/src/reefs/reefs.controller.ts
@@ -57,8 +57,12 @@ export class ReefsController {
 
   @Public()
   @Get(':id/daily_data')
-  findDailyData(@Param('id') id: number) {
-    return this.reefsService.findDailyData(id);
+  findDailyData(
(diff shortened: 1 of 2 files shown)


packages/api/src/reefs/reefs.controller.ts
```diff
@@ -57,8 +57,12 @@ export class ReefsController {
 
   @Public()
   @Get(':id/daily_data')
-  findDailyData(@Param('id') id: number) {
-    return this.reefsService.findDailyData(id);
+  findDailyData(
+    @Param('id', ParseIntPipe) id: number,
+    @Query('start') start?: string,
+    @Query('end') end?: string,
```
ericboucher, line 63: Do we have a way to check that these are valid maybe?

avalmas-programize: There is no Pipe to validate a param and also allow it being optional.

So I added an explicit if guard in the service

ericboucher (approved):

Issue #470, opened 2021-02-19
Remove temperatureThreshold from reef entity


ericboucher:
The temperatureThreshold is constructed as maxMonthlyMean + 1, so to avoid confusion, let's remove it from the reef entity and make sure that we only rely on maxMonthlymean in the backend and frontend

PR #491, opened 2021-03-18
Remove temperatureThreshold from reef entity

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/470.


tleraj:
@ericboucher I noticed that on the frontend we never use the `reef.temperatureThreshold` directly. Should I remove it completely from the reef entity, or is it OK as it is right now?

ericboucher:
I think we can remove it completely yes

avalmas-programize (approved):
Seems good to me

ericboucher (approved):

PR #492, opened 2021-03-18
Return time series data by source [backend only]

avalmas-programize:
Group time series data by source.
Returning object:
```js
{
  metric: {
    hobo: [data]
    spotter: [data]
  }
}
```

### DIFF
--- a/packages/api/src/time-series/time-series.service.ts
+++ b/packages/api/src/time-series/time-series.service.ts
@@ -59,7 +59,7 @@ export class TimeSeriesService {
     poiId?: number,
   ): Promise<TimeSeriesData[]> {
     const poiCondition = poiId
-      ? `time_series.poi_id = ${poiId}`
+      ? `(time_series.poi_id = ${poiId} OR time_series.poi_id is NULL)`
       : 'time_series.poi_id is NULL';
 
     return hourly
@@ -96,7 +96,7 @@ export class TimeSeriesService {
 
   private getDataRangeQuery(reefId: number, poiId?: number) {
     const poiCondition = poiId
-      ? `time_series.poi_id = ${poiId}`
+      ? `(time_series.poi_id = ${poiId} OR time_series.poi_id is NULL)`
       : 'time_series.poi_id is NULL';
 
     return this.timeSeriesRepository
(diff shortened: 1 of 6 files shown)


ericboucher:
@tleraj can you make the necessary code changes in the frontend?

packages/api/src/time-series/time-series.service.ts
```diff
@@ -45,36 +58,60 @@ export class TimeSeriesService {
...
+          .andWhere('time_series.reef_id = :reefId', { reefId })
           .andWhere(poiCondition)
           .andWhere('timestamp >= :startDate', { startDate })
           .andWhere('timestamp <= :endDate', { endDate })
           .orderBy('timestamp', 'ASC')
           .getRawMany();
   }
 
+  private getDataRangeQuery(reefId: number, poiId?: number) {
+    const poiCondition = poiId
```
ericboucher, line 98: Let's also keep it like this

`const poiCondition = poiId ? `poi_id = ${poiId}` : 'poi_id is NULL';` no?

packages/api/src/time-series/time-series.service.ts
```diff
@@ -45,36 +58,60 @@ export class TimeSeriesService {
     reefId: number,
     poiId?: number,
   ): Promise<TimeSeriesData[]> {
-    const poiCondition = poiId ? `poi_id = ${poiId}` : 'poi_id is NULL';
```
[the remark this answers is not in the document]
avalmas-programize: I changed that so on when specifying a poi on your params you will also get the data of the reef (poi_id is NULL)

packages/api/src/time-series/time-series.service.ts
avalmas-programize: Same as above. By grabbing both poiId values and NULL we are getting the ranges for the poi and also for the reef

packages/api/src/time-series/time-series.service.ts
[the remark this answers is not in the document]
ericboucher: ok but then it's grouped by source, but the source for the reef is called "null"?

ericboucher (approved):

packages/api/src/time-series/time-series.service.ts
ericboucher: Follow #495 for a future resolution

PR #496, opened 2021-03-23
Get spotter data from time series table [frontend]

tleraj:
The purpose of this PR is to use the time series spotter data for the charts and also close https://github.com/aqualinkorg/aqualink-app/issues/501.

- [x] align the two charts (expand the bottom one)
- [x] align start and end dates with the edges of the chart 
- [x] check drop shadows on range selectors, they should be the same as the ones we have on cards
- [x] less space around min/mean/max column in stats box (need a change from https://github.com/aqualinkorg/aqualink-app/pull/503)


Ry-DS:
Committed changes from #503 here by accident haha, ignore commit https://github.com/aqualinkorg/aqualink-app/pull/496/commits/9db29b9192f668737d1a794060eda1951987b42f

ericboucher (approved):

tleraj:
@Ry-DS The `cardWithSpotter` and `cardWithoutSpotter` classes in `src/common/Chart/ChartWithCard/index.tsx` look almost exactly the same, except that they use a different breakpoint. Any good ideas for reducing this duplication here?

PR #497, opened 2021-03-23
Add NETCDF libraries during deployment

ericboucher:


packages/api/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# Install netcdf
+RUN apt update
+RUN apt install -y sudo
+RUN sudo apt install -y libnetcdf-dev
+
+# Run application
+ENV NODE_ENV=production
+WORKDIR /app
+
+COPY ["package.json", "package-lock.json*", "./"]
```
avalmas-programize, line 12: Shouldn't we copy the `yarn.lock` since we are using yarn instead of npm?

ericboucher:
@tleraj we kept having issues with the tests so I mocked the date picker but it looks like we have some inconsistencies with `localizedMaxDate` which doesn't give the same result during CI and locally

PR #499, opened 2021-03-25
Move title up

### LINKED ISSUE #478: Move front page title text up.
See screenshot below. Move the entire text area, including map buttons. 
Screen Shot 2021-03-01 at 10 43 16 AM

ericboucher:
Fixes #478


Ry-DS (approved):
If verified on your end that it works, all good :+1:
Test link would be good too.

Issue #501, opened 2021-03-26
UI tweaks for new charts


ericboucher:
- align the two charts (expand the bottom one)
- less space around min/mean/max column in stats box
- align start and end dates with the edges of the chart
- check drop shadows on range selectors, they should be the same as the ones we have on cards

PR #503, opened 2021-03-27
Add CSV Download button

### LINKED ISSUE #500: Add download button to new charts
csv (timestamp, metric, value)

metric could be (if the names in ts are good, keep those)

spotterBottomTemp
spotterTopTemp
dailySST
hoboTemp

Screen Shot 2021-03-24 at 11 36 07 AM

Ry-DS:
Test: https://aqua-csv.surge.sh
Closes #500 
Works towards #501 
|1 | 2 |
| -- | -- |
| image | image

Points of discussion:
- The `dailySST` dates differ a lot from all other points. This means (most of the time) most rows don't have `dailySST` defined for a given timestamp (timestamps are precise and frequent because of the frequency of spotters and hobo sensors). Is this ok? Or should we always try fill `dailySST` with the closest match.


tleraj:
@Ry-DS With your latest changes, if I select a range that also has spotter data, this is how it looks like on desktop:

Screenshot_20210330_114724

Ry-DS:
> @Ry-DS With your latest changes, if I select a range that also has spotter data, this is how it looks like on desktop:
> 
> Screenshot_20210330_114724

Thanks for the catch! What reef id was that so I can open it on my end?

tleraj:
> > @Ry-DS With your latest changes, if I select a range that also has spotter data, this is how it looks like on desktop:
> > Screenshot_20210330_114724
> 
> Thanks for the catch! What reef id was that so I can open it on my end?

Unfortunately it's a local reef of mine to which I have added manually a spotter id, so that I could test this behavior.

Ry-DS:
> > > @Ry-DS With your latest changes, if I select a range that also has spotter data, this is how it looks like on desktop:
> > > --image snipped--
> > 
> > 
> > Thanks for the catch! What reef id was that so I can open it on my end?
> 
> Unfortunately it's a local reef of mine to which I have added manually a spotter id, so that I could test this behavior. The bug is still there 🙁

hmm, I'll try recreate this somehow on my end :+1:

tleraj:
> > > > @Ry-DS With your latest changes, if I select a range that also has spotter data, this is how it looks like on desktop:
> > > > --image snipped--
> > > 
> > > 
> > > Thanks for the catch! What reef id was that so I can open it on my end?
> > 
> > 
> > Unfortunately it's a local reef of mine to which I have added manually a spotter id, so that I could test this behavior. The bug is still there slightly_frowning_face
> 
> hmm, I'll try recreate this somehow on my end

Make sure to run a backfil in order to get spotter data. This is just running the command `yarn backfill-spotter-data -d 300` (backfill 300 days of data).

Ry-DS:
@tleraj let me know if you still face this issue. I managed to replicate it by hard coding 4 columns. Thanks again for letting me know :tada:

tleraj:
@Ry-DS Card alignment issue fixed :wink:

tleraj:
@Ry-DS I have a point for discussion here. When I select maximum range, the chart gets stuck at loading for a long time (an issue not met in the latest deployment of `use-time-series` [here](https://aqualink-app-programize.web.app/reefs/1632)). What I think that causes this behavior, is that we read a very big chunk of data from `redux` (`useSelector(reefTimeSeriesDataSelector)`) in two places now, and on the second place (the csv button component) we also process this data in order to create the csv. I think we should try and pass this data to `<DownloadCSVButton ... />` component via props (so that we get rid of the second read from `redux`), and if that doesn't work, then we should figure out another approach. Let me know what you think :slightly_smiling_face:

Ry-DS:
> @Ry-DS I have a point for discussion here. ...snipped

Fixed the performance issues. Good catch !

packages/website/src/common/Chart/ChartWithCard/DownloadCSVButton.tsx
```diff
@@ -0,0 +1,124 @@
...
+      constructCSVData({
+        dailyData: granularDailyData,
+        spotterTopTemp: spotterData?.surfaceTemperature,
+        spotterBottomTemp: spotterData?.bottomTemperature,
+        hoboData: hoboData?.bottomTemperature,
+      }),
+    [granularDailyData, hoboData, spotterData]
+  );
+  return (
+    <CSVLink filename={`export-${startDate}-to-${endDate}.csv`} data={csvData}>
```
tleraj, line 110: Just a styling preference here. Personally I'm not a huge fan of button texts being underlined on hover. Maybe a class with `textDecoration: "none"`? Totally up to you :slightly_smiling_face: 

Screenshot_20210330_173406

tleraj (commented):
I am still facing the performance issue we discussed earlier. Visiting [this](https://aqua-csv.surge.sh/reefs/1632) reef and selecting `MAX` range results to page being stuck in loading state for a long time and in some cases (like mine) it displays the chart normally at the end and in some other cases (like @avalmas-programize) the page crashed. In both cases the loading period was very long :slightly_frowning_face:

PR #507, opened 2021-04-01
Show historic max line and values for small intervals

tleraj:
The purpose of this PR is to fix the `HISTORIC` column hiding for small ranges issue.
The previous approach was looking for `maxMonthly` values in the range of `[pickerStartDate, pickerEndDate]`, and if it didn't find anything it returned nothing. Now, if no value exists to this interval, we return an array of the form:
```
[{pickerStartDate, valueClosestToStartDate}, ...filteredData, {pickerEndDate, valueClosestToEndDate}]
```

Screenshot_20210401_182145

### DIFF
--- a/packages/website/src/common/Chart/ChartWithCard/index.tsx
+++ b/packages/website/src/common/Chart/ChartWithCard/index.tsx
@@ -365,8 +365,8 @@ const ChartWithCard = ({
               )}
             >
               <DownloadCSVButton
-                startDate={startDate}
-                endDate={endDate}
+                startDate={pickerStartDate}
+                endDate={pickerEndDate}
                 reefId={reef.id}
                 pointId={pointId}
                 className={classes.button}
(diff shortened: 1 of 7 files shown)


packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -126,91 +121,42 @@ const ChartWithCard = ({
...
-      ).toISOString();
-      const reefLocalStartDate = new Date(
-        setTimeZone(
-          new Date(moment(pickerStartDate).format("MM/DD/YYYY")),
-          reef?.timezone
-        )
-      ).toISOString();
+    const pickerLocalEndDate = new Date(
+      setTimeZone(
+        new Date(moment(pickerEndDate).format("MM/DD/YYYY")),
```
Ry-DS, line 133: does `moment.toDate` work here?

Ry-DS (approved):
Looks good :+1: Just a couple small things:

packages/website/src/common/Chart/ChartWithCard/index.tsx
tleraj: I think it looks fine for now. In any case, we are going to use another library for date handling, so all of these will be refactored in a future PR :wink:

PR #509, opened 2021-04-01
[RELEASE] Transition to hosted timeseries and integrate HOBO data

ericboucher:
Let's all do a final review now that all the pieces are put together


packages/api/cloud-functions/index.ts
```diff
@@ -90,3 +91,28 @@ exports.pingService = functions.pubsub
...
+    // eslint-disable-next-line no-undef
+    const entities = dbEntities.map(extractEntityDefinition);
+    const conn = await createConnection({
+      ...dbConfig,
+      url: dbUrl,
+      entities,
+    });
+    try {
+      await runSpotterTimeSeriesUpdate(conn);
+      // eslint-disable-next-line no-console
```
Ry-DS, line 113: We probably don't need this rule for the `api` package, since console logs are normal in an API context. Thoughts?

packages/api/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN yarn global add typescript
+RUN yarn global add ts-node
+
+# Run application
+ENV NODE_ENV=production
+WORKDIR /app
+
+# TODO - Find a way to get the yarn.lock which is "out of context".
+# Because we are in a sub-package it is located at ../../yarn.lock
+COPY ["package.json", "./"]
```
Ry-DS, line 17: Not too experienced with Docker, but one approach might be to have all the docker files live in the root folder, so the yarn lock can be accessed. Not sure if its worthwhile however since we cannot sort them into their rightful folders anymore.

packages/api/scripts/sst-backfill.ts
```diff
@@ -0,0 +1,66 @@
+import { createConnection, In } from 'typeorm';
+import Bluebird from 'bluebird';
+import { Point } from 'geojson';
+import { Reef } from '../src/reefs/reefs.entity';
+import { getNOAAData } from './utils/netcdf';
+import { DailyData } from '../src/reefs/daily-data.entity';
+
+const dbConfig = require('../ormconfig');
+
+const selectedReefs = [];
```
Ry-DS, line 10: we should type this, since the array type is ambiguous right now.

packages/api/scripts/sst-backfill.ts
```diff
@@ -0,0 +1,66 @@
...
+      selectedReefs.length > 0
+        ? {
+            id: In(selectedReefs),
+          }
+        : {},
+  });
+
+  const dailyDataEntities = PalauReefs.reduce((entities: DailyData[], reef) => {
+    console.log(`Processing reef ${reef.name || reef.id}...`);
+    const allYearEntities = [2017, 2018, 2019, 2020].reduce(
```
Ry-DS, line 27: we should move this array to the top and use as a constant. Also would be nice to add types to the reduce

packages/api/scripts/sst-backfill.ts
```diff
@@ -0,0 +1,66 @@
...
+  const PalauReefs = await reefRepository.find({
+    where:
+      selectedReefs.length > 0
+        ? {
+            id: In(selectedReefs),
+          }
+        : {},
+  });
+
+  const dailyDataEntities = PalauReefs.reduce((entities: DailyData[], reef) => {
```
Ry-DS, line 25: We should type this reduce, either with generics or with a cast on the `[]` at the second argument.

packages/api/scripts/utils/netcdf.ts
```diff
@@ -0,0 +1,54 @@
+import { times } from 'lodash';
+import moment from 'moment';
+import { Extent, pointToIndex } from '../../src/utils/coordinates';
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+const netcdf4 = require('netcdf4');
```
Ry-DS, line 6: Why can't we import this normally with ES6 imports?

packages/website/src/common/Chart/ChartWithCard/Chart.tsx
```diff
@@ -0,0 +1,245 @@
...
+      )}
+    </>
+  );
+};
+
+const styles = () =>
+  createStyles({
+    chart: {
+      height: 285,
+      marginBottom: 16,
```
Ry-DS, line 218: `margin: 16px 0` maybe?

packages/website/src/common/Chart/ChartWithCard/Chart.tsx
```diff
@@ -0,0 +1,245 @@
...
+    },
+
+    datePickersWrapper: {
+      margin: "0 7px 0 27px",
+    },
+  });
+
+interface ChartIncomingProps {
+  reef: Reef;
+  pointId: number | undefined;
```
Ry-DS, line 229: the `| undefined` could be replaced with optionals (`?:`), but up to you

packages/website/src/common/Chart/ChartWithCard/DownloadCSVButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+  | "dailySST";
+
+interface CSVRow extends Partial<Record<CSVDataColumn, number>> {
+  timestamp: string;
+}
+
+const DATE_FORMAT = "YYYY_MM_DD";
+
+/**
+ * Construct CSV data to pass into react-csv
```
Ry-DS, line 25: We aren't using `react-csv`, just `download-csv`. More detail about args would be nice in JSDoc

packages/website/src/helpers/map.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (point.type === "Point") {
+    return point.coordinates;
+  }
+
+  const coordArray = point.coordinates[0];
+  const len = coordArray.length;
+  const lngArray = coordArray.map((item) => item[0]);
+  const latArray = coordArray.map((item) => item[1]);
+
+  const lngMean = lngArray.reduce((a, b) => a + b) / len;
```
Ry-DS, line 14: I think there's a `mean` function by lodash we could use here.

packages/website/src/helpers/surveys.ts
```diff
@@ -28,4 +34,15 @@ const filterSurveys = (
   );
 };
 
-export default filterSurveys;
+export const findImagesAtSurveyPoint = (
+  surveyList: SurveyListItem[],
+  point: number
+) => {
+  const imagesAtPoint = surveyList.map(
+    (survey) => survey.surveyPointImage![point]
```
Ry-DS, line 42: Is this safe? Is there a null behaviour we could consider? Otherwise, all good

packages/website/src/routes/Surveys/View/index.tsx
```diff
@@ -78,76 +78,32 @@ const SurveyViewPage = ({ reef, surveyId, classes }: SurveyViewPageProps) => {
...
-        <Grid
-          style={{ margin: "4rem 0 1rem 0" }}
-          container
-          alignItems="center"
-          item
-          xs={11}
-        >
+  }, [dispatch, pointId, reef.id, surveyDetails]);
+
+  return loading ? (
```
Ry-DS, line 99: maybe
```ts
if(loading)
return <Loading/>
return <Rest/>
```
so it's easier to read.

Ry-DS (commented):
Quick review since most was reviewed already.

tleraj:
@Ry-DS Addressed your comments in https://github.com/aqualinkorg/aqualink-app/pull/514/commits/12db96cd5665708dba97688341e11369ccfe5c35 :wink:

tleraj (approved):
Looks good :+1:

packages/api/Dockerfile
avalmas-programize: If we moving all files to the root directory, we will have to upload all the project on every deployment. (or add them to gcloudignore)
As a better solution, we could just copy the `yarn.lock` from the root directory to each package directory before the deployment runs and then delete it afterwards

packages/api/cloud-functions/index.ts
avalmas-programize: Using console is treated as a warning to remind us of leftover `console.log` used for debugging.
In our API we use the `Logger` class provided by Nest which is much cleaner
However in the `cloud-functions` Nest is not in context so we use console. We can disable the warning for the folder since it is our only means of output. Thoughts @ericboucher ?

Ry-DS:
> @Ry-DS Addressed your comments in [12db96c](https://github.com/aqualinkorg/aqualink-app/commit/12db96cd5665708dba97688341e11369ccfe5c35) 😉

Awesome! Thanks for touching on the review comments from my PR aswell :) I'll push into your branch to elaborate on the JSDoc. Feel free to make a quick PR so we can merge that commit here.

packages/api/Dockerfile
Ry-DS: That's a good idea :+1: not using the yarn.lock could lead to some hard-to-track bugs down the line from mismatched dependencies, so I think its worthwhile to try out at least something to make sure we build with yarn.lock. The approach you mentioned sounds good!

packages/api/scripts/sst-backfill.ts
avalmas-programize: From the type declaration of the `reduce` function this is properly typed because we have added a type on `previousValue` which is declared to have the same type as the initialValue
```js
reduce<U>(callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: T[]) => U, initialValue: U): U;
```

packages/api/scripts/utils/netcdf.ts
avalmas-programize: I have added a new declaration file and used the import syntax

avalmas-programize:
@Ry-DS comments addressed on https://github.com/aqualinkorg/aqualink-app/pull/517

avalmas-programize (approved):
Seems good to me as well.

packages/api/cloud-functions/index.ts
ericboucher: sounds good to me

packages/api/scripts/sst-backfill.ts
```diff
@@ -0,0 +1,69 @@
+import { createConnection, In } from 'typeorm';
+import Bluebird from 'bluebird';
+import { Point } from 'geojson';
+import { Reef } from '../src/reefs/reefs.entity';
+import { getNOAAData } from './utils/netcdf';
+import { DailyData } from '../src/reefs/daily-data.entity';
+
+const dbConfig = require('../ormconfig');
+
+const reefsToProcess = [];
```
Ry-DS, line 10: Is this typed?

packages/api/src/reefs/reefs.service.ts
```diff
@@ -221,20 +217,33 @@ export class ReefsService {
...
+    if (!moment(start).isValid() || !moment(end).isValid()) {
+      throw new BadRequestException('Start or end is not a valid date');
+    }
+
+    return this.dailyDataRepository
+      .createQueryBuilder('daily_data')
+      .where('reef_id = :id', { id })
+      .orderBy('date', 'DESC')
+      .andWhere('date <= :endDate', {
+        endDate: (end && new Date(end)) || new Date(),
```
Ry-DS, line 240: `new Date(end || undefined)` works? Same for `start`, but undefined is 0.

packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -0,0 +1,404 @@
...
+        break;
+      case "one_year":
+        setPickerEndDate(localizedMaxDate);
+        setPickerStartDate(subtractFromDate(localizedMaxDate, "year"));
+        break;
+      case "max":
+        setPickerEndDate(localizedMaxDate);
+        setPickerStartDate(localizedMinDate);
+        break;
+      default:
```
Ry-DS, line 194: Is eslint forcing you to have a default? Since this seems redundant. If it is, all good to keep it as is.

packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -0,0 +1,404 @@
...
+        default:
+          break;
+      }
+    }
+  };
+
+  return (
+    <Container
+      disableGutters={disableGutters}
+      className={`${classes.chartWithRange} ${
```
Ry-DS, line 234: I would use the classNames library to make this more readable, but if it's not installed we can leave as is.

Ry-DS (approved):
Looks good! A few more things but not urgent.

packages/api/scripts/sst-backfill.ts
avalmas-programize: This has already been fixed by  https://github.com/aqualinkorg/aqualink-app/pull/509/commits/5ccdc053b493426089117f5a68641287d3c12339

packages/api/src/reefs/reefs.service.ts
avalmas-programize: `new Date(undefined)` returns `Invalid Date` so although being a really good idea, it doesn't work.

packages/website/src/common/Chart/ChartWithCard/index.tsx
tleraj: Yes eslint rule

PR #512, opened 2021-04-03
Add reef_id and zoom params support for the /map endpoint

abouillis:
This branch was created on top of the PR #509. It must be merged after the PR #509 approvment.


packages/website/src/routes/HomeMap/Map/index.tsx
```diff
@@ -41,6 +42,13 @@ const currentLocationMarker = L.divIcon({
   iconSize: L.point(16, 16, true),
 });
 
+function useQueryZoomParam() {
+  const query = new URLSearchParams(useLocation().search);
+  const zoomParam = query.get("zoom");
+  const zoom: number = zoomParam ? +zoomParam : INITIAL_ZOOM;
```
ericboucher, line 48: You need to find a safer way to do this. Currently, entering zoom=hello breaks the app

ericboucher, line 48: You will need to find a safer way to do this. Currently, entering zoom=hello breaks the app

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -11,21 +12,43 @@ import SwipeableBottomSheet from "react-swipeable-bottom-sheet";
...
+  setReefOnMap,
+} from "../../store/Homepage/homepageSlice";
+
 import { surveysRequest } from "../../store/Survey/surveyListSlice";
+import { Reef } from "../../store/Reefs/types";
 
-const featuredReefId = process.env.REACT_APP_FEATURED_REEF_ID || "";
+function useQueryReefIParam() {
+  const query = new URLSearchParams(useLocation().search);
+  return query.get("reef_id") || process.env.REACT_APP_FEATURED_REEF_ID || "";
```
ericboucher, line 30: Same here, you will need to make sure that the ID is valid AND that it is indeed in the list of available reefs. If you enter a random reef_id, then the `Selected Site` section stays empty and we get 400 errors from the backend in the console.

ericboucher (changes requested):
This is a great start!

Let's make sure that the url params are parsed in a safe way to avoid any errors. See comments inline.

And keep the current behavior of the app, on start, we want to zoom to the reef_id but NOT open the tooltip.

packages/website/src/routes/HomeMap/Map/index.tsx
abouillis: I changed the zoom parameter management.

packages/website/src/routes/HomeMap/index.tsx
abouillis: I completely changed the way I process the reef_id param and I add the reef existence checking.

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -56,15 +95,19 @@ const Homepage = ({ classes }: HomepageProps) => {
     drawer.scrollTop = 0;
   });
 
-  return (
+  return reefsList.length > 0 && featuredReefId && !featuredReef ? (
```
ericboucher, line 98: I'm not a huge fan of the "not found" here. I think I would prefer to test if the queryParam reef_id is in the reefsList, and if not default to `REACT_APP_FEATURED_REEF_ID`. And not change the current behavior.

Note that the current default behavior is to focus on the reef with the worst alert level (most dire situation). And the featuredReef is independent from that zoom. We want to keep that behavior as the default. The behavior should only change if there is a reef_id URL parameter.

ericboucher: Testing for reef_id in reefsList will also help you avoid dispatching a reef that you know doesn't exist and get 400 responses from the backend

Eg `http://localhost:3000/map?reef_id=hello` creates requests to the backend and of course gets 400s back

abouillis: Ok, thanks for your feedback, I didn't really understand what's the purpose of the featured reef versus the reef_id URL parameter. It's a bit more clear now. I just pushed a update where I removed the "not found" and set the behaviour you described.

packages/website/src/routes/HomeMap/Map/index.tsx
```diff
@@ -187,6 +188,6 @@ const styles = () =>
     },
   });
 
-type HomepageMapProps = WithStyles<typeof styles>;
+type HomepageMapProps = WithStyles<typeof styles> & { zoomLevel: number };
```
Ry-DS, line 191: If we are adding props, we should make a new interface to hold these props
```ts
interface HomepageMapIncomingProps {
    zoomLevel: number;
}
```

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -11,21 +12,56 @@ import SwipeableBottomSheet from "react-swipeable-bottom-sheet";
...
+  setMapInitialReef,
+} from "../../store/Homepage/homepageSlice";
+
 import { surveysRequest } from "../../store/Survey/surveyListSlice";
+import { Reef } from "../../store/Reefs/types";
 
-const featuredReefId = process.env.REACT_APP_FEATURED_REEF_ID || "";
+enum QueryParamKeys {
+  reefId = "reef_id",
+  zoomLevel = "zoom",
```
Ry-DS, line 31: Maybe case enum constants as `ZOOM_LEVEL` and `REEF_ID`? Not too important, up to you :)

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -11,21 +12,56 @@ import SwipeableBottomSheet from "react-swipeable-bottom-sheet";
...
+}
+
+function useQuery() {
+  return new URLSearchParams(useLocation().search);
+}
 
 const Homepage = ({ classes }: HomepageProps) => {
   const dispatch = useDispatch();
   const reefOnMap = useSelector(reefOnMapSelector);
+  const reefsList = useSelector(reefsListSelector) || [];
```
Ry-DS, line 41: I like this approach, but I feel this should all be moved into a hook, since this code could really be anywhere (not dependant on HomeMap). Doing so will clean things up and help make this feature easier to track. Maybe you can expand the new `useQuery` hook?

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -34,7 +70,7 @@ const Homepage = ({ classes }: HomepageProps) => {
       dispatch(reefRequest(`${reefOnMap.id}`));
       dispatch(surveysRequest(`${reefOnMap.id}`));
     }
-  }, [dispatch, reefOnMap]);
+  }, [dispatch, featuredReefId, reefOnMap]);
```
Ry-DS, line 73: Was this from eslint? Surprised it wasn't caught before

packages/website/src/routes/HomeMap/Map/index.tsx
Ry-DS: Also, maybe we name this `initialZoom`? One might think changing this value actively changes the zoom.

packages/website/src/routes/HomeMap/Map/Markers/index.tsx
```diff
@@ -84,13 +87,20 @@ export const ReefMarkers = () => {
               <Marker
                 onClick={() => {
                   dispatch(setSearchResult());
+                  dispatch(unsetMapInitialReef());
                   dispatch(setReefOnMap(reef));
```
Ry-DS, line 91: this is a longshot, but maybe we can unset the initial reef with `setReefOnMap`? Might not work for all cases but might be worth a shot since it seems like everywhere `setReefOnMap` is, `unsetMapInitialReef` follows.

Ry-DS (commented):
Coming along well! A few suggestions:

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -11,21 +12,56 @@ import SwipeableBottomSheet from "react-swipeable-bottom-sheet";
...
+    return reef.id.toString() === queryParamReefId;
+  });
 
   useEffect(() => {
     dispatch(reefsRequest());
   }, [dispatch]);
 
+  useEffect(() => {
+    if (queryParamReef) {
+      dispatch(setMapInitialReef(queryParamReef));
```
tleraj, line 60: Not sure why we need `mapInitialReef`. It seems that wherever we `setMapInitialReef` we also `setReefOnMap`, and same with `unsetMapInitialReef`. I feel like we could achieve our goal by using only `reefOnMap`. If there is a deeper reason, then ignore this :slightly_smiling_face:

packages/website/src/routes/HomeMap/Map/Markers/index.tsx
```diff
@@ -84,13 +87,20 @@ export const ReefMarkers = () => {
...
                 key={`${reef.id}-${offset}`}
                 icon={buoyIcon(alertIconFinder(weeklyAlertLevel))}
                 position={[lat, lng + offset]}
               >
-                <Popup reef={reef} autoOpen={offset === 0} />
+                <Popup
+                  reef={reef}
+                  autoOpen={
+                    offset === 0 &&
+                    reef.id !== (initialReef ? initialReef.id : 0)
```
tleraj, line 101: Not sure if it could possibly happen, but what if there is a reef with id 0? Maybe
```javascript
reef.id !== initialReef?.id
```

packages/website/src/routes/HomeMap/ReefTable/body.tsx
```diff
@@ -118,6 +119,7 @@ const ReefTableBody = ({ order, orderBy, classes }: ReefTableBodyProps) => {
   const handleClick = (event: unknown, reef: Row) => {
     setSelectedRow(reef.tableData.id);
     dispatch(setSearchResult());
+    dispatch(unsetMapInitialReef());
```
tleraj, line 122: Same as my comment above

tleraj (commented):
Looks good. Just some comments along with @Ry-DS's suggestions :slightly_smiling_face:

ericboucher:
@abouillis the implementation was causing a weird behavior because of a conflict between the findMaxAlertReefPosition and the new reefId initialization so I tried another approach. Let me know what you think. Last thing to wrap up is address Ryan's comment and extract the hook and we should be good to go!

ericboucher (approved):

packages/website/src/routes/HomeMap/index.tsx
abouillis: Yes, I had to add featuredReefId to clear the eslint warning.

abouillis:
@ericboucher ok I understand your approach, it's more clear. I try it but I didn't take that way, because without using setReefOnMap, the reef is not selected into the reefs list at the bottom right. I thought it was important for the screen consistence. setReefOnMap seems to centralize the triggering of the behaviours of all the components. N.B: You still have a strange behaviour I noticed: when the reef_id parameter is used, there is a double data initialization into the "Selected site" block, one to load the featured reef's data, and another to load the reef from the reef_id parameter. It finalize with a delay after the full page display.

packages/website/src/routes/HomeMap/index.tsx
abouillis: I moved the parameters processing into the useQuery hook.

abouillis:
@ericboucher I moved the parameters processing into the useQuery hook.

ericboucher:
Good catch! Any idea where it could come from? I'll try to track it down.

And good point on the screen consistency. I think it is fine for now to keep things simpler.

ericboucher:
I think I got it, we didn't check that the reefList was loaded before loading the featuredReef but we wait for the list to check the existence of the initialReef which is causing a double fetch. I have a fix in progress

ericboucher:
@abouillis I think I fixed it but I am not sure this is the most efficient! Any ideas on how to improve the efficiency/speed of the first load?

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -44,10 +44,12 @@ function useQuery() {
...
   const reefsList = useSelector(reefsListSelector) || [];
   const featuredReefId = process.env.REACT_APP_FEATURED_REEF_ID || "";
-  const initialReefId =
-    findReefById(reefsList, queryParamReefId)?.id.toString() ||
-    findReefById(reefsList, featuredReefId)?.id.toString() ||
-    "";
+  const initialReefId = queryParamReefId
+    ? findReefById(reefsList, queryParamReefId)?.id.toString() ||
+      findReefById(reefsList, featuredReefId)?.id.toString() ||
+      ""
```
tleraj, line 50: Is this necessary? Couldn't we just say
```javascript
const initialReefId = queryParamReefId
    ? findReefById(reefsList, queryParamReefId)?.id.toString() ||
      findReefById(reefsList, featuredReefId)?.id.toString()
    : featuredReefId;
```

ericboucher: no because the queryParamId could be a fake input, eg. "hello" but exist
and the featuredId not exist = "426225" so we would end up with an undefined

abouillis:
@ericboucher I think it was a good way but I think we could improve the first load.
I noticed there is a double call to the reefs endpoint. 

I found a useEffect hook containing an explicit call with dispatch(reefsRequest()); into the HomeMap and the ReefList components.

Is that useEffect hook really needed into the both components? I tried to remove them without visible behaviour change.

PR #513, opened 2021-04-03
Fix breaking NavBar and Sign In/Up dialog error issue

tleraj:
The purpose of this PR is to add the following UI tweaks:

### Navbar `SIGN IN`/`SIGN UP` buttons
For several screen sizes, these buttons were getting too squished and were collapsing one below the other.
| before | after |
|---|---|
| Screenshot_20210404_023514 | Screenshot_20210404_023611 |

### Sign In/Up dialog persistent error issue
If there is a sign in or register error and the the dialog gets re-opened, the error message was still there and not cleared as it should be. Fixed that and also did some refactoring.


packages/website/src/common/NavBar/index.tsx
```diff
@@ -37,6 +39,8 @@ const NavBar = ({
 }: NavBarProps) => {
   const user = useSelector(userInfoSelector);
   const dispatch = useDispatch();
+  const theme = useTheme();
+  const isTablet = useMediaQuery(theme.breakpoints.up(768));
```
Ry-DS, line 43: is it possible to use a string `'md'` instead of a constant number?

packages/website/src/common/RegisterDialog/index.tsx
```diff
@@ -53,24 +50,23 @@ const RegisterDialog = ({
...
-      };
-      dispatch(createUser(registerInfo));
-    },
-    [dispatch]
-  );
+  const onSubmit = (
+    data: any,
+    event?: BaseSyntheticEvent<object, HTMLElement, HTMLElement>
+  ) => {
+    if (event) {
```
Ry-DS, line 57: can we optional chain a function call?

packages/website/src/common/RegisterDialog/index.tsx
```diff
@@ -53,24 +50,23 @@ const RegisterDialog = ({
...
-        organization: data.organization,
-        email: data.emailAddress,
-        password: data.password,
-      };
-      dispatch(createUser(registerInfo));
-    },
-    [dispatch]
-  );
+  const onSubmit = (
+    data: any,
```
Ry-DS, line 54: if you know the type of this, would be awesome to type this now :eyes:

packages/website/src/common/SignInDialog/index.tsx
```diff
@@ -77,35 +56,42 @@ const SignInDialog = ({
...
-      if (event) {
-        event.preventDefault();
-      }
-      dispatch(resetPassword({ email: data.emailAddress }));
-      setPasswordResetEmail(data.emailAddress);
-    },
-    [dispatch]
-  );
+  const onSubmit = (
+    data: any,
```
Ry-DS, line 60: same comments above apply here

Ry-DS (commented):
Nice! Most comments are focused on cleaning up the old code - would be nice to fix it up now

packages/website/src/common/RegisterDialog/index.tsx
tleraj: You mean like `event?.preventDefault();`? This raises the following `eslint` error
```
Expected an assignment or function call and instead saw an expression.
```

ericboucher (approved):

PR #515, opened 2021-04-05
Spotter backfill up to now instead of yesterday

avalmas-programize:
Start spotter backfill from the current day

### DIFF
--- a/packages/api/src/utils/spotter-time-series.ts
+++ b/packages/api/src/utils/spotter-time-series.ts
@@ -103,14 +103,8 @@ export const addSpotterData = async (
       Bluebird.map(
         times(days, (i) => i),
         (i) => {
-          const startDate = moment()
-            .subtract(i + 1, 'd')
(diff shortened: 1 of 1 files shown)


packages/api/src/utils/spotter-time-series.ts
```diff
@@ -103,14 +103,8 @@ export const addSpotterData = async (
...
-          const startDate = moment()
-            .subtract(i + 1, 'd')
-            .startOf('day')
-            .toDate();
-          const endDate = moment()
-            .subtract(i + 1, 'd')
-            .endOf('day')
-            .toDate();
+          const startDate = moment().subtract(i, 'd').startOf('day').toDate();
+          const endDate = moment().subtract(i, 'd').endOf('day').toDate();
```
ericboucher, line 107: Do you think we should limit the endDate to be < `now()`?

ericboucher (approved):

Issue #510, opened 2021-04-02
🐛 Email SIGN UP is case sensitive


ericboucher:
Using an all caps email leads to a 400 from the backend. The expected behavior is for email signup and login to be case-insensitive.

<img width="442" alt="Screen Shot 2021-04-02 at 2 30 49 PM" src="https://user-images.githubusercontent.com/16843267/113455734-45dc4700-93c0-11eb-9244-d652b4341e9e.png">

PR #516, opened 2021-04-05
Make user email case insensitive

avalmas-programize:
https://github.com/aqualinkorg/aqualink-app/issues/510


packages/api/src/users/users.service.ts
```diff
@@ -27,7 +27,9 @@ export class UsersService {
     if (!firebaseUser) {
       throw new BadRequestException('Invalid Firebase token.');
     }
-    if (firebaseUser.email !== createUserDto.email) {
+    if (
```
ericboucher, line 30: So do we know what was happening? Firebase always return the email in lowercase?

ericboucher:
@tleraj maybe we could also solve this in the frontend actually and make sure the email input are of type `email` to control the case.

packages/api/src/users/users.service.ts
avalmas-programize: Yes all user emails are stored in firebase in lowercase

avalmas-programize:
Since we plan to expose the backend API to the public we need have checks in the backend as well

avalmas-programize:
Class/validator has a guard `IsLowercase` that we can use. Quite unexpected but also quite handy
So this will simplify the backend checks

packages/website/src/common/RegisterDialog/index.tsx
```diff
@@ -64,7 +65,7 @@ const RegisterDialog = ({
       const registerInfo: UserRegisterParams = {
         fullName: `${data.firstName} ${data.lastName}`,
         organization: data.organization,
-        email: data.emailAddress,
+        email: data.emailAddress.toLowerCase(),
```
ericboucher, line 68: Ideally we can only type lowercase (eg. ovio.org)

Or at least also transform to lowercase in the sign in

tleraj:
@ericboucher With the latest change, the input value on the email text field is always transformed into lowercase (in both sign in and sign up dialogs) by using a `react-hook-form` workaround
```javascript
onChange={() =>
  setValue(
    "emailAddress",
    getValues("emailAddress").toLowerCase()
  )
}
```

Also the values in the request payloads on the submit functions are transformed into lowercase just in case.

tleraj (approved):
Looks good :+1:

ericboucher (changes requested):
Let's make the backend safe instead of using guards

ericboucher (approved):

ericboucher (approved):

PR #520, opened 2021-04-07
Use require for netcdf4 to fix unresolved import

ericboucher:


avalmas-programize:
Yeah sorry about that, I don't understand how can we make typescript detect the module

PR #522, opened 2021-04-08
Api docs

avalmas-programize:
To view the api docs:
- Start nestJS server
- Navigate from your browser to `/api`

Issues with the documentation:
- 401 (unauthorized), 403 (forbidden) errors are not mentioned
- Response body schema is not accurate since it has no limit to its depth on entities
- User role permissions per endpoint are not handled by swagger

Breaking front-end changes;
**All** entities declared in DTOs have been changed from `model`  to `modelId` in order to more properly describe the contents of each variable.


ericboucher:
@avalmas-programize @tleraj can we make the necessary frontend changes so that this PR is not "breaking" anything?

avalmas-programize:
Changes have already been applied to the front-end in https://github.com/aqualinkorg/aqualink-app/pull/522/commits/f381020eea08be125d38f84d6cdde4408b4bfaf5

avalmas-programize:
By upgrading geoTIFF to 1.0.4 and setting `forceXHR` flag to true seems to resolve issue.

The issue seems to be that swagger provides a partial implementation of the fetch function, which is not the one expected by the geoTIFF package. However because the `fetch` function is now defined, the package thinks it can use it instead of HTTPClient.
So by seeting `forceXHR` we force geoTIFF XMLHttpRequest which is undefined and therefore use the fallback HTTPClient.

PR #523, opened 2021-04-08
Display gaps in charts

tleraj:
The purpose of this PR is to display large gaps of data in the charts rather than interpolating them.

### Desings:
| before | after |
|---|---|
| Screenshot_20210408_180705 | Screenshot_20210408_161405 |


packages/website/src/routes/Surveys/View/SurveyMediaDetails.tsx
```diff
@@ -106,23 +106,34 @@ const SurveyMediaDetails = ({
...
-                      <Link
-                        className={classes.link}
-                        to={`/reefs/${reefId}/points/${point.pointId}`}
-                      >
-                        <Typography
-                          className={`${classes.titleName} ${
-                            point.name === selectedPoi && classes.selectedPoi
-                          }`}
-                          variant="h6"
+                  {point.pointId ? (
```
ericboucher, line 109: what is this change?

tleraj: If I upload a survey and then delete the survey point and visit the survey detail page, the survey point name we display is `Other` but has no `pointId`, and it worked as a link with `undefined` url. With this change, we just display `Other` without the link.

ericboucher (approved):

PR #526, opened 2021-04-09
Remove unnecessary data filtering

tleraj:
The purpose of this PR is to clean up the `ChartWithCard` component, by removing some unnecessary data filtering and relying on the API to fetch correct data inside the desired range.

### DIFF
--- a/packages/website/src/common/Chart/ChartWithCard/index.tsx
+++ b/packages/website/src/common/Chart/ChartWithCard/index.tsx
@@ -271,9 +271,13 @@ const ChartWithCard = ({
             pickerStartDate={pickerStartDate || subtractFromDate(today, "week")}
             pickerEndDate={pickerEndDate || today}
             startDate={startDate || subtractFromDate(today, "week")}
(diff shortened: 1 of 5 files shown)


ericboucher (approved):

packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -296,9 +300,13 @@ const ChartWithCard = ({
               }
               pickerEndDate={pickerEndDate || today}
               chartStartDate={startDate || subtractFromDate(today, "week")}
-              chartEndDate={moment(endDate)
-                .tz(reef.timezone || "UTC")
-                .endOf("day")
+              chartEndDate={moment
```
ericboucher, line 303: Is this change necessary? If it is, and we calculate the date the same way in the Chart and TempAnalysis component, it might be worth to extract these lines at the top and avoid duplicating in these two components

ericboucher (approved):

PR #531, opened 2021-04-14
Update axios library

ericboucher:


avalmas-programize:
The issue seems to be caused by a mismatch in axios version.
In the root directory `axios@0.18` is stored because many other package we use haven't had their axios dependency upgraded yet
So the new `axios@0.21.1` is stored in the `api` and `website` folders
However `axios-retry` which strangely doesn't depend on axios (devDependency only) is also stored in the root directory along with the older version of axios, which it use
So when we import the `axios-retry` in the code the types are fetched from axios@0.18 which are incompatible with axios@0.21.1 and causes the type issue

To solve this I used the [`nohoist`](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) option in the api `package.json`
```json
  "workspaces": {
    "nohoist": ["axios-retry"]
  },
```
I am not a complete fun of this solution as it might cause more incompatibilities in the future, however it is unlikely to add a package that will depend on axios-retry so we might get away with this

ericboucher:
Seems reasonable to me. Thanks for looking into it @avalmas-programize !

PR #532, opened 2021-04-17
Add sensor date range

### LINKED ISSUE #502: Add sensor date range
Add the available sensor date range below the title in grey. (grey area only for now, no button needed)

Screen Shot 2021-03-24 at 11 36 07 AM

jrn2778:
The purpose of this PR is to close #502 

Here is a screenshot of the result:

Screenshot_20210506_192602


ericboucher:
Great start @jrn2778! Ideally, we want to show the availability date for the different sensors.
At the moment, the two sensors that we have are "HOBO" and "Spotter"

So in your screenshot, there is only Spotter data (the two buoy lines) and their availability is shorter than the period displayed.

We should have a range that looks like this:

Spotter Range: 03/05/2021 - 04/15/2021

jrn2778:
@ericboucher Thanks for the input!

Seems like I misunderstood exactly what that date was, thank you for clarifying. I shall see what I can do in order to get that to be the proper date and sensor.

jrn2778:
@ericboucher Perhaps I am in a little over my head as far as the actual science behind it is concerned, but I attempted a fix. (Screenshot has been updated to reflect the most recent state).

Now, it checks the hasSpotterData variable, if true it will grab from spotterData otherwise it will grab from hoboData. The date is coming from the timestamp on the bottomTemperature of the respective data set. First timestamp in the data is used for the start date and last timestamp for the end date.

If you can tell me where this needs improvements, I may be able to continue to improve on it, but I'm just a little lost in regards to the actual science behind it all.

Thank you!

tleraj:
@jrn2778 We are seeking to display the available time ranges for each sensor type. The data we are looking for can be grabbed with `useSelector(reefTimeSeriesDataRangeSelector)`. This selector gives exactly what we want :wink:. It could look like this:

```javascript
const { hobo: hoboRanges, spotter: spotterRanges } =
  useSelector(reefTimeSeriesDataRangeSelector) || {};
const { minDate: hoboMinDate, maxDate: hoboMaxDate } =
  hoboRanges?.bottomTemperature?.[0] || {};
const { minDate: spotterMindDate, maxDate: spotterMaxDate } =
   spotterRanges?.bottomTemperature?.[0] || {};
```

I hope this helped :slightly_smiling_face:

ericboucher:
Hi @jrn2778 please let us know if you will continue your work on this PR or if we should pursue :)

jrn2778:
@ericboucher Hi, sorry for the lack of updates. I've been very busy and probably wont have a chance to continue for at least another week or two, so perhaps it would be better for you to pursue. Thanks for understanding!

ericboucher:
No worries @jrn2778, thanks for giving it a try and come back anytime :)

packages/website/src/common/Chart/ChartWithCard/ViewRange.tsx
```diff
@@ -50,12 +59,44 @@ const ViewRange = ({
...
+    : undefined;
+  const spotterRangeString =
+    spotterFormattedStartDate && spotterFormattedEndDate
+      ? `Spotter range: ${spotterFormattedStartDate} - ${spotterFormattedEndDate}`
+      : undefined;
+  const hoboRangeString =
+    hoboFormattedStartDate && hoboFormattedEndDate
+      ? `HOBO range: ${hoboFormattedStartDate} - ${hoboFormattedEndDate}`
+      : undefined;
+
```
ericboucher, line 93: This seems quite duplicative. We will likely have more sensors in the future so let's try to make it a bit more future proof :)

ericboucher (changes requested):
Nice work both. Let's make it a bit more future proof and then we should be gtg

packages/website/src/common/Chart/ChartWithCard/ViewRange.tsx
```diff
@@ -50,12 +58,19 @@ const ViewRange = ({
     },
   ];
 
+  const spotterRangeString = availableRangeString(
```
ericboucher, line 61: Let's go one step further. This file should get `timeSeriesDataRanges` entirely and use a mapping/array of available sensors instead.

const allSensors = ['HOBO', 'Spotter'] which is in theory the only piece we would need to change to add a new sensor range.

and then something like that:

```
allSensors.map((sensor) => {
	dateRangeString = availableRangeString(sensor, timeSeriesDataRanges[sensor], .....)
	if (!dateRangeString){
	  return null
	} 
	return (<Grid item>
                  <Alert
                    classes={{
                      icon: classes.rangeIcon,
                      root: classes.rangeItem,
                    }}
                    severity="info"
                  >
                    <Typography variant="subtitle2">
                      {dateRangeString}
                    </Typography>
                  </Alert>
                </Grid>)
))
```

What do you think?

ericboucher (approved):

PR #534, opened 2021-04-21
Add collection update functionality

tleraj:
The purpose of this PR is to offer users the ability to edit their collections by adding/removing sites and changing their names. Designs:

| add site | remove site |
|---|---|
| Screenshot_20210428_204836 | Screenshot_20210428_204845 |

| name update enabled |
|---|
| Screenshot_20210421_203847 |

TODOs:
 - [x] `/dashboard`: Display signed in user collection
 - [x] `/collection/:collectionName`: Display the collection with name equal to `collectionName`
 - [x] On redux store, create a new field `collection` that holds information for the collection that is being displayed (either user's or public)
 - [x] On redux `user.userInfo.collection`, we should only hold the collection's reefsList, and any additions/removals should be reflected to that list.
 - [x] Imporve alignment of collection add/remove button on small screens


ericboucher:
For the "add/remove to your collection" I think a simple empty/full star next to the reef name would work well. @drewjgray what do you think?

drewjgray:
@ericboucher I agree! Although I like the bookmark icon you have here as well.

packages/api/src/collections/collections.service.ts
```diff
@@ -233,6 +237,8 @@ export class CollectionsService {
       };
     }, {});
 
+    const hasHoboDataSet = await hasHoboDataSubQuery(this.sourcesRepository);
```
ericboucher, line 240: Dataset

ericboucher: or `hasHoboData`

packages/website/src/assets/unwatch.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 24 24" height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor"><rect fill="none" height="24" width="24"/><path d="M17,11v6.97l-5-2.14l-5,2.14V5h6V3H7C5.9,3,5,3.9,5,5v16l7-3l7,3V11H17z M21,7h-6V5h6V7z"/></svg>
```
ericboucher, line 1: there are no relevant icons in the libraries we are using?

tleraj: MUI offers a generic bookmark icon, but without the `+`/`-` signs. Maybe we could use that and change its color? Or a completely different icon :slightly_smiling_face:

packages/api/src/collections/collections.service.ts
avalmas-programize: Changed to `hasHoboData`

Ry-DS (commented):
Nice! Overall everything is organized quite well

packages/website/src/store/User/helpers.ts
```diff
@@ -11,14 +11,15 @@ export const constructUserObject = async (
   const { data: administeredReefs } = await userServices.getAdministeredReefs(
     token
   );
+  const { id, email, fullName, organization, adminLevel, firebaseUid } = user;
 
   return {
-    id: user.id,
```
Ry-DS: Why not `...user`?

ericboucher (approved):

packages/website/src/routes/Dashboard/index.tsx
```diff
@@ -14,26 +14,56 @@ import {
...
+      dispatch(collectionRequest({ id: urlCollectionId, isPublic: true }));
+    }
+  }, [urlCollectionId, dispatch]);
+
+  useEffect(() => {
+    if (!urlCollectionId && user?.token && user.collection?.id) {
+      dispatch(
+        collectionRequest({
+          id: user.collection.id,
+          isPublic: false,
```
ericboucher, line 49: we shouldn't specify this here. what if the user made their collection public?

packages/website/src/routes/Dashboard/index.tsx
```diff
@@ -14,26 +14,56 @@ import {
...
+    ? collections[urlCollectionName]
+    : undefined;
+
+  useEffect(() => {
+    if (urlCollectionId) {
+      dispatch(collectionRequest({ id: urlCollectionId, isPublic: true }));
+    }
+  }, [urlCollectionId, dispatch]);
+
+  useEffect(() => {
```
ericboucher, line 44: I don't like this behavior here. Are we defaulting to a user's collection if they request a collection that doesn't have an id?

I think we should make the split more explicit.

If you are in `/dashboard` -> collectionId = user.collection?.id

Otherwise, urlCollectionId = urlCollectionName
    ? collections[urlCollectionName] : undefined
    
if !urlCollectionId -> `The collection you requested is not available`.

ericboucher (approved):

PR #538, opened 2021-04-27
Warning - Add migration to deploy steps

ericboucher:
Migrations will be run by default when we deploy to make sure that the API and db are in sync.


avalmas-programize (approved):
I think it looks good

PR #539, opened 2021-04-27
Fix "deployed buoys only " switch not working

Ry-DS:
A `makeStyles` hook wasn't defined properly, meaning we ended up conditionally calling it.
Test:  http://aqua-bug.surge.sh


tleraj (approved):
It looks good :slightly_smiling_face:

ericboucher (approved):

PR #543, opened 2021-04-29
Fix ping url

avalmas-programize:
The ping URL was not correct. An extra `/` was present causing the request to fail with 404 error code.

### DIFF
--- a/packages/api/cloud-functions/index.ts
+++ b/packages/api/cloud-functions/index.ts
@@ -87,7 +87,7 @@ exports.pingService = functions.pubsub
   .onRun(async () => {
     const backendBaseUrl = functions.config().api.base_url;
     console.log('Pinging server');
-    await Axios.get(`${backendBaseUrl}health-check`);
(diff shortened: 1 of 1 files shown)


packages/api/cloud-functions/index.ts
```diff
@@ -87,7 +87,7 @@ exports.pingService = functions.pubsub
   .onRun(async () => {
     const backendBaseUrl = functions.config().api.base_url;
     console.log('Pinging server');
-    await Axios.get(`${backendBaseUrl}/health-check`);
+    await Axios.get(`${backendBaseUrl}health-check`);
```
ericboucher, line 90: is there a safer way to append URLs to avoid this type of issue?

ericboucher (approved):

packages/api/cloud-functions/index.ts
avalmas-programize: Yes there is. I can use the URL object
```js
new URL('health-check', backendBaseUrl).href
```

Issue #541, opened 2021-04-29
Add detailed temperature chart for all sites


ericboucher:
We would like to change the site page by:
- Replace the featured image by the daily temperature chart, renamed "Heat Stress Analysis". Make it a bit similar to the screenshot below with the following changes:
	- only display the last 2 months of data
	- have the title a bit higher, aligned with LAT/LNG maybe?
<img width="1680" alt="Screen Shot 2021-04-28 at 12 21 19 PM" src="https://user-images.githubusercontent.com/16843267/116490106-6aa2cd80-a84b-11eb-8bec-3d027fc34f45.png">

- Display the detailed temperature chart for all sites, renamed "Temperature Analysis"
	- the title can be removed from the stats box
	- there should be at most 3 columns in the analysis box (SST is has the lowest priority)

tleraj:
So for the temp analysis card, the priority is `spotter > hobo > sst`?

tleraj:
Also, should the chart next to the map have a tooltip?

PR #544, opened 2021-05-05
Add detailed temperature chart for all sites

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/541.
Designs:

| spotter | hobo | sst |
|---|---|---|
| spotter | hobo | sst |

TODOs:
 - [x] Fix tests


packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -61,24 +57,22 @@ const ChartWithCard = ({
...
-      case !hasSpotterData && hasHoboData:
-        return classes.largeChart;
-      default:
-        return "";
-    }
-  };
+  const cardType = findCardType(hasSpotterData, hasHoboData, hasDailyData);
+
+  const chartWidthClass =
+    cardType === "spotter" ? classes.smallChart : classes.largeChart;
```
ericboucher, line 75: I think we should re-rethink this in terms of "number of available datasets" instead of "cardType".

With ordering:
- historicalMean
- HOBO
- spotterBottom
- spotterTop
- SST

And how maximum 3 columns for now. What do you think? This was the card width will be linked t a number of columns instead of a "type", what do you think?

packages/website/src/common/Chart/ChartWithCard/Chart.tsx
```diff
@@ -213,11 +229,11 @@ const Chart = ({
...
 };
 
-const styles = () =>
+const styles = (theme: Theme) =>
   createStyles({
     chart: {
-      height: 290,
-      margin: "16px 0",
+      height: 275,
+      margin: `${theme.spacing(1)}px 0`,
```
Ry-DS, line 236: think you can do `theme.spacing(1, 0)`

packages/website/src/common/Chart/ChartWithCard/helpers.ts
```diff
@@ -110,6 +110,23 @@ export const findDataLimits = (
...
+  hasSpotterData: boolean,
+  hasHoboData: boolean,
+  hasDailyData: boolean
+): CardType => {
+  switch (true) {
+    case hasSpotterData:
+      return "spotter";
+    case hasHoboData:
+      return "hobo";
+    case hasDailyData:
```
Ry-DS, line 123: can you do:
```ts
switch{
...
case hasDailyData:
case default:
return "sst";
...}
```

Ry-DS (commented):
Looks good :+1:

packages/website/src/common/Chart/ChartWithCard/index.tsx
tleraj: Since we only display one dataset, it might be a good idea to just rename `type` to `dataset`.

ericboucher (approved):

PR #546, opened 2021-05-06
Possible fix to blinking buoy markers

tleraj:
This PR is likely to fix https://github.com/aqualinkorg/aqualink-app/issues/536. The main idea behind the implementation came from https://developer.apple.com/library/archive/documentation/InternetWeb/Conceptual/SafariVisualEffectsProgGuide/AnimatingWithKeyframes/AnimatingWithKeyframes.html


avalmas-programize:
Changing `stroke-opacity` to `opacity` seems to be the answer.

ericboucher (approved):

PR #552, opened 2021-05-17
Add /sensors endpoint

avalmas-programize:
Three new endpoints have been created
GET /sensors : Fetches all sites with a sensor, includes sensor type and sensor position
Notable properties:
- name: The name of the site
- polygon: The location of the site as Point (GeoJSON)
- spotterId: The id of the sensor
- sensorPosition: The position of the sensor as Point (GeoJSON)
- sensorType: The type of the sensor (currently one is available: 'sofar_spotter'

GET /sensors/:id/data : Fetches all data for the specified sensor `:id`
URL params:
- id: The sensor Id (required)
Query Params:
- startDate: The start date of the range in iso format (required)
- endDate: The end date of the range in iso format (required)
- metrics: An array of metrics to fetch (required)
  - Possible values: `['top_temperature', 'bottom_temperature', 'significant_wave_height', 'wave_peak_period', 'wave_mean_direction', 'wind_speed', 'wind_direction']`

The data will be stored under `response.body.spotter[metric]`

GET /sensors/:id/surveys: Fetches all surveys and survey media (images) for the specified sensor `:id`
URL params:
- id: The sensor Id (required)
The locations where surveys took place are stored under the `surveyMedia[number].poiId.polygon` property,


ericboucher:
Nice start @avalmas-programize, let's drop the "coral-atlas" mention as these endpoints could be used by other apps and integrations in the future

avalmas-programize:
Changed prefix from `coral-atlas` to `sensors`

drewjgray:
@avalmas-programize for the GET surveys endpoint, the response should include the survey media as well as the corresponding sensor data at that time. For example, if there was one picture uploaded on May 18th, the response should include the date, the image, all metadata (comments/observation), as well as the temp (both sensor and SST) at that time. @ericboucher Do we have proposed reeponse structures somewhere?

avalmas-programize:
Thank @bencpeters for the feedback
What data should we include? Date_of_upload plus minus 10 minutes?
Should those data be taken from the poi location? Is reef location also acceptable?
Should data from all sensor be included (hobo, spotter)?
What should we do if there are no data in this window? (there could be NULL values for said parameters)
@ericboucher should we get the data from the time-series table?

ericboucher:
> Thank @drewjgray for the feedback
> What data should we include? Date_of_upload plus minus 10 minutes?
> Should those data be taken from the poi location? Is reef location also acceptable?
> Should data from all sensor be included (hobo, spotter)?
> What should we do if there are no data in this window? (there could be NULL values for said parameters)
> @ericboucher should we get the data from the time-series table?


- Let's grab the closest data within a day (+/- 24 hours)
- Take the data from the reef for now, we don't have SST at the poi level
- Take data from all available sensors
- Let's only return non-null values. It is unlikely that we would have empty values for all since we backfill SST.
- Yes let's get the data from the time series table except for SST, grab the daily data for now

packages/api/src/sensors/sensors.controller.ts
```diff
@@ -21,6 +19,13 @@ export class SensorsController {
   @ApiTimeSeriesResponse()
   @ApiNestNotFoundResponse('No data were found with the specified sensor id')
   @ApiOperation({ summary: 'Get data from a specified sensor' })
+  @ApiParam({ name: 'id', example: 'SPOT-0000' })
```
ericboucher, line 22: From an API perspective, wouldn't it make more sense to use the `id` column of the sensor instead?

avalmas-programize: We don't have a dedicated sensor table for the sensors (for the time being)
The closest thing we have to a unique sensor id is the `spotterId`

packages/api/src/utils/spotter-time-series.ts
```diff
@@ -144,8 +144,8 @@ export const addSpotterData = async (
               ),
             )
             .flat(),
-          (props, i) => {
-            logger.log(
+          (_, i) => {
+            logger.debug(
```
ericboucher, line 148: @avalmas-programize these logs are too noisy, how do we make them less noisy?

I think one log per reef is enough. Eg:

`Spotter data updated for SPOT-O3242 between minStartDate and maxEndDate`

avalmas-programize: I reduced the verbosity by adding a single log after the completion of each reef iteration

ericboucher (approved):

PR #505, opened 2021-03-30
Time series add SST

avalmas-programize:
https://github.com/aqualinkorg/aqualink-app/issues/493

I wasn't able to test the implementation on the sst-backfill script.

PR #556, opened 2021-05-19
FIX - Update 'sofar_api' split in the frontend

ericboucher:
In the #505 PR that was just merged the sensor data schema has changed a bit, which broke the frontend.

This is a first pass at fixing it but I think we should have a better structure for the sensor data in the frontend. Most of them will never have all these fields but simply a subset.

### DIFF
--- a/packages/website/src/mocks/mockHoboDataRange.ts
+++ b/packages/website/src/mocks/mockHoboDataRange.ts
@@ -32,7 +32,20 @@ export const mockHoboDataRange: TimeSeriesDataRange = {
     windSpeed: [],
     windDirection: [],
   },
-  sofarApi: {
+  sofarNoaa: {
+    bottomTemperature: [],
+    alert: [],
+    dhw: [],
+    satelliteTemperature: [],
+    topTemperature: [],
+    sstAnomaly: [],
+    significantWaveHeight: [],
+    wavePeakPeriod: [],
+    waveMeanDirection: [],
+    windSpeed: [],
+    windDirection: [],
+  },
+  sofarGfs: {
     bottomTemperature: [],
     alert: [],
     dhw: [],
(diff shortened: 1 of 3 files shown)


packages/website/src/mocks/mockHoboDataRange.ts
```diff
@@ -32,7 +32,20 @@ export const mockHoboDataRange: TimeSeriesDataRange = {
...
+    alert: [],
+    dhw: [],
+    satelliteTemperature: [],
+    topTemperature: [],
+    sstAnomaly: [],
+    significantWaveHeight: [],
+    wavePeakPeriod: [],
+    waveMeanDirection: [],
+    windSpeed: [],
+    windDirection: [],
```
ericboucher, line 46: @tleraj  I think we should revamp this `TimeSeriesDataRange` format to be more aligned with what we will actually receive most of the time.

Eg. HOBO and spotter will never return dhw or sstAnomaly, ...

tleraj (approved):
Looks good!!

packages/website/src/mocks/mockHoboDataRange.ts
tleraj: I thought that in the future we would have the same exact data from multiple sensors. Let me know which items we should remove :slightly_smiling_face:

ericboucher: no, hobo will only have bottom temperature, noaa will only have SST and dhw, and so on.

What is the backend returning?

antoniosValmas: The backend is returning all metrics as properties with empty arrays in order to be easier to define the type of the response and have a uniform approach.

I can have a look and change it to be more specialized for each source

ericboucher: @avalmas-programize @tleraj please open an issue to discuss this. Maybe it's safer for the backend to return empty arrays but for the frontend to have stronger typing.

PR #560, opened 2021-05-31
Replace daily data with time series on reef map

avalmas-programize:
Replace daily data on reef entity with the time series data
Front-end changes needed: Get latest data from collectionData for map page

### DIFF
--- a/packages/api/migration/1623058289647-AddWeeklyAlertMetric.ts
+++ b/packages/api/migration/1623058289647-AddWeeklyAlertMetric.ts
@@ -1,0 +1,118 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddWeeklyAlertMetric1623058289647 implements MigrationInterface {
+  name = 'AddWeeklyAlertMetric1623058289647';
+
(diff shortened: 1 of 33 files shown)


ericboucher:
Is this ready for review? @avalmas-programize

ericboucher:
I am not sure I understand why we are using `collectionData` here.
Also, the idea is to get the data in the table from the time series data as well

avalmas-programize:
Amended SST script to backfill both degree heating weeks and alert level (daily and weekly)
Rewrote collection util `getCollectionData` to include data only from the `latestData` table
Breaking Front-end changes:
CollectionDataDto is based now on the metrics enum so the keys have been changed (from camelCase to snakeCase)

ericboucher:
@avalmas-programize I deployed on programize. It looks good overall but it feels like the `/map` page is a lot slower than before. Could investigate and see where things are slow?

ericboucher (approved):

ericboucher:
@avalmas-programize can you confirm which backfills we should be running right after this deploy?

avalmas-programize:
We will need to run the sst_backfill script to backfill for dhw, alert, weekly_alert and sst_anomaly

PR #561, opened 2021-06-01
Fix detailed chart not fetching all data on first load bug

tleraj:
The purpose of this PR is to fix the reef detailed chart not fetching all data on first load bug

### DIFF
--- a/packages/website/src/common/Chart/ChartWithCard/index.tsx
+++ b/packages/website/src/common/Chart/ChartWithCard/index.tsx
@@ -28,7 +28,7 @@ import {
   setTimeZone,
   subtractFromDate,
 } from "../../../helpers/dates";
-import { findCardDataset, findDataLimits } from "./helpers";
+import { findCardDataset, findDataLimits, localizedEndOfDay } from "./helpers";
 import { RangeValue } from "./types";
 import ViewRange from "./ViewRange";
 import DownloadCSVButton from "./DownloadCSVButton";
@@ -56,10 +56,7 @@ const ChartWithCard = ({
   const [pickerErrored, setPickerErrored] = useState(false);
   const [range, setRange] = useState<RangeValue>("three_months");
 
-  const today = moment()
-    .tz(reef.timezone || "UTC")
-    .endOf("day")
-    .toISOString();
+  const today = localizedEndOfDay(undefined, reef.timezone);
 
   const dailyDataSst = granularDailyData?.map((item) => ({
     timestamp: item.date,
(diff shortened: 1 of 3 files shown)


packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -56,7 +56,10 @@ const ChartWithCard = ({
   const [pickerErrored, setPickerErrored] = useState(false);
   const [range, setRange] = useState<RangeValue>("three_months");
 
-  const today = new Date(moment().format("MM/DD/YYYY")).toISOString();
+  const today = moment()
```
ericboucher, line 59: why not use `now`?

packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -88,9 +91,21 @@ const ChartWithCard = ({
...
+      const localizedMaxDate = moment(maxDate)
+        .tz(reef.timezone || "UTC")
+        .endOf("day")
+        .toISOString();
+      const pastThreeMonths = moment(
+        subtractFromDate(localizedMaxDate || today, "month", 3)
+      )
+        .tz(reef.timezone || "UTC")
+        .startOf("day")
+        .toISOString();
```
ericboucher, line 103: this might be overkill

packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -88,9 +91,21 @@ const ChartWithCard = ({
   useEffect(() => {
     if (hoboBottomTemperatureRange) {
       const { maxDate } = hoboBottomTemperatureRange?.[0] || {};
-      const pastThreeMonths = subtractFromDate(maxDate || today, "month", 3);
+      const localizedMaxDate = moment(maxDate)
```
ericboucher, line 94: don't we have a util function somwhere to do:

moment(date)
	.tz(reef.timezone || "UTC")
    .endOf("day")
    .toISOString();
    
    
if not, would it be worth it?

packages/website/src/common/Chart/ChartWithCard/index.tsx
tleraj: Without this change, the same issue occurs with the `START DATE` date picker: on first load it fetches data until `pastThreeMonths`, but if I open the date picker and select the same date, the value is set to `pastThreeMonths.startOf("day")`, and fetches new data (with the seemingly same date).

packages/website/src/common/Chart/ChartWithCard/index.tsx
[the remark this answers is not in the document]
tleraj: We need the reef local notion of end of day. With the previous approach, we were setting today to user local start of day, which caused the bug.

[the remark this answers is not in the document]
ericboucher: The data cannot be more recent than "now" though

packages/website/src/common/Chart/ChartWithCard/index.tsx
ericboucher: agreed, but people won't be able to notice because they would be looking at a 3 month window anyway

ericboucher (approved):

Issue #562, opened 2021-06-02
Integrate OceanSense data and charts


ericboucher:
We would like to integrate new datasets coming from OceanSense and display it in new charts.

- The data will come directly from an external API
- Only a very small sample of reefs will be affected

![REEF DETAIL WITH OCEANSENSE_small](https://user-images.githubusercontent.com/16843267/120553943-f90bf100-c3ad-11eb-8fcf-2b3531e7a7cd.jpg)

drewjgray:
Here are the 5 icons for acidity, conductivity, pressure, dissolved oxygen, and oxidation reduction potential. 

[icons.zip](https://github.com/aqualinkorg/aqualink-app/files/6591532/icons.zip)

PR #565, opened 2021-06-07
Integrate OceanSense Data

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/562 and https://github.com/aqualinkorg/aqualink-app/issues/594.

Todo's:
 - [x] Investigate `RangeError: Maximum call stack size exceeded` when trying to display `CONDUCTIVITY (μS)`

The charts y axis limits are set as following:
 - `yAxisMax = dataMaxValue + percentage * dataRange`
 - `yAxisMin = dataMinValue - percentage * dataRange`,

where `dataRange = (dataMaxValue - dataMinValue)` and `percentage = 0.1`

The y axis tick step size is set as the axis range (`yAxisMax - yAxisMin`) divided by the number of ticks we want to display (customizable, set to 5).

## Important note
Right now we store the reefs with ocean sense data into the `src/constants/oceanSenseConfig.ts` file.

### Analysis Card
Screenshot_20210707_143542

### Desktop
| loading | complete |
|---|---|
| loading | desktop |

### Other devices
| iPad Pro | iPad | iPhone 6/7/8 |
|---|---|---|
| iPadPro | iPad | iPhone |


Ry-DS:
It would be pretty cool if we could fit those icons at the top near every corresponding chart :eyes:
what do you think?

packages/website/src/common/Chart/ChartWithCard/TempAnalysis.tsx
```diff
@@ -37,10 +44,16 @@ const TempAnalysis: FC<TempAnalysisProps> = ({
   dailyDataSst,
```
Ry-DS, line 44: Fair to call this file tempanalysis anymore? Just a small thing but not a big deal haha.

packages/website/src/common/Chart/ChartWithCard/TempAnalysis.tsx
```diff
@@ -80,6 +97,13 @@ const TempAnalysis: FC<TempAnalysisProps> = ({
       rows: calculateCardMetrics(2, hoboBottomTemperature, "hobo"),
       display: dataset === "hobo",
     },
+    {
+      title: "SENSOR",
```
Ry-DS, line 101: could consider having the sensor name here, or (unit) SENSOR like `ph Sensor` but up to personal preference

packages/website/src/common/Chart/ChartWithCard/TempAnalysis.tsx
```diff
@@ -80,6 +97,13 @@ const TempAnalysis: FC<TempAnalysisProps> = ({
...
+      key: "oceanSense",
+      color: "#f78c21",
```
Ry-DS, line 103: If this is copied from line 96 consider making it a constant

packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -325,6 +347,73 @@ const ChartWithCard = ({
...
+        constructOceanSenseDatasets(oceanSenseData).map((item) => (
+          <Box mt={4} key={item.title}>
+            <ViewRange
+              range={range}
+              onRangeChange={onRangeChange}
+              disableMaxRange={!hoboBottomTemperatureRange?.[0]}
+              title={item.title}
+              timeSeriesDataRanges={timeSeriesDataRanges}
+              timeZone={reef.timezone}
+              showRangeButtons={false}
```
Ry-DS, line 360: If we are using this component's  range buttons, is it just for the chart title? If so we might want to explore other means, but if this component offers more we can keep this implementation.

packages/website/src/common/Chart/ChartWithCard/index.tsx
```diff
@@ -325,6 +347,73 @@ const ChartWithCard = ({
...
+                  showDatePickers={false}
+                  oceanSenseData={item.data}
+                  oceanSenseDataUnit={item.unit}
+                  hideYAxisUnits
+                  displayHistoricalMonthlyMean={false}
+                />
+              </Grid>
+              {!pickerErrored && (
+                <Grid className={classes.card} item>
+                  <TempAnalysis
```
Ry-DS, line 395: This component was initially designed to hold one chart with a card, but now that we're expanding it like this we're making it pretty specialized. We could either rename and/or move this file to explain better what its doing now, or maybe have the single abstract <chartwithcard/> and use it multiple times (one for each sensor). What do you think? Feel free to take the path of least resistance.

Ry-DS: Keeping a ChartWithCard with a single chart would make more sense since right now we're duplicating the grid layout for the initial hobo/spotter chart, and it may only get worse with more chart types. See if its easy but otherwise we could probably get away with a rename.

packages/website/src/common/Chart/ChartWithTooltip.tsx
```diff
@@ -82,45 +86,59 @@ function ChartWithTooltip({
...
-        getHistoricalMonthlyMeanDataClosestToDate(
-          historicalMonthlyMeanData,
-          new Date(date)
-        )
-      )?.value || null;
+    const closestHistorical =
+      historicalMonthlyMeanData &&
+      getHistoricalMonthlyMeanDataClosestToDate(
+        historicalMonthlyMeanData,
+        new Date(date)
```
Ry-DS, line 93: there's a lot of `new Date(date)`

packages/website/src/common/Chart/utils.ts
```diff
@@ -572,7 +593,18 @@ export const createChartData = (
       },
       {
         label: "HOBO BOTTOM",
-        data: createGaps(hoboBottom, 1),
+        data: createGaps(hoboBottom, 24),
+        fill: false,
+        borderColor: "#f78c21",
+        borderWidth: 2,
+        pointBackgroundColor: "#ffffff",
```
Ry-DS, line 600: if this is powered by CSS can we just say 'white'? Or is only hex supported?

packages/website/src/common/Chart/utils.ts
```diff
@@ -572,7 +593,18 @@ export const createChartData = (
       },
       {
         label: "HOBO BOTTOM",
-        data: createGaps(hoboBottom, 1),
+        data: createGaps(hoboBottom, 24),
+        fill: false,
+        borderColor: "#f78c21",
```
Ry-DS, line 598: If this color is somewhere else in the app make sure to make a const for it, otherwise this is fine

packages/website/src/common/SiteDetails/OceanSenseMetrics/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { toRelativeTime } from "../../../helpers/dates";
+
+interface Metric {
+  label: string;
+  value: string;
+  measure: string;
+  icon: JSX.Element;
+}
+
+const metrics = (data?: OceanSenseData): Metric[] => [
```
Ry-DS, line 40: Think its possible to introduce some type check to make sure that this matches up with constructOceanSenseDatasets? So if one of them is missing a value, the other complains?

packages/website/src/common/SiteDetails/OceanSenseMetrics/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+      border: `1px solid ${grey[300]}`,
+      padding: theme.spacing(2),
+    },
+
+    triangle: {
+      width: 0,
+      height: 0,
+      borderStyle: "solid",
+      borderWidth: "40px 40px 0 0",
+      borderColor: "#c4c4c4 transparent transparent transparent",
```
Ry-DS, line 154: this color isn't available in theme grey pallete?

packages/website/src/constants/oceanSenseConfig.ts
```diff
@@ -0,0 +1,3 @@
+export const oceanSenseConfig: Record<number, string> = {
```
Ry-DS, line 1: comment here would be nice

packages/website/src/store/Reefs/helpers.ts
```diff
@@ -86,3 +90,25 @@ export const mapTimeSeriesDataRanges = (
...
+  response.data[key].map((value, index) => ({
+    value,
+    timestamp: response.timestamps[index],
+  }));
+
+export const mapOceanSenseData = (
+  response: OceanSenseDataResponse
+): OceanSenseData => {
+  return {
+    TEMP: mapOceanSenseMetric(response, "TEMP"),
```
Ry-DS, line 107: I'm starting to feel like we need an enum or something to help keep track of all these places where we mention the 6 metrics. If we need to add/remove one, it'll be somewhat hard for someone new to find all the places to change. Think we can strongly type all places with an enum so if the enum is edited, these places complain?

packages/website/src/store/Reefs/types.ts
```diff
@@ -163,6 +163,21 @@ export type TimeSeriesDataRangeResponse = Record<
 
 export type TimeSeriesDataRange = Record<Sources, TimeSeriesRange>;
 
+export type OceanSenseKeys = "TEMP" | "DO" | "EC" | "ORP" | "PH" | "PRESS";
```
Ry-DS, line 166: perfect, think we can incorporate this in constructOceanSenseDatasets and anywhere else we use the 6 sensors so if this changes, they complain?

Ry-DS (commented):
Nice! Pretty cool to have so many sensors now :eyes:
Let me know if you need any clarification :+1:

packages/website/src/common/Chart/ChartWithCard/index.tsx
tleraj: I renamed it to `Header`. What do you think?

tleraj:
@Ry-DS I changed the arrays to `Record<OceanSenseKeys, ...>`, and used `Object.values(constructOceanSenseData(data)).map(...)`. Now if we update the `OceanSenseKeys`, these functions complain.
Also I did some renaming so that the old components make more sense. Let me know what you think :slightly_smiling_face:

ericboucher:
Clicking the "download" button should also download the OceanSense data.

packages/website/src/common/Chart/ChartWithTooltip.tsx
```diff
@@ -71,56 +75,69 @@ function ChartWithTooltip({
...
-          historicalMonthlyMeanData,
-          new Date(date)
-        )
-      )?.value || null;
+    const closestHistorical =
+      historicalMonthlyMeanData &&
+      getHistoricalMonthlyMeanDataClosestToDate(
+        historicalMonthlyMeanData,
+        dateObject
+      )?.value;
```
ericboucher, line 96: Maybe we could create a small util to handle the getClosest to date and isNumberOrNull extract?
Eg.

```
getClosestValue(data, extractFunction, date, threshold) =>
  if (!data) return null
  const value = extractFunction(data, date, threshold)?.value
  return isNumber(value) ? value : null
```

tleraj: Good idea!! I tried to strongly type this util. Let me know what you think :slightly_smiling_face:

packages/website/src/common/Chart/ChartWithTooltip.tsx
```diff
@@ -18,13 +18,31 @@ import {
...
 } from "./utils";
+import { HistoricalMonthlyMeanData, SofarValue } from "../../store/Reefs/types";
 
 export interface ChartWithTooltipProps extends ChartProps {
   depth: number | null;
   className?: string;
   style?: CSSProperties;
 }
 
+const getClosestValue = <T,>(
```
Ry-DS, line 29: This name is somewhat misleading, are we doing anything related to getting the closest value in here?

Ry-DS (approved):
Nice! One nit-picky thing but otherwise lgtm!

packages/website/src/common/Chart/ChartWithTooltip.tsx
```diff
@@ -18,13 +18,31 @@ import {
...
+    maxHours?: number
+  ) => (T & { value: number }) | undefined,
+  date: Date,
+  data?: T[],
+  maxHours?: number
+): number | null => {
+  if (!data) {
+    return null;
+  }
+  const value = extractor(data, date, maxHours)?.value;
```
Ry-DS, line 42: Why are we passing variables into this function, then just passing them along to this function? Maybe we can make this function less complicated by taking the return type of this function as an arg, then doing the relevant `isNumber` logic after? For the `!data` check consider `data||[]` Then a good name would be `numberOrNull`

tleraj: Not sure that I get what you're suggesting :slightly_frowning_face:

Ry-DS: I made a commit with what I was thinking since its kinda complicated over text haha: https://github.com/aqualinkorg/aqualink-app/pull/565/commits/eebc6885020f4ff1bab968577c6e75d4c49cc7af

Let me know what you think! Its just another approach imo. Don't personally like the idea of passing args into a function to pass into the same function without much pre-processing (what was done before).

We could remove `numberOrNull` entirely if we refactor all the `getClosestXXX` functions to return null in the first place if it fails to find a number (make sure that these functions aren't used elsewhere if you consider this). Let me know what you think :)

Ry-DS: If you feel like this isn't a good approach feel free to revert the commit I linked

tleraj: It looks great. I just thought that the idea was to pass the data along with the appropriate function and wrap those in a function. Up to @ericboucher. :wink:

ericboucher:
@tleraj remember to add a quick comment in the PR description explaining how to link an oceansense sensor to a site.

ericboucher (approved):

PR #567, opened 2021-06-10
Add E2E tests including migrations

avalmas-programize:
Add e2e tests on the backend app
- Initialize an test instance of the app and save it in a singleton class to reuse it
- Mock firebase authentication process


.circleci/config.yml
```diff
@@ -9,6 +9,7 @@ jobs:
       - image: node:12
         environment:
           DATABASE_URL: postgres://postgres:CHANGE-ME@postgres:5432/postgres
+          TEST_DATABASE_URL: postgres://postgres:CHANGE-ME@postgres:5432/postgres
```
ericboucher, line 12: Why not simply use the DATABSE_URL we already have here?

packages/api/ormconfig.ts
```diff
@@ -13,16 +13,19 @@ try {
 const env = process.env.NODE_ENV || 'development';
 
 // If we have a DATABASE_URL, use that
-const connectionInfo = process.env.DATABASE_URL
-  ? { url: process.env.DATABASE_URL }
+const prefix = env === 'test' ? 'TEST_' : '';
+const connectionInfo = process.env[`${prefix}DATABASE_URL`]
```
ericboucher, line 17: one advantage of just relying on databse_url is that you won't need this extra logic

packages/api/src/collections/collections.service.ts
```diff
@@ -126,14 +126,10 @@ export class CollectionsService {
   }
 
   async update(collectionId: number, updateCollectionDto: UpdateCollectionDto) {
+    // No need to check if collectionId exists because the collection guard had made sure
```
ericboucher, line 129: had made sure -> ensures

packages/api/src/reef-applications/reef-applications.entity.ts
```diff
@@ -17,7 +17,6 @@ import { hashId } from '../utils/urls';
 @Entity()
 export class ReefApplication {
   @ApiProperty({ example: 1 })
-  @Exclude()
```
ericboucher, line 20: why this change?

.circleci/config.yml
avalmas-programize: The process of the test is the following:
- Cleans database
- Creates mock data
- Tests run

I didn't want to disturb any local setup we have by cleaning the database.
So by adding a test URL we isolate the testing to a specific database, different than the executing one.
In the circleCI there was no easy way to create a new database in PostgreSQL so I reused the `postgres` one.
However in my local I have added a `test_ovio` one.
Also if for some reason we test the staging environment, having overridden our local env with the staging env, and the tests run, it will result in deleting all contents of the staging database.

packages/api/ormconfig.ts
avalmas-programize: However that way there is no clear distinction between the testing and running processes, which can be problematic as mentioned above.

packages/api/src/collections/collections.service.ts
avalmas-programize: Fixed

packages/api/src/reef-applications/reef-applications.entity.ts
avalmas-programize: For some reason when fetching reef application the ids of all entities were missing (not just the `reef-application` ones, but also the `reef` and `user` ones)
Which was kinda odd. I tried to delete it, to understand why the id exclusion was cascading on the inner models, and it still seems to be excluded after deleting it (not sure why though).
I will revert this change for now since it didn't work. Good catch. Thanks

packages/api/ormconfig.ts
ericboucher: See my comment above. I'm fine keeping it like this but let's extract

`const databaseUrl = process.env[${prefix}DATABASE_URL]` to make things cleaner

.circleci/config.yml
ericboucher: Ok, another option would be to use another database on the same instance. A bit similar to what we do for Ovio (see `ovio-xplore-app/packages/api/scripts/prepare-test-db.sh`)

packages/api/ormconfig.ts
ericboucher: And add a small comment next to `// If we have a DATABASE_URL, use that` explaining why we add a prefix

.circleci/config.yml
avalmas-programize: yeah I saw that, when I was looking the ovio tests, and I think that just adding a new env variable seemed to me a much easier option (with of course having the user add the test database himself)

packages/api/ormconfig.ts
avalmas-programize: ok will do

ericboucher: @avalmas-programize can you fix this so that we can wrap this PR up? :)

packages/api/package.json
```diff
@@ -31,7 +31,7 @@
     "start:debug": "nest start --debug --watch",
     "start:prod": "node dist/src/main.js",
     "lint": "eslint \"{src,apps,libs,test,scripts}/**/*.{ts,js}\" --max-warnings 0 --fix",
-    "test": "jest",
+    "test": "NODE_ENV=test jest --config ./test/jest.json",
```
Ry-DS, line 34: use cross-env if its available in the project so this line also works on Windows/DOS

packages/api/src/auth/firebase-auth.utils.ts
```diff
@@ -0,0 +1,22 @@
+import admin from 'firebase-admin';
+
+const extractAuthHeaderAsBearerToken = (req: any): string | undefined => {
```
Ry-DS, line 3: Is it possible to to replace this with something stricter than any? Maybe we can make our own interface for this.

packages/api/src/auth/firebase-auth.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  const authHeader = req.headers.authorization;
+  const match = authHeader && authHeader.match(/bearer (.*)/i);
+  return match && match[1];
+};
+
+export const extractAndVerifyToken = async (
+  req: any,
```
Ry-DS, line 10: Same here (stricter type)

packages/api/src/auth/firebase-auth.utils.ts
Ry-DS: I found a Request type in `@types/express` you could probably use that. Try replace all the any use with this if possible :)

packages/api/src/collections/collections.service.ts
```diff
@@ -143,7 +139,7 @@ export class CollectionsService {
     } = updateCollectionDto;
 
     const filteredAddReefIds = addReefIds?.filter(
-      (reefId) => !collection.reefIds.includes(reefId),
+      (reefId) => !collection!.reefIds.includes(reefId),
```
Ry-DS, line 142: Since this is a filter, would it be safer to just use `?.`?

packages/api/src/collections/collections.service.ts
```diff
@@ -126,14 +126,10 @@ export class CollectionsService {
   }
 
   async update(collectionId: number, updateCollectionDto: UpdateCollectionDto) {
+    // No need to check if collectionId exists because the collection guard ensures
+    // that a collection with this collectionId exists
     const collection = await this.collectionRepository.findOne(collectionId);
```
Ry-DS, line 131: To help TS know this, you should make a type predicate https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates , or just keep the previous if statement so we don't have to use `!.`. If this function suddenly doesn't guarantee that collection is not null, TS will fail to alert us.

packages/api/src/collections/collections.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+      reefIds: [floridaReef.id || 0, californiaReef.id || 0],
+    };
+    updateCollectionDto = {
+      name: 'Updated Collection',
+      isPublic: false,
+      removeReefIds: [floridaReef.id || 0],
+    };
+  });
+
+  describe('Create a test collection', () => {
```
Ry-DS, line 69: `Create a test collection` isn't too accurate here, since we're doing more than creation in this test. Also, we could probably omit `test` since we know we're in a test when we're reading the logs. Maybe we could just call it `Collections: ` or similar.

packages/api/src/collections/collections.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+        .put(`/collections/${createdCollectionId}`)
+        .send({
+          addReefIds: [floridaReef || 0],
+        });
+
+      expect(rsp2.status).toBe(200);
+      expect(rsp2.body.reefIds.length).toBe(2);
+    });
+
+    it('GET /public filter public collections', async () => {
```
Ry-DS, line 149: We're cramming a lot of tests into a `collections.spec.ts` file. Can't we just make all of these different `describe`s? The file would be a lot neater.

packages/api/src/reef-pois/reef-pois.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+      imageUrl: 'http://some-sample-url.com',
+      latitude: 37.95813968719644,
+      longitude: 23.478834700473044,
+    };
+
+    createPoiWithoutLocationDto = {
+      name: 'Atlantis',
+      reefId: athensReef.id || 0,
+      imageUrl: 'http://some-sample-url.com',
+    };
```
Ry-DS, line 33: Could these constants possibly be outside at the top to make things neater?

packages/api/src/regions/regions.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const createParentRegionDto: CreateRegionDto = {
+    name: 'Europe',
+    polygon: createPoint(23.666694170726828, 37.92090950501416),
+  };
+
+  const createRegionDto: CreateRegionDto = {
+    name: 'Greece',
+    polygon: createPoint(24.666694170726828, 38.92090950501416),
+  };
```
Ry-DS, line 23: Constants outside at the top?

packages/api/src/sensors/sensors.service.ts
```diff
@@ -140,10 +131,6 @@ export class SensorsService {
 
     return Promise.all(
       surveyDetails.map(async (survey) => {
-        if (!survey.surveyMedia) {
-          return survey;
-        }
-
```
Ry-DS, line 146: Curious what changes are responsible for ensuring this and other fields can no longer be null :eyes:

packages/api/src/surveys/surveys.spec.ts
```diff
@@ -0,0 +1,355 @@
...
+import { Observations } from './survey-media.entity';
+import { floridaPoiOne } from '../../test/mock/poi.mock';
+
+export const surveyTests = () => {
+  const testService = TestService.getInstance();
+  let app: INestApplication;
+  const createSurveyDto: CreateSurveyDto = {
+    diveDate: new Date(),
+    weatherConditions: WeatherConditions.Stormy,
+    comments: 'No comments',
```
Ry-DS, line 27: Move constants outside this function to the top?

packages/api/src/time-series/time-series.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+} from '../../test/mock/time-series.mock';
+
+export const timeSeriesTests = () => {
+  const testService = TestService.getInstance();
+  let app: INestApplication;
+  let poiDataRange: [string, string] = [
+    new Date().toISOString(),
+    new Date(0).toISOString(),
+  ];
+  let reefDataRange: [string, string] = [
```
Ry-DS, line 22: make a type DataRange=[string, string]? Or possibly exists in the project already?

packages/api/test/app.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  describe('Region (e2e) /regions', regionTests);
+
+  describe('Reef Application {e2e) /reef-applications', reefApplicationTests);
+
+  describe('Survey (e2e) /reefs/:id/surveys', surveyTests);
+
+  describe('Reefs (e2e) /reefs', reefTests);
+
+  describe('Sensors (e2e) /sensors', sensorTests);
```
Ry-DS, line 45: Nice! I like how this is laid out, but I thought Jest has the ability to find tests by itself? What advantages you think this approach has, or is it not supported here?

packages/api/test/app.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+import { sensorTests } from '../src/sensors/sensors.spec';
+
+describe('AppController (e2e)', () => {
+  const testService = TestService.getInstance();
+  let app: INestApplication;
+
+  beforeAll(async () => {
+    // Initialize app
+    app = await testService.getApp();
+  });
```
Ry-DS, line 21: I notice this pattern often across all the tests, maybe we can share an app instance, or make a function that creates the app for us and takes 2 `before, after` functions to call on app?

Ry-DS (commented):
Nice! I left a few small comments but overall great!

packages/api/package.json
avalmas-programize: Done

packages/api/src/auth/firebase-auth.utils.ts
avalmas-programize: Done
Added express in dependencies

packages/api/src/auth/firebase-auth.utils.ts
avalmas-programize: Done

packages/api/src/collections/collections.spec.ts
avalmas-programize: I renamed it to `Test main collection flow` since that's what it is.

packages/api/src/collections/collections.spec.ts
avalmas-programize: The describe wasn't correct. This is intended to test the main collection flow, i.e. test all endpoint of collection controller in the average case

packages/api/src/reef-pois/reef-pois.spec.ts
avalmas-programize: Unfortunately no. The `id` column gets dynamically populated to all mock entities after they are saved to database (initialization phase).
Moving them outside the functions will result in having all ids set to 0 since those lines will execute before the initialization of the tests.

packages/api/src/regions/regions.spec.ts
avalmas-programize: Yes here we can move them. Done

packages/api/src/sensors/sensors.service.ts
avalmas-programize: The surveyMedia cannot be undefined.
The query on line 124 (current change) *inner joins* the `survey` table with the `surveyMedia` one. If a survey does not have any surveyMedia it will not be returned by the query, i.e. they will be no surveys with no surveyMedia fetched.
Also if a survey has surveyMedia but all of them are hidden, I also had added a where condition to filter out those surveys as well.

When I first implemented this I didn't pay much attention on the true result of the query and just respected the TS types. However with the tests I could see that this statement was unreachable and reevaluated my implementation.

packages/api/src/surveys/surveys.spec.ts
avalmas-programize: Yes thanks

packages/api/src/time-series/time-series.spec.ts
avalmas-programize: Done

packages/api/test/app.spec.ts
avalmas-programize: Yes this is an excellent note. However this will not work.
The `app` instance must be initialized at some point. Any code outside the `before*`, `after*` and `it` blocks will not accept any promise awaiting (describe blocks do not await promises).
So the first initialization of the app should take place in one of those `async-able` blocks. That means that *any* code outside those blocks (including calls to external functions to load tests from other files, the calculation of their arguments etc) will run before the initialization of the `app`.
So all test will get an undefined instance of the app or an unresolved Promise.

By having the initialization of the first beforeAll block you ensure that the app will initialize before all tests, and by grabbing the app instance of beforeAll of all the children tests you ensure that you will get the live app instance (that's why I went with a Singleton pattern for the TestService). Also by calling the same function for all `beforeAll` blocks and by having the app being a static instance, any test suite can be the initializer of the app. However due to execution order of the tests in jest this role is always appointed to `app.spec`

However this was not the first implementation that came to mind.
I initially tried the solution you propose.
After that I tried using `setupFilesAfterEnv` jest config option to create a script that will initialize the `app` and pass the instance as a jest global. However the initialization of the app for some reason was producing an error.
Finally I ended up in my current implementation with the Singleton TestService.

packages/api/test/app.spec.ts
avalmas-programize: First of all I need somehow to ensure that an app was initialized, along with a database and its mock entities.
If I let jest just find the tests by itself it would create a separate execution environment for each one and that means that I must initialize my app and database at each and every test suite (since the global thing didn't work).
This could cause many issues like database deadlocks (because many threads would simultaneously edit and read the database), connection issues or even race conditions. Also the test could not be so general because much functionality stretches across many controllers. 

By having all tests run under the same execution environment I can create many tests that depend on one another and also keep them fairly generalized.
However this way we lose some performance, but the test are quire fast (some few ms each one)

I cannot ensure this structure is the best way to create e2e tests while also doing integration testing and some functional testing.  But this took me way too long and I cannot imagine how much more would have taken me if I ended up trying to debug problems like the ones mentioned above

packages/api/src/collections/collections.service.ts
avalmas-programize: First I thought that the CollectionGuard ensures the existance of the collection entity. However this is not true when this function is called directly and not via the REST API endpoint. So I will revert my changes for this.

packages/api/src/collections/collections.service.ts
avalmas-programize: Reverted

packages/api/test/app.spec.ts
Ry-DS: thanks for going into detail :)
While it would be nice to simplify this, if it proves difficult we can keep as is since it works fine.

packages/api/test/app.spec.ts
Ry-DS: ah got it. This solution is already decent, and what I mentioned would only be a nice touch. Looks good for now :+1:

packages/api/src/sensors/sensors.service.ts
Ry-DS: It's a little annoying how TS doesn't reflect this change. If you can think of a way to tell TS that these fields are not going to be null, that would be great! Otherwise, we should tread carefully, since using `variable!` and other shortcuts can come back at us if these fields do start becoming `undefined` again - TS won't be able to warn us if it does happen.

packages/api/src/auth/firebase-auth.utils.ts
Ry-DS: Make sure its just the types in `devDependancies` :+1:

packages/api/ormconfig.ts
avalmas-programize: Done

packages/api/src/sensors/sensors.service.ts
avalmas-programize: I also do not like that we force TS to ignore the `undefined` type. However the query at this situation is making sure that we either have surveyMedia or we don't even have a survey.

ericboucher (approved):

Issue #571, opened 2021-06-16
Show video stream on reef page


ericboucher:
If a reef has a `videoStream` url, we should show it instead of the featured image. And make it 16/9 for a better video experience like this:

![Screen Shot 2021-06-16 at 9 54 13 AM](https://user-images.githubusercontent.com/16843267/122261556-61b49c80-ce89-11eb-8bdd-6360b6146dda.png)

PR #572, opened 2021-06-17
Video stream

avalmas-programize:
Related Issue: https://github.com/aqualinkorg/aqualink-app/issues/571

If stream is present it shows up in the reef info cards and it forces them to follow the 16:9 ratio
Stream auto-plays on load of the page (only for **desktop** devices)

Designs:
|---|---|---|
| desktop | ipad | iphone |


ericboucher:
Nice work. Let's remove the grey side round up the video instead.

Screen Shot 2021-06-17 at 2 14 04 PM

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
```diff
@@ -34,13 +33,13 @@ const FeaturedMedia = ({
...
       <Card className={classes.card}>
         <CardContent className={classes.content}>
-          <ReactPlayer
-            height="100%"
-            width="100%"
-            playing
-            muted
-            light
-            url={`${url}`}
+          <CardMedia
```
ericboucher, line 36: what's the rationale for this change? (just curious)

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
```diff
@@ -12,7 +12,6 @@ import {
   Theme,
 } from "@material-ui/core";
 import { Link } from "react-router-dom";
-import ReactPlayer from "react-player";
```
ericboucher, line 15: are we using react-player anywhere else? If not maybe we can get of it as a dependency

tleraj: We use it in the About and FAQ pages

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
tleraj: The goal is to use Material UI components as much as possible.

tleraj (approved):
Looks good!!

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
ericboucher: should we use CardMedia there as well then?

tleraj: I could try it yes 🙂

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -72,61 +73,118 @@ const SiteDetails = ({
...
+              isoDate: surveyDiveDate,
+              format: "MMM DD[,] YYYY",
+              displayTimezone: false,
+              timeZone: reef.timezone,
+            })}`,
+            variant: "subtitle2",
+            marginRight: 0,
+          },
+        ];
+      case !!videoStream:
```
ericboucher, line 101: This case should come first

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -72,61 +73,118 @@ const SiteDetails = ({
...
-    <Box mt="1rem">
-      {(!surveyDiveDate || !featuredSurveyPoint) && (
-        <CardTitle values={mapTitleItems} />
-      )}
+  const featuredMediaTitle = featuredMediaTitleItems();
 
-      <Grid container justify="space-between" spacing={2}>
-        <Grid item xs={12} md={6}>
-          {surveyDiveDate && featuredSurveyPoint && (
+  return (
```
ericboucher, line 116: This return statement suddenly got quite complicated. Can we try to clean it up a bit / make it a bit more future-proof?

ericboucher (approved):

Issue #575, opened 2021-06-28
Test video_stream links regularly and send an alert


ericboucher:
It is important to test our video_stream links and make sure that they are up.

We could have a simple firebase cloud function to go through all the streams and send an email or Slack alert when something is off.

ericboucher:
We should also test on the website using the YouTube api to check if the feed is playing.

* player.getPlayerState():Number
      *
      * -1 – unstarted
      *  0 – ended
      * 1 – playing
      *  2 – paused
      *  3 – buffering
      *  5 – video cued
      *
      */
      function checkIfPlaying(){
        var playerStatus = player.getPlayerState();

         return  playerStatus == 1;
      }

ericboucher:
@avalmas-programize let's implement this as well. You can ask @tleraj how to use the Youtube Data API to check.

What would be a smart way to send alerts? Slack?

PR #576, opened 2021-06-30
Check video stream before loading and other tweaks

tleraj:
The purpose of this PR is to close the following issues:
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/575.
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/580
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/581
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/583
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/591 

## Important note
In order to enable the whole live video checking functionality, you should enable the `YouTube Data API` (https://developers.google.com/youtube/v3) for your `REACT_APP_FIREBASE_API_KEY`.

### With featured image
| desktop | mobile |
|---|---|
| Screenshot_20210702_124743 | Screenshot_20210702_124821 |

### Without featured image
| desktop | mobile |
|---|---|
| Screenshot_20210702_125125 | Screenshot_20210702_125204 |

### Reef details on mobile

Screenshot_20210706_100957


ericboucher:
If the video is not live, we want to act as if video_strean was null @tleraj

tleraj:
So check if there is a survey image, and if that doesn't exist, fall back to the add survey image, right?

tleraj:
- Regarding the YouTube branding reduction, there is no `iframe` option to remove it at all. Also, since an `iframe` is necessary to load YouTube videos, there is no way to remove the logo with CSS. So the only logo that appears, is the one on the bottom right when we hover on the video.
 - Regarding the autoplay on mobile, according to the first answer [here](https://stackoverflow.com/questions/15090782/youtube-autoplay-not-working-on-mobile-devices-with-embedded-html5-player) and [here](https://stackoverflow.com/questions/3009888/autoplay-audio-files-on-an-ipad-with-html5/3056220#3056220), this option is not available. Also I think we don't want to allow autoplay on mobile, considering the mobile data consumption.

packages/website/src/common/NooaChip/index.tsx
```diff
@@ -0,0 +1,111 @@
+import React from "react";
```
ericboucher, line 1: The name of this component doesn't seem very appropriate if we also use it for "live video". no?

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
```diff
@@ -29,14 +30,26 @@ const FeaturedMedia = ({
...
+
   if (url) {
     return (
       <Card className={classes.card}>
         <CardContent className={classes.content}>
-          <CardMedia
-            src={`${url}?autoplay=1&controls=0&mute=1`}
-            className={classes.card}
-            component="iframe"
+          <YouTube
```
ericboucher, line 48: Is this able to read streams other than youtube? It is possible that in the future we will have non-youtube feed.

If not:
	1. is it easy to add these playerOpts while using the CardMedia?
	2. if it's not easy, let's keep it like this for now but keep in mind that we will most certainly have different types of feed In the future

packages/website/src/common/NooaChip/index.tsx
tleraj: It also covers these
Screenshot_20210702_184525
Maybe rename it to `Chip`?

ericboucher: yeah just Chip or FloatingChip maybe?
NOAA is just one out of 5 use cases so it just seems a bit odd to call it NoaaChip ;)

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
tleraj: I reused the `CardMedia` component and passed the options to the `iframe` src property :wink:

ericboucher (approved):

packages/website/src/common/Chart/ChartWithCard/DownloadCSVButton.tsx
```diff
@@ -118,8 +118,8 @@ function DownloadCSVButton({
 }
 
 DownloadCSVButton.defaultProps = {
-  startDate: "unknown",
-  endDate: "unknown",
+  startDate: undefined,
```
Ry-DS, line 121: Definitely wasn't great that these were invalid dates and passed into moment. We should make sure however the button still gives an accurate file name (I think it'll now be the current time instead of "unknown")

packages/website/src/common/Chip/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    chip: ({ width }: { width?: number }) => ({
+      backgroundColor: "#dddddd",
+      borderRadius: 8,
+      height: 24,
+      width: width || 60,
+      display: "flex",
+      [theme.breakpoints.between("md", "md")]: {
```
Ry-DS, line 20: Does breakpoints.only work?

packages/website/src/common/Chip/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+  makeStyles,
+  createStyles,
+} from "@material-ui/core";
+import { Link } from "react-router-dom";
+import { grey } from "@material-ui/core/colors";
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    chip: ({ width }: { width?: number }) => ({
+      backgroundColor: "#dddddd",
```
Ry-DS, line 15: Try get this colour from the theme grey palette

packages/website/src/common/Chip/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    }),
+    chipText: {
+      fontSize: 9,
+      color: grey[600],
+      [theme.breakpoints.between("md", "md")]: {
+        fontSize: 7,
+      },
+    },
+    circle: {
+      backgroundColor: "#51DD00",
```
Ry-DS, line 32: If this colour is nowhere else, would be nice if this was a constant variable at the top of this function

packages/website/src/common/Chip/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        <Link
+          to={to || { pathname: href }}
+          target={href ? "_blank" : undefined}
+          className={classes.link}
+        >
+          {live ? (
+            <>
+              <div className={classes.circle} />
+              <Typography className={classes.chipText}>
+                {liveText || "LIVE"}
```
Ry-DS, line 77: I prefer defaults specified in props instead of the Dom area. Up to you :)

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
```diff
@@ -34,8 +34,8 @@ const FeaturedMedia = ({
       <Card className={classes.card}>
         <CardContent className={classes.content}>
           <CardMedia
-            src={`${url}?autoplay=1&controls=0&mute=1`}
-            className={classes.card}
+            className={classes.fullHeightAndWidth}
+            src={`${url}?autoplay=1&mute=1&controls=0&modestbranding=1&fs=0&playsinline=1`}
```
Ry-DS, line 38: Is there some built-in function we can use to produce this URL from a plain object? Then we can try make this easier to read. If not we could keep as is (don't want to add packages here ;))

packages/website/src/helpers/video.ts
```diff
@@ -0,0 +1,3 @@
+import { last } from "lodash";
+
+export const getYouTubeVideoId = (url?: string | null) => last(url?.split("/"));
```
Ry-DS, line 3: This doesn't get the exact id since it's possible for a URL to have extra query options with `?`. Is this ok?
If the URL is user input we should definitely be striping out queries as they could be a security risk

packages/website/src/hooks/useLiveStreamCheck.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect, useState } from "react";
+import videoServices from "../services/videoServices";
+
+export const useLiveStreamCheck = (videoId?: string | null): boolean => {
+  const [isLive, setIsLive] = useState(false);
+
+  useEffect(() => {
```
Ry-DS, line 7: I'd prefer if we had one util with this hook and the extractor function for the video id to keep things less coupled ;)

packages/website/src/services/videoServices.ts
```diff
@@ -0,0 +1,21 @@
...
+    liveBroadcastContent: "live" | "none" | "upcoming";
+  };
+}
+
+export interface VideoQueryResponse {
+  items: VideoInfo[];
+}
+
+const getVideoInfo = (id: string) =>
+  youtubeApi.searchVideo(id).then((data) => data as VideoQueryResponse);
```
Ry-DS, line 19: You could specify the return type of this function or cast the whole statement with `Promise<VideoQueryResponse>` for better type safety (TS will be more likely to warn us if this library we're using introduces types and they are different to ours).

Ry-DS (commented):
Nice! A few suggestions let me know what you think 👍

packages/website/src/common/Chart/ChartWithCard/DownloadCSVButton.tsx
tleraj: Yes, `moment(undefined) = "today"`. So is this OK?

packages/website/src/common/Chip/index.tsx
tleraj: I used `green[300]` from MUI :wink:

packages/website/src/helpers/video.ts
tleraj: Here the URLs come from the database, where I assume they'll all be of the following format:
`https://www.youtube.com/embed/<VIDEO_ID>`. The URL's are not user input (at least for now :wink:)

packages/website/src/hooks/useLiveStreamCheck.ts
tleraj: Is this too much of a problem? Because I am a fan of such a modularization (hooks in a separate folder, helpers in a separate folder). Taking also into account that I moved all the other custom hooks into a separate folder. Let me know what you think :slightly_smiling_face:

packages/website/src/helpers/video.ts
Ry-DS: Nice. Still would be nice to just get the id in case a query param does sneak its way in there. Could help stop attackers from messing with queries we pass to youtube (and accidental issues if we give queries ourselves). What do you think?

packages/website/src/hooks/useLiveStreamCheck.ts
Ry-DS: Yupp that sounds good but since there's only one thing in each of these files I thought we could have one file with both. But I'm also fine with how its done now.
Completely up to personal preference, and since you know the project conventions more than me, your call :)

packages/website/src/common/Chart/ChartWithCard/DownloadCSVButton.tsx
Ry-DS: IMO its good, but ultimately up to you; I'm not too sure if my opinion is valid since I don't know what use-cases lead to an undefined date, but if you think it makes sense for the end user we can leave as is :+1:

packages/website/src/helpers/video.ts
tleraj: Looking better? :slightly_smiling_face:

packages/website/src/common/Chart/ChartWithCard/DownloadCSVButton.tsx
tleraj: The dates are `undefined` only at the component's initialization (`pickerStartDate = useState<string>(undefined)`). But they are always set to a date throughout the component's lifecycle, with default values `pickerEndDate = today` and `pickerStartDate = three months ago`. I think there's no need to worry about this too much.

PR #585, opened 2021-07-01
Add fallback to latest_data table for live data

avalmas-programize:
Fallback to latest_data table for `SST` and `SST_ANOMALY`, if live data from Sofar does not contain the `satelliteTemperature` metric, in order to avoid null values.

### DIFF
--- a/packages/api/src/reefs/reefs.service.ts
+++ b/packages/api/src/reefs/reefs.service.ts
@@ -39,6 +39,8 @@ import { createPoint } from '../utils/coordinates';
 import { Sources } from './sources.entity';
 import { getCollectionData } from '../utils/collections.utils';
 import { LatestData } from '../time-series/latest-data.entity';
+import { SourceType } from './schemas/source-type.enum';
+import { Metric } from '../time-series/metrics.entity';
 
 @Injectable()
 export class ReefsService {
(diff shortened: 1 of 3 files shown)


packages/api/src/reefs/reefs.service.ts
```diff
@@ -311,12 +313,35 @@ export class ReefsService {
...
+      : {};
+
     return {
+      ...satelliteTemperatureEntry,
       ...liveData,
-      sstAnomaly: getSstAnomaly(
-        reef.historicalMonthlyMean,
-        liveData.satelliteTemperature,
-      ),
+      sstAnomaly:
```
ericboucher, line 340: You could just use the satelliteTemperatureEntry. We redo a calculation that has been done elsewhere but this way you keep a very similar flow and only do one query to the DB. What do you think?

packages/api/src/time-series/latest-data.entity.ts
```diff
@@ -28,6 +28,8 @@ import { TimeSeries } from './time-series.entity';
     return connection
       .createQueryBuilder()
       .from(() => subQuery, 'time_series')
+      .addSelect('reef_id')
+      .addSelect('poi_id')
```
ericboucher, line 32: Why that change?

avalmas-programize: I just forgot to add them in the view declaration.
The migration does not generate the correct query (it has some issues with the DISTINCT part). so I write it by hand every time. So it was a bit misleading.

packages/api/src/reefs/reefs.service.ts
avalmas-programize: It's the same to me. Re-using the sst anomaly calculation will be cleaner however it, a calculation that we have already done. But I don't mind it.

ericboucher: let's just grab the sst data and use it in sst.

so let's re-order things a bit. Since you already await for the live data, let's check for liveData.satelliteTemperature before grabbing data from latestDataRepository

Then use the consolidated `satelliteTemperature` to get the sstAnomaly and replace satelliteTemperature in the final object.

Something like that?

```
return {
      ...liveData,
      sstAnomaly:
        getSstAnomaly(
          reef.historicalMonthlyMean,
          satelliteTemperature
        ),
      satelliteTemperature,
      weeklyAlertLevel: getMaxAlert(liveData.dailyAlertLevel, weeklyAlertLevel),
}
```

ericboucher (approved):

Issue #579, opened 2021-06-30
[dashboard] Scroll to map when clicking on reef on mobile


ericboucher:
When we are using the dashboard on mobile and click on a reef, a popup appears on the map but since it is out of view, it feels like nothing is happening.

Instead, we should also scroll the view to the top of the map so that the user can see the popup.

Issue #578, opened 2021-06-30
Extend title area in reef details view


ericboucher:
On desktop, there is room for the title to take more space and avoid having it on two lines. Let's make this area larger.

<img width="966" alt="Screen Shot 2021-06-30 at 5 03 43 PM" src="https://user-images.githubusercontent.com/16843267/123996534-461abd00-d984-11eb-830a-c28c484f5088.png">

Issue #577, opened 2021-06-30
Make "Add Survey" text clickable


ericboucher:
<img width="1469" alt="Screen Shot 2021-06-30 at 6 18 16 PM" src="https://user-images.githubusercontent.com/16843267/123996439-2edbcf80-d984-11eb-934a-c2444b95e0ca.png">

drewjgray:
The vertical alignment is slightly off - see screenshot attached. 
![Screen Shot 2021-07-06 at 5 12 53 AM](https://user-images.githubusercontent.com/6283065/124597915-e6576280-de18-11eb-91a1-a0ede417868b.png)

PR #586, opened 2021-07-02
Improve heat wave page and other UI Enhancements

tleraj:
The purpose of this PR is to close
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/579
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/577
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/578 
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/589
 - [x] https://github.com/aqualinkorg/aqualink-app/issues/590

# Designs

## Reef details header
|---|---|---|
| Screenshot_20210705_142509 | Screenshot_20210705_142455 | Screenshot_20210705_142437 |

## Heat wave tracker
|---|---|---|
| desktop | ipad | iphone |


packages/website/src/common/SiteDetails/Surveys/AddButton/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+const styles = () =>
+  createStyles({
+    addSurveyButton: {
+      color: grey[500],
+      "&:hover": {
+        color: grey[500],
+      },
+    },
+    addSurveyButtonIcon: {
+      fontSize: "32px !important",
```
Ry-DS, line 39: We generally want to avoid using `!important` since it can lead to annoying bugs in the future if someone tries to build on top of this. Is there a way you can think of to avoid this? Maybe passing a `classes` into Button? Or targeting the Typography?

packages/website/src/routes/HomeMap/Map/Markers/index.tsx
```diff
@@ -48,8 +48,11 @@ export const ReefMarkers = ({ collection }: ReefMarkersProps) => {
   // zoom in and center on reef marker when it's clicked
   useEffect(() => {
     if (map && reefOnMap?.polygon.type === "Point") {
+      const { clientHeight: mapContainerHeight } = map.getContainer();
       const [lng, lat] = reefOnMap.polygon.coordinates;
-      setCenter(map, [lat, lng], 6);
+      // Set map center a little above the reef center, so that it fits
+      // the popup.
+      setCenter(map, [lat + (3 * 180) / mapContainerHeight, lng], 6);
```
Ry-DS, line 55: Consider naming the constants 3 and 180 if they mean anything (for example, popup height). Otherwise, good as is.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
```diff
@@ -278,6 +270,16 @@ const styles = () =>
     button: {
       minWidth: 180,
     },
+    headerButtonWrapper: {
+      width: 48,
+    },
+    headerWrapper: {
+      maxWidth: "calc(100% - 48px)",
```
Ry-DS, line 277: Comments to explain why you need to do this and the same thing on line 280 would be nice :+1:

Ry-DS (commented):
Nice! Small comments but otherwise g2g!

packages/website/src/routes/HomeMap/Map/Markers/index.tsx
tleraj: I'll add a comment. I would like to elaborate on this a little bit, so that you can give me your thoughts:
This is an attempt to convert pixels to latitude units (&deg;).
Let's say that `mapContainerHeight` corresponds to 180&deg;. Then, by using the rule of three, a `dy` in pixels corresponds to:

`dLat = dy * 180 / mapContainerHeight`

If now for example we want to move the map center on reef selection 3px below, so that it fits the popup, with this formula we calculate how many latitude degrees we have to add to the reef center so that it fits the popup.
What do you think?

tleraj: A dummy way to test if this works, is by setting the reef name to the maximul possible length (100 chars) and check if it fits on all our maps for various screeen sizes.

drewjgray:
Screen Shot 2021-07-06 at 5 18 24 AM
Increase the font size of the "Tracking Heatwaves" title that is overlaid on the header image. It should be bigger than the other headings (see design screenshots).

tleraj:
@drewjgray Looking better? :slightly_smiling_face:

packages/website/src/routes/HomeMap/ReefTable/body.tsx
```diff
@@ -280,13 +284,15 @@ type ReefTableBodyIncomingProps = {
...
+  scrollTableOnSelection?: boolean;
+  scrollPageOnSelection?: boolean;
 };
 
 ReefTableBody.defaultProps = {
   isExtended: false,
   collection: undefined,
-  scrollOnSelection: true,
+  scrollTableOnSelection: true,
+  scrollPageOnSelection: undefined,
```
ericboucher, line 295: shouldn't this be a boolean as well?

packages/website/src/routes/Dashboard/Table/index.tsx
```diff
@@ -1,17 +1,21 @@
...
+
   return (
     <Box width="100%" mt="55px" mb="20px">
       <ReefTable
         showCard={false}
         showSpottersOnlySwitch={false}
         isExtended
         collection={collection}
-        scrollOnSelection={false}
+        scrollTableOnSelection={false}
```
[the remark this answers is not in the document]
tleraj: On the dashboard page, the dashboard does not have a fixed height. It gets bigger as we add reefs to our collection. This means that we don't have the option to scroll the table. As a result, `scrollOnTableSelection` here would scroll the whole page when we select a reef on the map, which I think is not preferable (in the case of a big collection where the table is really long).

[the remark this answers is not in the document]
ericboucher: Should we consider having a maxHeight and pagination? (not urgent, let's open an issue if we think it's a good idea)

ericboucher (approved):

ericboucher:
The popup looks a lot wider than it was on mobile. Is it expected @tleraj ?

Now:
Screen Shot 2021-07-07 at 7 30 51 PM

Prod:
Screen Shot 2021-07-07 at 7 31 01 PM

tleraj:
Yeah I did that on purpose, so that for long reef names , the tooltip header takes less lines, thus it has smaller height. If that's not wanted, I can revert it 😉

ericboucher:
Large like this is a bit odd as it overlaps under the buttons on the map. I'd say let's keep the same width as before for now.

PR #587, opened 2021-07-02
Extend backfill days for hourly SST updates

ericboucher:
Because NOAA only publishes data one or two days late, the hourly update needs to look back more than 24 hours to find data to update with more certainty.


avalmas-programize (approved):
Looks good to me

PR #592, opened 2021-07-06
Fix API url issue in upload media services

tleraj:
The purpose of this PR is to add an upload media hotfix.

### DIFF
--- a/packages/website/src/services/uploadServices.ts
+++ b/packages/website/src/services/uploadServices.ts
@@ -1,20 +1,17 @@
-import axios from "axios";
+import requests from "../helpers/requests";
 
 const uploadMedia = (
   formData: FormData,
   reefId: string,
   token?: string | null
 ) =>
-  axios({
+  requests.send<string>({
     method: "POST",
-    url: `${process.env.REACT_APP_API_BASE_URL}reefs/${reefId}/surveys/upload`,
+    url: `reefs/${reefId}/surveys/upload`,
     data: formData,
-    headers: {
-      "Content-Type": "multipart/form-data",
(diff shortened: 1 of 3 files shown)


packages/website/src/services/uploadServices.ts
```diff
@@ -1,20 +1,17 @@
...
   reefId: string,
   token?: string | null
 ) =>
-  axios({
+  requests.send<string>({
     method: "POST",
-    url: `${process.env.REACT_APP_API_BASE_URL}reefs/${reefId}/surveys/upload`,
+    url: `reefs/${reefId}/surveys/upload`,
     data: formData,
-    headers: {
```
ericboucher, line 12: So keeping the headers in here as override didn't work?

tleraj: The `requests.send(request: Request)` function accepts a request config, whose type is defined in `src/helpers/requests.ts` as
```ts
interface Request {
  method: AxiosRequestConfig["method"];
  url: AxiosRequestConfig["url"];
  data?: AxiosRequestConfig["data"];
  params?: AxiosRequestConfig["params"];
  token?: string | null;
  responseType?: AxiosRequestConfig["responseType"];
  cancelToken?: AxiosRequestConfig["cancelToken"];
  contentType?: string;
}
```
So this function does not come from `axios`, but is defined by us. This means that we couldn't pass `headers` as a param here (it would require to redefine the `send` function). With this approach, we just tweak the agent to accept a `contentType` param and set the `"Content-Type"` header. I think that this approach is a little better, as we pass only the `contentType` we want, rather than a massive `headers` object.

ericboucher (approved):

PR #597, opened 2021-07-07
Monitor reef status changes

avalmas-programize:
Add `AuditReefStatus` table schema:
Contains `old_status` (previous value), `new_status` (new value), `reef_id` (foreign key) and `sensor_id`.
Auto-generated `created_at` column is used to monitor the timestamp of the change.
Also `sensor_id` is optional, because during the approval procedure of the `reef-application` the `reef` might have `NULL` `sensor_id`.
Lastly I added indices on `old_status`, `new_status` and `created_at` that can be proven useful later on.

Create a trigger to fire on `UPDATE` events on the `reef` table and log the reef status changes.
The trigger does *not* monitor `UPDATE` events where `NEW.status` is equal to `OLD.status`

### DIFF
--- a/packages/api/migration/1625665541080-CreateSpotterStatusTrigger.ts
+++ b/packages/api/migration/1625665541080-CreateSpotterStatusTrigger.ts
@@ -1,0 +1,61 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CreateSpotterStatusTrigger1625665541080
+  implements MigrationInterface {
+  name = 'CreateSpotterStatusTrigger1625665541080';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "audit_reef_status" (
+        "id" SERIAL NOT NULL,
+        "old_status" character varying NOT NULL,
+        "new_status" character varying NOT NULL,
+        "sensor_id" character varying,
+        "created_at" TIMESTAMP NOT NULL DEFAULT now(),
+        "updated_at" TIMESTAMP NOT NULL DEFAULT now(),
+        "reef_id" integer NOT NULL,
+        CONSTRAINT "PK_77f6aa781b31d8e3f7c64072121" PRIMARY KEY ("id")
+      )`,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_39195166bb789a9236f1141c1b" ON "audit_reef_status" ("old_status") `,
+    );
+    await queryRunner.query(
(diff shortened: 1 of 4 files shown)


ericboucher:
@Zach I am wondering if we should also log sensor_id changes while we're at it. What do you think?

cc @avalmas-programize

ericboucher:
@avalmas-programize let's figure out a way to log all the changes on the reef table (not creation steps, only incremental changes)

avalmas-programize:
Changed the the table to be able to track all major columns change for the reef table. Namely the column that are tracked are:
```
  NAME = 'name',
  SENSOR_ID = 'sensor_id',
  STATUS = 'status',
  VIDEO_STREAM = 'video_stream',
  MAX_MONTHLY_MEAN = 'max_monthly_mean',
  APPROVED = 'approved',
```

The audit_reef table is working in the same principle as the time_series table. Each row contains a change event and it is indexed by `reef_id` (FK) and `column_name` (enum)

packages/api/src/audit/audit-reef.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+  NAME = 'name',
+  SENSOR_ID = 'sensor_id',
+  STATUS = 'status',
+  VIDEO_STREAM = 'video_stream',
+  MAX_MONTHLY_MEAN = 'max_monthly_mean',
+  APPROVED = 'approved',
+}
+
+@Entity()
+export class AuditReef {
```
ericboucher, line 22: I would actually call it ReefAudit to keep it next to the Reef table in alphabetical order

packages/api/src/audit/audit-reef.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+  columnName: ReefColumn;
+
+  @ManyToOne(() => Reef, { nullable: false, onDelete: 'CASCADE' })
+  reef: Reef;
+
+  @Index()
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @UpdateDateColumn()
```
ericboucher, line 44: Does it make sense to have an updatedAt column?

packages/api/src/audit/audit-reef.entity.ts
avalmas-programize: Done

packages/api/src/audit/audit-reef.entity.ts
avalmas-programize: It is really not important, I just added it for consistency.

ericboucher (approved):

PR #598, opened 2021-07-08
Add scheduled task to check video streams

avalmas-programize:
New firebase config variables:
- `slack.channel`: The channel id in which the alerts will be sent. Currently using the `dev-bots` channel.
- `slack.token`: The slack OAuth token of the `Video Streams Alerter` app.
  @ericboucher and @tleraj, you should also have access to the app configuration (https://api.slack.com/apps)
- `google.api_key`: The google api key for the youtube api. Currently using the firebase kay.
- `google.project_id`: The project id of each google cloud project (e.g. `aqualink-programize`). Used to provide a more distinguishable alert for each project. Also It can be used as a flag to mute the alerts for certain environments.
 
The current template used can be viewed in the slack `dev-bots` channel

Related issue: https://github.com/aqualinkorg/aqualink-app/issues/575


avalmas-programize:
Function deployed on programize and run successfully.

packages/api/src/workers/check-video-streams.ts
```diff
@@ -0,0 +1,218 @@
...
+    url: 'https://slack.com/api/chat.postMessage',
+    method: 'post',
+    data: payload,
+    headers: {
+      Authorization: `Bearer ${token}`,
+    },
+  });
+};
+
+export const checkVideoStreams = async (
```
ericboucher, line 105: @tleraj @avalmas-programize now that we have this functionality in the backend, should we check the video_stream in the backend and remove it from the "reef" object if it is not live? What do you think?

packages/api/cloud-functions/index.ts
```diff
@@ -147,3 +148,32 @@ exports.scheduledSSTTimeSeriesUpdate = functions
...
   });
+
+exports.scheduledVideoStreamsCheck = functions
+  .runWith({ timeoutSeconds: 540 })
+  .pubsub.schedule('0 0 * * *')
+  .timeZone('America/Los_Angeles')
+  .onRun(async () => {
+    const dbUrl = functions.config().database.url;
+    // eslint-disable-next-line fp/no-mutation
+    process.env.FIREBASE_KEY = functions.config().google.api_key;
```
ericboucher, line 159: Remember to update the command in the README that indicates how to set the environment variable.

We could probably improve how it's done as well, maybe with a command that uses the environment name to fetch the correct env file?

packages/api/cloud-functions/index.ts
```diff
@@ -147,3 +148,32 @@ exports.scheduledSSTTimeSeriesUpdate = functions
       conn.close();
     }
   });
+
+exports.scheduledVideoStreamsCheck = functions
+  .runWith({ timeoutSeconds: 540 })
+  .pubsub.schedule('0 0 * * *')
```
ericboucher, line 154: Explicitly state how often this will run for people not fluent in cron-job schedules ;)

# VideoStreamCheck will run daily at 12:00am

packages/api/cloud-functions/index.ts
```diff
@@ -147,3 +148,32 @@ exports.scheduledSSTTimeSeriesUpdate = functions
...
+  .timeZone('America/Los_Angeles')
+  .onRun(async () => {
+    const dbUrl = functions.config().database.url;
+    // eslint-disable-next-line fp/no-mutation
+    process.env.FIREBASE_KEY = functions.config().google.api_key;
+    // eslint-disable-next-line fp/no-mutation
+    process.env.SLACK_BOT_TOKEN = functions.config().slack.token;
+    // eslint-disable-next-line fp/no-mutation
```
ericboucher, line 162: It might be worth adding this eslint disable at the top of the file and explain why since we start to have quite a few iterations of it

packages/api/src/workers/check-video-streams.ts
```diff
@@ -0,0 +1,218 @@
...
+
+          return template;
+        })
+        .filter((messages) => messages),
+    ],
+  } as SlackMessage;
+
+  // No irregular video streams were found
+  // So skip sending an alert on slack
+  if (messageTemplate.blocks.length === 2) {
```
ericboucher, line 212: I see what you're trying to do here but it feels a bit odd. Maybe filter the `youTubeIds` array beforehand and return early if it's empty?

packages/api/src/workers/check-video-streams.ts
```diff
@@ -0,0 +1,218 @@
...
+    where: { videoStream: Not(IsNull()) },
+  });
+
+  // Extract the youTube id from the URLs
+  const youTubeIdToVideoStreamDetails = reefsWithStream.reduce<
+    reefIdToYouTubeIdMap
+  >((mapping, reef) => {
+    const id = getYouTubeVideoId(reef.videoStream!);
+
+    if (!id) {
```
ericboucher, line 140: Why are you treating this case differently? You could change the logic a tiny bit so that this is just another case and have a message `Video does not exist` in `youTubeIdToError`. This will also simplify your Slack text block and remove the need for the `(id in youTubeIdToError)` ternary

packages/api/package.json
```diff
@@ -21,9 +21,9 @@
...
-    "deploy:cloud-functions:programize": "yarn build:cloud-functions && firebase deploy -P aqualink-programize --only functions:pingService,functions:scheduledDailyUpdate,functions:scheduledSpotterTimeSeriesUpdate,functions:scheduledSSTTimeSeriesUpdate",
+    "deploy:cloud-functions": "yarn build:cloud-functions && firebase deploy --only functions:pingService,functions:scheduledDailyUpdate,functions:scheduledSpotterTimeSeriesUpdate,functions:scheduledSSTTimeSeriesUpdate,functions:scheduledVideoStreamsCheck",
+    "deploy:cloud-functions:staging": "yarn build:cloud-functions && firebase deploy -P aqualink-dev --only functions:pingService,functions:scheduledDailyUpdate,functions:scheduledSpotterTimeSeriesUpdate,functions:scheduledSSTTimeSeriesUpdate,functions:scheduledVideoStreamsCheck",
+    "deploy:cloud-functions:programize": "yarn build:cloud-functions && firebase deploy -P aqualink-programize --only functions:pingService,functions:scheduledDailyUpdate,functions:scheduledSpotterTimeSeriesUpdate,functions:scheduledSSTTimeSeriesUpdate,functions:scheduledVideoStreamsCheck",
```
Ry-DS, line 26: does `deploy:cloud-functions:xxx: yarn deploy:cloud-functions -P xxx` work? Since all these are basically the same except the P argument, and it looks like the P argument can be moved to the end.

packages/api/src/utils/urls.ts
```diff
@@ -8,3 +8,23 @@ export const hashId = (id: number) => hasher.encode(id);
...
+    .replace(/(>|<)/gi, '')
+    .split(/(vi\/|v=|\/v\/|youtu\.be\/|\/embed\/)/);
+
+  if (parsedUrl[2] !== undefined) {
+    const id = parsedUrl[2].split(/[^0-9a-z_-]/i);
+    return id[0];
+  }
+
+  return undefined;
+};
```
Ry-DS, line 30: Possible duplicate implementation of @tleraj 's in https://github.com/aqualinkorg/aqualink-app/blob/775fba177ad7b5bb910ab3631c32795d7a75b213/packages/website/src/helpers/video.ts

Is there a common package in this project to consolidate these two implementations? If not, might be best to try make them similar, then comment that this code is the exact same as whats in `website/xxx`

packages/api/cloud-functions/index.ts
avalmas-programize: Done

packages/api/src/workers/check-video-streams.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  if (item.liveStreamingDetails.actualEndTime) {
+    return 'The live stream has ended';
+  }
+
+  if (!item.liveStreamingDetails.actualStartTime) {
+    return 'The live stream has not started yet';
+  }
+
+  return '';
```
Ry-DS, line 83: returning undefined or null makes more sense here right? Then `getError() || ''`.

packages/api/cloud-functions/index.ts
avalmas-programize: Done

packages/api/src/workers/check-video-streams.ts
```diff
@@ -0,0 +1,218 @@
...
+            text: {
+              type: 'mrkdwn',
+              text: !(id in youTubeIdToError)
+                ? `*ReefId*: ${reefId}\n` +
+                  `*Video Stream URL*: ${url}\n` +
+                  `*Error*: Video does not exist
+                  `
+                : `*ReefId*: ${reefId}\n` +
+                  `*Video Stream URL*: ${url}\n` +
+                  `*Error*: ${youTubeIdToError[id]}`,
```
Ry-DS, line 200: These two strings are almost identical. Can't we put the ternary in the part that does change?

packages/api/src/workers/check-video-streams.ts
Ry-DS: agreed, you wouldn't expect a defensive return this far down. In the future one might put an undoable action above this without knowing of its existence.

Ry-DS (commented):
Nice! Small comments to improve quality but looks pretty good :+1:

packages/api/package.json
avalmas-programize: Unfortunately package.json doesn't support variable substitution. However we can do something similar with what we did with the `deploy:app`.
Create three commands for each environment, each command will export an env variable and call another generic function that will use the exported variable to do the necessary deployment.

packages/api/src/workers/check-video-streams.ts
avalmas-programize: If we end up assigning an empty string in the returning variable, why return `null` in the first place? Also the function returns the error found. I can assume that an empty string means an empty error and not have to add extra logic.

packages/api/src/workers/check-video-streams.ts
avalmas-programize: Yeah I changed that in my latest commit.

packages/api/src/workers/check-video-streams.ts
avalmas-programize: Yeah I could just filter the `youtubeIds`.  However now I should iterate both `youtubeIds` and `reefIdToVideoStreamDetails`. So basically it would be much more effort just to stop a bit earlier.

packages/api/src/workers/check-video-streams.ts
```diff
@@ -0,0 +1,224 @@
...
+          if (!reportedError) {
+            return undefined;
+          }
+
+          const template = {
+            type: 'section',
+            text: {
+              type: 'mrkdwn',
+              text:
+                `*ReefId*: ${reefId}\n` +
```
ericboucher, line 202: I'm being picky but can we use something a bit simpler like that maybe:

Reef: ReefName - ReefUrl (clickable)
Video: URL
Error: error

packages/api/package.json
Ry-DS: This isn't possible?
```json
"deploy:cloud-functions": "yarn build:cloud-functions && firebase deploy --only functions:pingService,functions:scheduledDailyUpdate,functions:scheduledSpotterTimeSeriesUpdate,functions:scheduledSSTTimeSeriesUpdate,functions:scheduledVideoStreamsCheck",
"deploy:cloud-functions:staging": "yarn deploy:cloud-functions -P aqualink-dev",
"deploy:cloud-functions:programize": "yarn deploy:cloud-functions -P aqualink-programize",
```
if not, I think its fine to explore your solution if its relatively simple and doesn't need much prior knowledge from new developers :+1:

Ry-DS: Ahh I see you went with your solution which looks great! Only downside I see is that it would probably not work on windows, but isn't a big issue since it could be run manually by a windows developer. If the solution I commented above also works, might be better to use that instead :+1:

packages/api/src/workers/check-video-streams.ts
Ry-DS: My thinking is that if the function is called `getError` it should return an error, or *nothing* if there is none. We could be lenient here since JS treats `''` effectively as `undefined` since they're both falsey. 
However if we were in a stricter language (like Java), and were looking for the existence of an error in the first place, it would be more complex and less intuitive to check for an empty string than for null/undefined. A developer would have to read this function to know how to use it, and if they didn't read it at all it could introduce a bug.

I'm probably overthinking this haha, but it would be nice to either rename the function to `getErrorMessage` so it's more clear it only returns a string, or return `null/undefined` if there is no error. Let me know what you think :)

packages/api/src/workers/check-video-streams.ts
```diff
@@ -0,0 +1,218 @@
...
+      },
+      ...youTubeIds
+        .map((id) => {
+          const { reefId, url } = youTubeIdToVideoStreamDetails[id];
+
+          if (id in youTubeIdToError && !youTubeIdToError[id]) {
+            return undefined;
+          }
+
+          const template = {
```
Ry-DS, line 197: Why make a const if its not used? Consider just using a return
```suggestion
          return {
```

packages/api/src/workers/check-video-streams.ts
avalmas-programize: As discussed I will remove the iteration from inside the object and add it to a intermediate variable to check if the array is empty rather than checking if all the blocks have size of 2.

packages/api/src/workers/check-video-streams.ts
avalmas-programize: Integrated with the other errors

packages/api/src/workers/check-video-streams.ts
avalmas-programize: Done

packages/api/src/workers/check-video-streams.ts
avalmas-programize: I changed the iteration from a map to reduce to avoid the filtering. So now the const is needed, in my opinion, because it feels cleaner

packages/api/src/workers/check-video-streams.ts
avalmas-programize: Ok I agree with changing to `getErrorMessage`.

packages/api/src/utils/urls.ts
avalmas-programize: The function I added is more specialized than the one @tleraj had added. However after talking with him we don't need that much specialization.

Unfortunately we don't have any common package, as far as I know. We will probably end up moving all the logic in the backend. But for now I will add a comment mentioning that it is a duplicate.

packages/api/package.json
avalmas-programize: Your solution also works and its cleaner than mine. I just didn't want to overthink it and reused the one we used for the app deployment.
However I will add a dedicated script for the production because having it as the default it assumes that the developer has selected the firebase production target as default.

packages/api/cloud-functions/index.ts
avalmas-programize: I created a new script based on the discussion here https://github.com/firebase/firebase-tools/issues/406#issuecomment-853992173
I added a new script in the package json
Updated README

packages/api/src/workers/check-video-streams.ts
avalmas-programize: I implemented the video stream checks in the GET /reefs/:reef_id endpoint.
If the video stream is not valid (it is not a live stream, cannot be found, hasn't started etc) it returns `null` to the front-end

tleraj (approved):
Looks good to me :slightly_smiling_face:

packages/api/src/workers/check-video-streams.ts
ericboucher: cc @tleraj

ericboucher (approved):

Issue #569, opened 2021-06-14
/map page performance improvements


ericboucher:
Perf follow up for #560 

In the front-end we are making a lot of unnecessary calls to the /reefs endpoint, because many of the links re-renders the map page instead of just changing the current component and reuse the state.

We should explore caching options in the front-end and backend.

Eg.

- caching the main `/reefs` requests locally and expiring it after a few hours
- add caching on the backend on queries that take a significant amount of time and either have a fixed expiration time or an invalidation policy (I am personally not a huge fan of putting a cache so close to the DB)

Issue #599, opened 2021-07-09
Ocean sense data makes app slow


tleraj:
Consider the following scenario:
 - Visit a reef details page with ocean sense data
 - Click the back button (we go to the map)
 - Select any other reef from the reef table
 
The app seems very slow, like it's not responding. The main cause of this is that we carry the ocean sense data in redux store.
We should consider:
 - Delete the data when they're not needed
 - Not storing them on redux at all (just display them when we visit the reef details page)

PR #601, opened 2021-07-12
Improve map page performance

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/569, https://github.com/aqualinkorg/aqualink-app/issues/599 and https://github.com/aqualinkorg/aqualink-app/issues/606.

Improvements:
 - [x] Use `to` instead of `href` on side menu buttons
 - [x] Make reefs list request on map page only when `!reefsList`
 - [x] Make selected reef request for card only when `reefOnMap.id !== selectedReefId` (useful when going back to the map from the reef details page).
 - [x] Delete ocean sense data once we leave the reef details page

To be done:
 - [x] Investigate table not scrolling issue when coming to `/map` from `reefs/:reefId`.


packages/website/src/common/MenuDrawer/index.tsx
```diff
@@ -127,6 +125,12 @@ const styles = () =>
         color: "#000000",
```
Ry-DS, line 125: not related, but can't this just be `'black'`?

Ry-DS: Would appreciate if the other colour in this file is also either coming from the theme, or defined as a const with a name so we know what colour it is without opening a colour picker ;)

packages/website/src/routes/HomeMap/ReefTable/body.tsx
```diff
@@ -143,6 +145,15 @@ const ReefTableBody = ({
...
 
+  const scrollToRow = useCallback(
+    (element: HTMLElement) =>
+      setTimeout(
+        () => element.scrollIntoView({ block: "center", behavior: "smooth" }),
+        SCROLLT_TIMEOUT
+      ),
+    []
+  );
+
```
Ry-DS, line 156: this is only used once, and very simple, therefore imo its better to just copy this into the useEffect and delete this callback. What do you think?

packages/website/src/routes/HomeMap/index.tsx
```diff
@@ -75,18 +80,20 @@ const Homepage = ({ classes }: HomepageProps) => {
...
-    dispatch(reefsRequest());
-  }, [dispatch]);
+    if (!reefs) {
+      dispatch(reefsRequest());
+    }
+  }, [dispatch, reefs]);
 
   useEffect(() => {
-    if (!reefOnMap && initialReefId) {
+    if (!reefOnMap && initialReefId && `${reefOnCardId}` !== initialReefId) {
```
Ry-DS, line 89: do we need this template string check? Not too sure what you're trying to do here. Are we trying to avoid a case where they're both undefined?

Ry-DS (commented):
Overall looks great! Small comments, let me know what you think :)

packages/website/src/routes/HomeMap/ReefTable/body.tsx
tleraj: I just wanted the `useEffect` to look cleaner, but yeah that also works :slightly_smiling_face:

packages/website/src/routes/HomeMap/index.tsx
tleraj: Added some explanatory comments :wink:

ericboucher:
It looks good overall but I deployed on programize and the map tooltip is blinking at the very end after the focusing is finished. Can you check why this happens @tleraj ? It is not happening on prod.

ericboucher (changes requested):
Let's try and get rid of the "blink" after focus.

ericboucher (approved):

Issue #588, opened 2021-07-02
Handle long reef names


tleraj:
We have multiple places where we don't handle the case of a large reef name (e.g. map tooltip, reef name on reef detail page, etc).
Here are some breaking example:

 
![Screenshot_20210702_162559](https://user-images.githubusercontent.com/64922890/124281288-4072d200-db52-11eb-88d0-9ffc5b5350fb.png)

![Screenshot_20210702_162643](https://user-images.githubusercontent.com/64922890/124281355-4ec0ee00-db52-11eb-933c-b4db4fc08ed4.png)

We should handle these cases a little better.

PR #603, opened 2021-07-14
Handle long names

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/588.
The worst case scenario is that the reef name or any survey point is named with one 100 character long string, like `AAAA..A`.
The following screenshots display the above scenario.

### Reef Name
| popup | homepage table| search dropdown | user dropdown | selected reef card | reef details page header | new survey page| media upload page |
|---|---|---|---|---|---|---|---|
| Screenshot_20210714_184215 | Screenshot_20210714_184305 | Screenshot_20210714_184350 | Screenshot_20210714_184434 | Screenshot_20210714_184600 | Screenshot_20210714_184703 | Screenshot_20210714_184756 | Screenshot_20210714_184854 |

### Survey Point Name
| popup | featured media header | point select dropdown | edit name dialog | survey card | survey media header | survey point page info card | survey point page survey title |
|---|---|---|---|---|---|---|---|
| Screenshot_20210714_185101 | Screenshot_20210714_185135 | Screenshot_20210714_185222 | Screenshot_20210714_185300 | Screenshot_20210714_185334 | Screenshot_20210714_185429 | Screenshot_20210714_185517 | Screenshot_20210714_185604 |


packages/website/src/common/SiteDetails/Surveys/PointSelector/index.tsx
```diff
@@ -93,8 +94,15 @@ const PointSelector = ({
               fullWidth
               value={editPoiNameDraft}
               onChange={onChangePoiName}
-              error={!editPoiNameDraft}
-              helperText={!editPoiNameDraft ? "Cannot be empty" : ""}
+              error={errored}
+              helperText={
+                // eslint-disable-next-line no-nested-ternary
```
ericboucher, line 99: It might be worth creating a small getHelperText function instead of a nested ternary

packages/website/src/routes/HomeMap/Map/Markers/index.tsx
```diff
@@ -40,22 +40,21 @@ export const ReefMarkers = ({ collection }: ReefMarkersProps) => {
...
     []
   );
   // zoom in and center on reef marker when it's clicked
   useEffect(() => {
     if (map && reefOnMap?.polygon.type === "Point") {
-      const { clientHeight: mapContainerHeight } = map.getContainer();
       const [lng, lat] = reefOnMap.polygon.coordinates;
-      // Set map center a little above the reef center, by converting
-      // the desired pixels to latitude degrees, so that it fits the popup.
-      const deltaPixels = 5;
```
ericboucher, line 55: Why is this logic disappearing from here?

tleraj: This approach introduced the following bug on`/map`:
 - Go to https://aqualink.org/map
 - Select a reef from the table (e.g. `Fairhaven Park, Bellingham, WA`)
 - When the map centering is finished, zoom on another area of the map (doesn't have to be a reef, just focus on another area of the map)
 - Now select a different reef from the table (e.g. `Edmonds Underwater Park, Edmonds, WA`)
 - The map does not zoom on the new selected reef

I must say that this approach seems a little hacky.
With the new approach, we use `leaflet`'s built-in methods to achieve what we want (focus on the reef and fit the popup inside the map container). It's more stable.

packages/website/src/common/SiteDetails/Surveys/PointSelector/index.tsx
```diff
@@ -39,6 +39,7 @@ const PointSelector = ({
 }: PointSelectorProps) => {
   const [editDialogOpen, setEditDialogOpen] = useState(false);
   const [editPoi, setEditPoi] = useState<Pois>();
+  const errored = !editPoiNameDraft || editPoiNameDraft.length > 100;
```
Ry-DS, line 42: why is the length of the draft relative to the error status? A comment to explain or something more intuitive would be nice :+1:

packages/website/src/common/SiteDetails/Surveys/PointSelector/index.tsx
```diff
@@ -52,6 +53,17 @@ const PointSelector = ({
     }
   };
 
+  const getHelperText = () => {
+    switch (true) {
+      case !editPoiNameDraft:
+        return "Cannot be empty";
+      case editPoiNameDraft && editPoiNameDraft.length > 100:
```
Ry-DS, line 60: Ahh I see. Consider making 100 a constant at the top so its a little easier to understand at the first look.

packages/website/src/routes/HomeMap/Map/Popup/index.tsx
```diff
@@ -28,6 +28,8 @@ const Popup = ({ reef, classes, autoOpen }: PopupProps) => {
   const { map } = useLeaflet();
   const reefOnMap = useSelector(reefOnMapSelector);
   const popupRef = useRef<LeafletPopup>(null);
+  const { name, region } = getReefNameAndRegion(reef);
+  const isNameLong = name && name.length > 50;
```
Ry-DS, line 32: optional chaining?

packages/website/src/routes/HomeMap/Map/Popup/index.tsx
```diff
@@ -61,8 +63,12 @@ const Popup = ({ reef, classes, autoOpen }: PopupProps) => {
             content: classes.popupHeaderContent,
             subheader: classes.subheader,
           }}
-          title={getReefNameAndRegion(reef).name}
-          subheader={getReefNameAndRegion(reef).region}
+          title={
+            <span title={isNameLong && name ? name : undefined}>
+              {isNameLong ? `${name?.substring(0, 50)}...` : name}
```
Ry-DS, line 68: 50 should be a constant and referenced for both use cases

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
```diff
@@ -86,6 +94,26 @@ const ReefNavBar = ({
...
+          );
+          dispatch(
+            setReefs(
+              reefsList?.map((item) =>
+                item.id === reef.id && data.name
+                  ? { ...item, name: data.name }
+                  : item
+              )
+            )
+          );
```
Ry-DS, line 115: Redux and other global state libraries encourage you to try keep most logic in the slice itself. What do you think of making a new action like `setReefName` which takes the reef id, and the name of the reef? Up to personal preference :)

packages/website/src/routes/Surveys/New/MediaCard.tsx
```diff
@@ -45,6 +46,8 @@ const MediaCard = ({
   const [addPoiDialogOpen, setAddPoiDialogOpen] = useState<boolean>(false);
   const [newPoiName, setNewPoiName] = useState<string>("");
 
+  const errored = newPoiName.length > 100;
```
Ry-DS, line 49: we need all of these lengths (50, 100) stored somewhere as constants since we can start to easily lose track if we wish to change it in the future.

packages/website/src/routes/Surveys/New/MediaCard.tsx
```diff
@@ -72,6 +75,10 @@ const MediaCard = ({
                   placeholder="Survey Point"
                   value={newPoiName}
                   onChange={handleNewPoiNameChange}
+                  error={errored}
+                  helperText={
+                    errored ? "Must not exceed 100 characters" : undefined
```
Ry-DS, line 80: the 100 should also be reference a constant

packages/website/src/routes/Surveys/New/index.tsx
```diff
@@ -95,6 +98,9 @@ const styles = (theme: Theme) =>
       marginTop: "2rem",
       // overflow: "hidden",
```
Ry-DS, line 99: lets clean this up

packages/website/src/routes/Surveys/View/SurveyMediaDetails.tsx
```diff
@@ -85,71 +84,79 @@ const SurveyMediaDetails = ({
...
-                      <Grid item>
-                        <Typography variant="subtitle1">
-                          {`${videos} video`}
-                          {videos === 1 ? "" : "s"}
-                        </Typography>
+                </Grid>
+                <Grid item>
+                  <Grid container alignItems="center" spacing={2}>
+                    <Grid item>
+                      <Grid container alignItems="center" spacing={1}>
```
Ry-DS, line 132: This is a lot of grid nesting. Do we need this much nesting, or could it be simplified with css rules? From what I see here, there's only one grid item which doesn't justify the use of a grid imo.

Ry-DS (commented):
Nice! Most comments are around using constants more (so we can easily change the max length from 100 if needed)
Also left a comment about making a more specialized redux action rather than just `setReefs` (its easier to scale redux if the logic is kept in the slice compared to in the UI code)
Let me know what you think :+1:

packages/website/src/routes/HomeMap/Map/Popup/index.tsx
tleraj: Actually, the correct approach is
```ts
const isNameLong = name?.length && name.length > NAME_MAX_LENGTH;
```
since the length property is the one that might be `undefined`.

packages/website/src/routes/ReefRoutes/Reef/ReefInfo/index.tsx
tleraj: That sounds really good!! I will adopt this :wink:

packages/website/src/routes/Surveys/View/SurveyMediaDetails.tsx
tleraj: Actually, I took the chance to refactor this component a little bit. I broke it doen into three distinct components:

 - `MediaPointName`: `Selected Point: <point_name>` header part
 - `MediaCount`: `X images Y videos` header part
 - `SliderCard`: Card with image and comments

Reduced the `Grid` nesting as you proposed by using plain `div` and `Box` with CSS :wink:

packages/website/src/routes/HomeMap/Map/Popup/index.tsx
Ry-DS: ahh I see, all good then :+1:

ericboucher (approved):

packages/website/src/routes/Surveys/View/MediaDetails/SliderCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+      },
+    },
+    cardImage: {
+      width: "100%",
+      height: "100%",
+    },
+    mediaInfoWrapper: {
+      height: "100%",
+      overflowY: "auto",
+      padding: "1rem 1rem 1rem 1.5rem",
```
Ry-DS, line 127: Consider theme.spacing here

Ry-DS (approved):
Looks great! Very small suggestion to use theme.spacing but could be ignored since it isn't that important

PR #604, opened 2021-07-16
Make netcdf optional

avalmas-programize:
NetCDF is listed as optional dependency
This means that the app will install successfully even if it doesn't install successfully (i.e. `libnetcdf` package is missing)


packages/api/scripts/utils/netcdf.ts
```diff
@@ -2,7 +2,14 @@ import { times } from 'lodash';
 import moment from 'moment';
 import { Extent, pointToIndex } from '../../src/utils/coordinates';
 
-const netcdf4 = require('netcdf4');
+let netcdf4;
+try {
+  // eslint-disable-next-line global-require, fp/no-mutation
+  netcdf4 = require('netcdf4');
+} catch {
+  console.error('Netcdf is not downloaded');
```
ericboucher, line 10: NetCDF is not installed. Please install NetCDF before continuing.

ericboucher (approved):

packages/api/scripts/utils/netcdf.ts
avalmas-programize: Done

PR #607, opened 2021-07-28
Show full screen option on videos

ericboucher:


ericboucher:
@tleraj actually this doesn't seem to be enough on desktop and we need to enable fs at the iframe level. I found this - https://stackoverflow.com/questions/52766216/cannot-make-the-iframe-fullscreen but it feels like overkill. Maybe we should use another component than CardMedia? What do you think?

tleraj (approved):
Looks good!!

ericboucher:
@avalmas-programize looks like the changes in #604 are breaking the linter. Could you have a quick look?

Issue #608, opened 2021-08-02
Relax constraints in sources table


ericboucher:
The hourly spotter updates are running into errors because spotter ids have been updated for some reefs. I think we should relax the limit for spotters at least to avoid these errors.

PR #611, opened 2021-08-02
Add sensor id on partial constraint

avalmas-programize:
Related issue https://github.com/aqualinkorg/aqualink-app/issues/608

### DIFF
--- a/packages/api/migration/1627912428897-UpdateSourceConstraint.ts
+++ b/packages/api/migration/1627912428897-UpdateSourceConstraint.ts
@@ -7,6 +7,7 @@ export class UpdateSourceConstraint1627912428897 implements MigrationInterface {
     await queryRunner.query(
       `ALTER TABLE "sources" DROP CONSTRAINT "no_duplicate_sources"`,
     );
+    await queryRunner.query(`DROP INDEX "IDX_84204417a187000377a03dc532"`);
     await queryRunner.query(
       `CREATE UNIQUE INDEX "no_duplicate_sources" ON "sources" ("reef_id", COALESCE("poi_id", 0), "type", COALESCE("sensor_id", 'SPOT-IMPOSSIBLE'))`,
     );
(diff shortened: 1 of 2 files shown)


packages/api/src/reefs/sources.entity.ts
```diff
@@ -13,7 +13,10 @@ import { SourceType } from './schemas/source-type.enum';
 
 @Entity()
 @Unique('no_duplicate_sources', ['reef', 'poi', 'type'])
-@Index(['reef', 'type'], { unique: true, where: '"poi_id" IS NULL' })
+@Index(['reef', 'type', 'sensorId'], {
```
ericboucher, line 16: How about adding poi as well in the index and remove the `where null`?

avalmas-programize: This will not work. Also the solution I proposed won't work as well.
We need 3 indexes because the `NULL` value is always considered falsy and `NULL == NULL` is also false and so just merging everything into one index will result in not catching duplicate entries.
So we need
- Index for `reef, poi, type` poi IS NOT NULL (no reason to extend it to NULL values)
- Index for `reef, type` poi IS NULL AND sensor IS NULL
- Index for `reef, type, sensor` sensor IS NOT NULL

packages/api/migration/1627912428897-UpdateSourceConstraint.ts
```diff
@@ -0,0 +1,34 @@
...
+export class UpdateSourceConstraint1627912428897 implements MigrationInterface {
+  name = 'UpdateSourceConstraint1627912428897';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "IDX_84204417a187000377a03dc532"`);
+    await queryRunner.query(
+      `ALTER TABLE "sources" DROP CONSTRAINT "no_duplicate_sources"`,
+    );
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX "IDX_e67d06022c484ed9b3688160d8" ON "sources" ("reef_id", "type", "sensor_id") WHERE "poi_id" IS NULL AND sensor_id IS NOT NULL`,
```
ericboucher, line 12: Do we need the `poi_id IS NOT NULL` here?

avalmas-programize: Based on our logic `poi_id IS NULL` is not necessary to be added because having `sensor_id IS NOT NULL` ensures that the poi will be NULL. However a little bit of extra verbosity is always helpful.

Comment edited

ericboucher:
Looks much cleaner, thanks. Let's deploy and see how it goes @avalmas-programize 

Once it's in place, you can change the spotter id for a site and check the results after the next update has run

avalmas-programize:
Changed sensor_id of reef with id 1630 and everything seems to run as expected.

ericboucher (approved):

PR #612, opened 2021-08-03
Increase logging on check video streams script

avalmas-programize:
Add condition to check if slack responded with a successful request.
Log errors and warnings returned by slack, in case of an unsuccessful request.
Log message in stdout regardless to always have the info available.

### DIFF
--- a/packages/api/src/workers/check-video-streams.ts
+++ b/packages/api/src/workers/check-video-streams.ts
@@ -248,7 +248,7 @@ export const checkVideoStreams = async (
   // Slack returns { ok: false } if an error occurs
   if (!data.ok) {
     logger.error(
-      `Slack responded with an non-ok response with error '${data.error}' and warning '${data.warning}'`,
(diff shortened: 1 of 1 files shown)


packages/api/src/workers/check-video-streams.ts
```diff
@@ -231,6 +238,17 @@ export const checkVideoStreams = async (
...
+
   // Send an alert containing all irregular video stream along with the reason
-  await sendSlackMessage(messageTemplate, slackToken);
+  const resp = await sendSlackMessage(messageTemplate, slackToken);
+
+  const { data } = resp;
+  // Slack returns { ok: false } if an error occurs
+  if (!data.ok) {
+    logger.error(
+      `Slack responded with an non-ok response with error '${data.error}' and warning '${data.warning}'`,
```
ericboucher, line 251: -> Slack responded with a non-ok status. Error: ${data.error}. Warning: ${data.warning}.

avalmas-programize: Done

ericboucher (approved):

PR #613, opened 2021-08-03
Report firebase errors on slack

avalmas-programize:


packages/api/cloud-functions/index.ts
```diff
@@ -36,6 +36,32 @@ function hasProjectId(config): config is { projectId: string } {
...
+  };
+
+  Axios({
+    url: 'https://slack.com/api/chat.postMessage',
+    method: 'post',
+    data: payload,
+    headers: {
+      Authorization: `Bearer ${token}`,
+    },
+  });
```
ericboucher, line 62: let's create a small util for this method since we are already using it for the other bot function

avalmas-programize: Done

ericboucher (approved):

Issue #467, opened 2021-02-17
Consider caching temperature data on the frontend


ericboucher:
When a user is changing dates on a granular chart (reef spotter data or survey point hobo data), we would like to avoid re-downloading data from the backend if we don't need to.

Eg. if the user was looking at the data for week 1 to 7 but then want to change to look at weeks 4 to 8, we should only need to download the data for week 8.

PR #610, opened 2021-08-02
Generate google analytics

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/479.
We generate analytics for:
 - `/` page button clicks
 - `/map` page `"EXPLORE"` button clicks
 - side menu button clicks

Deployed [here](https://aqualink-app-programize.web.app/map).
Analytics can be viewed [here](https://analytics.google.com/analytics/web/#/report-home/a163899403w281402098p248240479) (access needed).

PR #614, opened 2021-08-04
Reduce unnecessary requests

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/467.
We introduce two new store variables:
 - `timeSeriesMinRequestDate`: The earliest date the user has requested data for
 - `timeSeriesMaxRequestDate`: The most recent date the user has requested data for

We can now distinguish four distinct cases:
 - If `requestStart < timeSeriesMinRequestDate`, we request data for the interval `[requestStart, timeSeriesMinRequestDate]` and attach the resulting data to the already existing data.
 - If `requestEnd > timeSeriesMaxRequestDate`, we request data for the interval `[timeSeriesMaxRequestDate, requestEnd]` and attach the resulting data to the already existing data.
 - If `timeSeriesMinRequestDate < requestStart < requestEnd < timeSeriesMaxRequestDate`, we just return the already existing data.
 - In any other case, we make a request for the interval `[requestStart, requestEnd]`.

One important note is that only on of `requestStart` and `requestEnd` can change at a time via the date pickers.

Deployed at https://aqualink-app-programize.web.app/, along with https://github.com/aqualinkorg/aqualink-app/pull/610.


packages/website/src/helpers/dates.ts
```diff
@@ -23,8 +23,10 @@ interface DisplayDateParams {
   timeZoneToDisplay?: string | null;
 }
 
-export const isBefore = (start: string, end: string) =>
-  new Date(start).getTime() <= new Date(end).getTime();
+export const isBefore = (start: string, end: string, strict?: boolean) =>
```
Ry-DS, line 26: Not sure if strict is a good name here. What about `includeEndDate` or similar?

packages/website/src/store/Reefs/helpers.ts
```diff
@@ -115,3 +120,145 @@ export const mapOceanSenseData = (
...
+  (Object.keys(previousData) as (keyof TimeSeries)[]).reduce<TimeSeries>(
+    (acum, key) => ({
+      ...acum,
+      [key]:
+        direction === "left"
+          ? [...newData[key], ...acum[key]]
+          : [...acum[key], ...newData[key]],
+    }),
+    previousData
+  );
```
Ry-DS, line 138: Could this be simplified by lodash's mapValues? Not too sure but it looks like we're remapping the values from previousData to new values while preserving the key

packages/website/src/store/Reefs/helpers.ts
```diff
@@ -115,3 +120,145 @@ export const mapOceanSenseData = (
...
+
+const attachTimeSeries = (
+  direction: "left" | "right",
+  newData: TimeSeriesData,
+  previousData: TimeSeriesData
+): TimeSeriesData =>
+  (Object.keys(previousData) as (keyof TimeSeriesData)[]).reduce<
```
Ry-DS, line 145: Same as: TODO add link

packages/website/src/store/Reefs/helpers.ts
```diff
@@ -115,3 +120,145 @@ export const mapOceanSenseData = (
...
+  @param storedStart - The earliest date the user has requested data for
+  @param storedEnd - The most recent date the user has requested data for
+*/
+export const timeSeriesRequest = async (
+  params: TimeSeriesDataRequestParams,
+  storedTimeSeries?: TimeSeriesData,
+  storedDailyData?: DailyData[],
+  storedStart?: string,
+  storedEnd?: string
+): Promise<[TimeSeriesData, DailyData[], string, string]> => {
```
Ry-DS, line 182: Think we can use this feature from TS? https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements (it's new in TS 4 I think but would be really nice to introduce this new version since it doesn't have any breaking changes, easy as updating pkg.json hopefully)

Ry-DS (commented):
Don't think I'm too qualified to give an in-depth review since I don't know how these systems work but in terms of code quality, things look great! Left a few small suggestions (lmk if you think they're possible to merge)

packages/website/src/store/Reefs/helpers.ts
Ry-DS: this will substantially help with documentation when using this function (the tuple with magically have names when used!!)

Ry-DS (commented):
Updated all needed dependancies so eslint co-operates with TS 4. Also includes some nice new rules which were also enforced.

packages/website/src/store/Reefs/helpers.ts
```diff
@@ -115,3 +120,130 @@ export const mapOceanSenseData = (
...
+  @param storedEnd - The most recent date the user has requested data for
+*/
+export const timeSeriesRequest = async (
+  params: TimeSeriesDataRequestParams,
+  storedTimeSeries?: TimeSeriesData,
+  storedDailyData?: DailyData[],
+  storedStart?: string,
+  storedEnd?: string
+  // TODO TEMPORARY LABELS HERE (UPDATE NEEDED)
+): Promise<[w: TimeSeriesData, x: DailyData[], y: string, z: string]> => {
```
Ry-DS, line 167: @tleraj this one is up to you :)

Ry-DS:
Note: CI in https://github.com/aqualinkorg/aqualink-app/pull/614/commits/2b5d6a6b3a8ccd41d65728ef955f30d5a1eecc20 should have failed (API isn't building). Concerning this wasn't caught and could cause problems in the future.

tleraj:
@Ry-DS Tuple names updated!! I also added the `--silent` flag on the `yarn test` script, so that we get rid of the annoying warnings. Let me know if you don't like that so that I undo it :wink:

Ry-DS:
> @Ry-DS Tuple names updated!! I also added the `--silent` flag on the `yarn test` script, so that we get rid of the annoying warnings. Let me know if you don't like that so that I undo it 😉

I'm on the side of keeping the errors coming since they don't fail the build. But only if they offer something constructive (something we can actually fix). If you think they're pointless (errors we can't really fix) we can disable them!

ericboucher (approved):

PR #617, opened 2021-08-05
Set mobile design breakpoint at 960px for `/map`

tleraj:
The purpose of this PR is to apply the mobile design of the `/map` page for screens < 960px.

Deployed [here](https://aqualink-app-programize.web.app/map).

Right now the designed is applied at < 600px which makes the iPad view a liitle squished.
Here's the difference of the previous and the current approach for iPad (768px):

| before | after |
|---|---|
| Screenshot_20210805_175012 | Screenshot_20210805_175033 |

### DIFF
--- a/packages/website/src/routes/HomeMap/ReefTable/index.tsx
+++ b/packages/website/src/routes/HomeMap/ReefTable/index.tsx
@@ -38,28 +38,32 @@ import { Collection } from "../../Dashboard/collection";
 
 const SMALL_HEIGHT = 720;
 
-const MOBILE_SELECT_MENU_ITEMS = Object.values(OrderKeys).reduce<ReactNode[]>(
-  (elements, val) => [
-    ...elements,
-    ...times(2, (i) => {
-      const itemOrder: Order = i % 2 === 0 ? "asc" : "desc";
-      return (
-        <MenuItem value={`${val}-${itemOrder}`} key={val + i}>
-          <Typography color="primary" variant="h4">
-            {getOrderKeysFriendlyString(val)}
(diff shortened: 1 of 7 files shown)


packages/website/src/routes/HomeMap/ReefTable/index.tsx
```diff
@@ -38,28 +38,32 @@ import { Collection } from "../../Dashboard/collection";
...
-              <ArrowUpward fontSize="small" />
-            )}
-          </Typography>
-        </MenuItem>
-      );
-    }),
-  ],
-  []
-);
+const DEFAULT_ITEMS = ["alert", "dhw", "locationName", "sst"];
```
Ry-DS, line 41: You could type this to ensure it only contains keys inside of OrderKeys `Array<keyof OrderKeys>` let me know if that works :+1:

Ry-DS (approved):
Great!! Very minor change (might not be possible) but otherwise all good :+1:

packages/website/src/routes/HomeMap/ReefTable/index.tsx
tleraj: That doesn't work :slightly_frowning_face:. I tried to find a more type-safe way to do this, but I didn't come up with anything too sophisticated.

Ry-DS: Did a quick commit for this let me know what you think :+1:

tleraj: Yep, that works :wink:

PR #618, opened 2021-08-09
Add onboarding doc and improve inline comments

avalmas-programize:
Add comments on all major script to increase verbosity and clarify selected solutions

### DIFF
--- a/packages/api/ONBOARDING.md
+++ b/packages/api/ONBOARDING.md
@@ -175,27 +175,7 @@ This approach however results in many performance issues since the size of the t
   - INDEX (metric, source, timestamp DESC): To help with the materialized view `latest_data` **\***
 - Create materialized view (`latest_data`) to speed up fetching the latest data for every metric.
 
-```sql
-CREATE MATERIALIZED VIEW "latest_data" AS
-    SELECT
-        "time_series"."id",
-        "time_series"."metric",
-        "time_series"."timestamp",
-        "time_series"."value",
-        "source"."type" AS "source",
-        "source"."reef_id" AS "reef_id",
-        "source"."poi_id" AS "poi_id"
-    FROM
-        (SELECT
-            DISTINCT ON ("metric", "source_id") metric AS "metric",
-            "id",
-            "timestamp",
-            "value",
-            "source_id"
-        FROM "time_series" "time_series"
-        ORDER BY "metric", "source_id", "timestamp" DESC) "time_series"
-    INNER JOIN "sources" "source" ON "source"."id" = "time_series"."source_id"
-```
+For the latest structure of the materialized view `latest_data` please see migration `1623058289647-AddWeeklyAlertMetric.ts`
 
 Query is broken into two parts the inner query uses `distinct on` and `order by` to fetch the latest row based on `timestamp` for each `metric` and `source_id`. The outer query joins the results from the inner query with the source table to have easy access to other tables and also have more source details about the data.\
 By breaking the query we can help the optimizer select the correct indices to optimize it.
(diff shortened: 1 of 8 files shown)


packages/api/ONBOARDING.md
```diff
@@ -0,0 +1,182 @@
...
+    - dailyAlert (derived from degreeHeatingDays, satelliteTemperature and reef's maxMonthlyMean
+    - weeklyAlert (calculated as the maximum dailyAlert of the last 7 days)
+
+**\*** If any wave data is missing it fallbacks to hindcast data
+
+Spotter data are discarded if the date-sensorId exists in ExclusionDates
+
+**Note**: Sofar API has a limited window of data stored so we can backfill only for a certain ammount in the past. As a result, we need to check if no data have been fetched to avoid NULL values being added to dailyData.
+
+**Warning**: This script is deprecated and will be discarded in the next steps as the project transitions in fully being dependent of the TimeSeries data format.
```
ericboucher, line 87: How far are we from this? cc @tleraj

packages/api/ONBOARDING.md
```diff
@@ -0,0 +1,182 @@
...
+- scheduledDailyUpdate: (2 options)
+    - `yarn daily-worker` (no backfill functionality)
+    - `yarn backfill-daily-data -d days-to-backfill [-r reefId1 reefId2 ...]`
+- scheduledSSTTimeSeriesUpdate
+    - `yarn backfill-sofar-time-series -t sst_backfill -d days-to-backfill [-r reefId1 reefId2 ...]`
+- scheduledSpotterTimeSeriesUpdate
+    - `yarn backfill-sofar-time-series -t spotter_backfill -d days-to-backfill [-r reefId1 reefId2 ...]`
+
+The rest of the scripts are used to either augment the models with missing data or perform backfills further back in the past than what sofar is capable of:
+- AugemntReef: Augment reef with values based on its geographical position (region, MaxMonthlyMean, timezone, HistoricalMonthlyMeanimums)
```
ericboucher, line 134: Typo - HistoricalMonthlyMeanimums

packages/api/ONBOARDING.md
```diff
@@ -0,0 +1,182 @@
...
+    FROM 
+        (SELECT
+            DISTINCT ON ("metric", "source_id") metric AS "metric",
+            "id",
+            "timestamp",
+            "value",
+            "source_id"
+        FROM "time_series" "time_series"
+        ORDER BY "metric", "source_id", "timestamp" DESC) "time_series"
+    INNER JOIN "sources" "source" ON "source"."id" = "time_series"."source_id"
```
ericboucher, line 175: Isn't this part of the migrations already?

packages/api/ONBOARDING.md
```diff
@@ -0,0 +1,182 @@
...
+```
+Query is broken into two parts the inner query uses `distinct on` and `order by` to fetch the latest row based on `timestamp` for each `metric` and `source_id`. The outer query joins the results from the inner query with the source table to have easy access to other tables and also have more source details about the data.\
+By breaking the query we can help the optimizer select the correct indices to optimize it.
+
+Since this is a materialized view, we need to reload any time we make any update to the source or time_series table. So at the end of all update scripts a refreshing query is run.
```
ericboucher, line 180: Maybe give an example of where we do this / what the command is?

packages/api/ONBOARDING.md
avalmas-programize: I don't see any typo there...

packages/api/ONBOARDING.md
avalmas-programize: Yes it is. I just wanted to add the SQL query for the materialized view to explain its structure in more detail.

packages/api/ONBOARDING.md
avalmas-programize: Done

packages/api/ONBOARDING.md
avalmas-programize: Regarding the backend dailyData are used for the following:
- findDailyData endpoint GET /reefs/:id/daily_data
- liveDailyData endpoint GET /reefs/:id/live_data : used to calculate the live weeklyAlert
- findSensorSurveys endpoint GET /sensors/:id/surveys : used to fetch the sst for the surveys
- findSurveys endpoint GET /reefs/:reef_id/surveys : used to fetch bottomTemperature and sst as fallback for the surveys temperature
- and of course in the dailyData script

The findSensorSurveys and findSurveys can be altered to be using the time_series table quite trivially. However in liveDailyData are we satisfied in just fetching the weeklyAlert from the time_series table, since the update is more frequent, i.e more "live", or do we need to keep a explicit calculation of the weeklyAlert based on the time_series data?

packages/api/ONBOARDING.md
ericboucher: Ok, it feels a bit overwhelming in here. I'd recommend adding a link to the migration instead :)

packages/api/ONBOARDING.md
ericboucher: Hmm indeed that's the name in the API, shouldn't it be `HistoricalMonthlyMeans` though?

packages/api/ONBOARDING.md
ericboucher: I guess it's about the same, we could probably just rely on the time_series table for the weeklyAlert

packages/api/ONBOARDING.md
avalmas-programize: Ok removed and added a reference to the latest migration file

packages/api/ONBOARDING.md
avalmas-programize: Yeah I think so. I will edit the script file as well.

packages/api/ONBOARDING.md
```diff
@@ -0,0 +1,211 @@
...
+  - Video hasn't been uploaded or processed yet (not available).
+  - Video is not embeddable.
+  - Video is not a live stream.
+  - Video is live stream but it hasn't started yet.
+  - Video is live stream but it ended.
+
+  Other checks:
+
+  - Video URL is not in the correct embeddable format
+  - Video does not exists
```
ericboucher, line 132: Too much details imho. You can probably get rid of the paragraph I highlighted.

packages/api/ONBOARDING.md
```diff
@@ -0,0 +1,211 @@
...
+- Finally, we also need to seed the database if we want to create more complex scenarios to test. All data used for seeding are located on the `mock` folder. Here we face yet another challenge. Many of our models require some foreign keys to exist. For example, creating a `survey` requires a `reef` entity. So we need to first create the reef entity, grab the `id` of the reef and add it to the `survey.reef_id` entry. Following this process step by step would have been a messy solution, as it would have required us to perform another initialization to our seeds to amend their foreign keys. Instead we reference the ids of all relations in the respective columns. After that we make sure that we add the entities in the correct order (dependencies first). As a result once the ids have been populated by TypeORM they will be populated in the foreign key columns as well.
+
+**Note**: We could have added the ids manually on all entities and reference their existing value. However TypeORM does not overwrite the value of an auto generated column, i.e. the `id` column . So this approach is not possible
```
ericboucher, line 211: This paragraph is interesting but I would rephrase it to focus on what a new developer needs to know. i.e.

If you want to do XYZ, look here and make sure that you follow these steps...

avalmas-programize: Yeah I think step by step instructions, would be better

packages/api/ONBOARDING.md
avalmas-programize: Done

ericboucher (approved):

packages/api/ONBOARDING.md
```diff
@@ -0,0 +1,227 @@
...
+In order to test our application we have created both `functional` and `e2e` tests. After all tests have run we also calculate our coverage. We aim for green coverage on all api endpoints and and all util functions that are tested. There is no need to try and fully cover all functions in the codebase as many of those rely on third party libraries, which makes it complicated and unnecessary to test.
+
+#### Functional
+Functional tests have suffix `.test.ts` and are used to test the util functions, such as `getMMM` or `calculateDegreeHeatingDays`.
+
+#### E2E tests
+E2E tests have suffix `.spec.ts` and are used to test the entirety of the api endpoints making sure that all cases are covered.
+
+##### Requirements
+- A new database (recommended name `test_ovio`). No need to run any migrations, the initialization procedure on the test script will cover that for you.
```
Ry-DS, line 188: is `test_ovio` valid here? Just wondering why since this project isn't related to ovio I think

Ry-DS (commented):
Nice!
I left some questions I think an outsider (such as myself with this API) would have coming into this project for the first time. Hoping we could elaborate on some/all of them if it's not too much work. Let me know what you think :+1:

packages/api/ONBOARDING.md
ericboucher: Agreed, it shouldn't be test_ovio

PR #619, opened 2021-08-10
Optimize circleci

avalmas-programize:
Break circleci CI/CD job into 4 flows:
- build_dependencies:
  Install dependencies, and saves them on a cache to be used on the other scripts
- test api:
  Initializes the db, runs the tests and the linter for the api package
  Added cache for the jest test to speed up their execution
- test web:
  Runs the tests and linter for the website package
- Build app:
  Build app and cloud-functions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,8 +70,17 @@ jobs:
       # Ensure that all migrations succeed
       - run: cd packages/api && yarn migration:run
 
+      - restore_cache:
+          keys:
+            - v1-test
+
       # Run tests
       - run: cd packages/api && yarn test
+
+      # Save test cache
+      - save_cache:
+          paths: packages/api/.jest-cache
+          key: v1-test
 
       # Run lint
       - run: cd packages/api && yarn lint
@@ -97,7 +106,6 @@ jobs:
     steps:
       - checkout
       - restore_dependencies
-      - install_psql
           
       # Ensure that a normal build succeeds
       - run: yarn build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,12 +2,12 @@
 #
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 
-version: 2
-jobs:
-  build:
+version: 2.1
+executors:
+  default:
     docker:
       - image: node:12
-        environment:
+        environment: 
           DATABASE_URL: postgres://postgres:CHANGE-ME@postgres:5432/postgres
           TEST_DATABASE_URL: postgres://postgres:CHANGE-ME@postgres:5432/postgres
 
@@ -18,17 +18,32 @@ jobs:
           POSTGRES_PASSWORD: "CHANGE-ME"
         name: postgres
 
-    working_directory: ~/repo
-
+commands:
+  install_psql:
+    description: Install psql
     steps:
-      - checkout
+      - run:
+          name: Install psql
+          command: apt-get update && apt-get install -y postgresql-client
 
+  restore_dependencies:
+    description: Restore project dependencies from cache
+    steps:
       # Download and cache dependencies
       - restore_cache:
           keys:
             - v3-dependencies-{{ checksum "yarn.lock" }}
             # fallback to using the latest cache if no exact match is found
             - v3-dependencies-
+
+jobs:
+  build_dependencies:
+    working_directory: ~/repo
+    executor: default
+
+    steps:
+      - checkout  
+      - restore_dependencies
 
       - run: yarn install --ignore-engines
       - run:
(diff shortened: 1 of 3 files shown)


ericboucher:
@Ry-DS anything else we could look into now that you are an expert?

.circleci/config.yml
```diff
@@ -43,22 +58,73 @@ jobs:
...
+            - build_dependencies
+
+      - test_web:
+          requires:
+            - build_dependencies
+
+      - build_app:
+          requires:
+            - test_api
+            - test_web
```
Ry-DS, line 130: `build_app` doesn't use any outputs from test_api and test_web so it would make more sense to run it after install_deps since then the whole CI process will finish faster.

.circleci/config.yml
```diff
@@ -18,18 +18,33 @@ jobs:
...
       # Download and cache dependencies
       - restore_cache:
           keys:
             - v3-dependencies-{{ checksum "yarn.lock" }}
             # fallback to using the latest cache if no exact match is found
             - v3-dependencies-
 
+jobs:
+  build_dependencies:
+    working_directory: ~/repo
```
Ry-DS, line 41: consider using yaml anchors here to avoid repeating the ~/repo tag: https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/

Example
```yml
references:
    working_directory: &working_directory ~/repo
jobs:
    job_name:
        working_directory: *working_directory
```

.circleci/config.yml
```diff
@@ -43,22 +58,73 @@ jobs:
...
+
       # Run tests
-      # This could also be run using lerna: `yarn test` in the main repo, but doing it
-      # explicitly like this separates out the tasks nicely in the UI.
       - run: cd packages/api && yarn test
+
+      # Save test cache
+      - save_cache:
+          paths: packages/api/.jest-cache
+          key: v1-test
```
Ry-DS, line 83: this cache will never update as the project evolves. Tying it to the yarn.lock would be a good way to ensure that it updates every once in awhile

.circleci/config.yml
```diff
@@ -43,22 +58,73 @@ jobs:
...
+    steps:
+      - checkout
+      - restore_dependencies
+      
+      # Run tests
       - run: cd packages/website && yarn test
 
-      - run: yarn lint:all
+      # Run lint
+      - run: cd packages/website && yarn lint
```
Ry-DS, line 100: It might make more sense to have a `lint` job which lints all packages instead of making a step per package. This will allow us to use lerna's lint command, which provides good output on a per-package basis if it fails.

.circleci/config.yml
```diff
@@ -43,22 +58,73 @@ jobs:
...
+
+  build_app:
+    working_directory: ~/repo
+    executor: default
 
+    steps:
+      - checkout
+      - restore_dependencies
```
Ry-DS, line 108: if these two commands are repeated every time, may make sense to combine them?

.circleci/config.yml
```diff
@@ -43,22 +58,73 @@ jobs:
...
       - run: cd packages/api && yarn migration:run
 
+      - restore_cache:
+          keys:
+            - v1-test
+
       # Run tests
-      # This could also be run using lerna: `yarn test` in the main repo, but doing it
-      # explicitly like this separates out the tasks nicely in the UI.
       - run: cd packages/api && yarn test
```
Ry-DS, line 78: We could possibly combine both jest tests into a single `tests` job which runs jest for both pkgs

Ry-DS (commented):
Looking good so far! Left a few suggestions to improve time and effectiveness.
I propose a new layout:
- install_deps
    - jest (runs tests for both api and website)
    - lint (lints both api and website)
    - build (builds both api and website)
    
Then there's only 2 layers instead of 3. If build is too slow for website and api, we could split it to `build_website` and `build_api`

.circleci/config.yml
avalmas-programize: I just didn't want to run a build that will crash because of an obvious linter error. That's why I separated the operations.
But I really don't have any strong objections

.circleci/config.yml
avalmas-programize: Done.

.circleci/config.yml
avalmas-programize: Added checkout command on restore_dependencies command

.circleci/config.yml
avalmas-programize: Done

.circleci/config.yml
avalmas-programize: I run the tests together with `lerna` and the output is a bit overwhelming, because it is indented to include the package tag.
So I think running them separately will be better for the engineer that will try to figure out what went wrong on the tests.

Ry-DS: Yup I agree, just meant to have a `tests` job which runs each package test separately instead of having build_api and build_website

.circleci/config.yml
Ry-DS: Just meant to do something like `v1-tests-(checksum of yarn.lock)` but really like the approach you have here!

Ry-DS (approved):
Looks great!! Works like a charm.
Only suggestion is to use `--frozen-lockfile` from yarn instead of having a dedicated step. Otherwise, g2g.
https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-frozen-lockfile

PR #621, opened 2021-08-12
Retire daily data

avalmas-programize:
Remove daily data usage from the backend
- findDailyData endpoint GET /reefs/:id/daily_data
- liveDailyData endpoint GET /reefs/:id/live_data : used to calculate the live weeklyAlert
- findSensorSurveys endpoint GET /sensors/:id/surveys : used to fetch the sst for the surveys
- findSurveys endpoint GET /reefs/:reef_id/surveys : used to fetch bottomTemperature and sst as fallback for the surveys temperature


avalmas-programize:
I think I am done with the backend only part. The only big thing that remains is the daily data endpoint.

avalmas-programize:
For both data endpoints hourly query parameter the `hourly` parameter was replaced by the `grouping`.
Grouping parameter can at the moment get two possible values: `hourly`, `daily` which indicate the type grouping the data will have. If `daily` is selected the timestamp of all data will be returned grouped at the end of each day.
Grouping parameter is optional. So if it is not specified it will returned the raw data.

Issue #437, opened 2021-01-26
Restrict non reef admins from accessing buoy application data


tleraj:
Right now, any signed up user has access to a reef's application. We need to change this and only allow reef managers to be able to access a reef's application.

avalmas-programize:
This is resolved by https://github.com/aqualinkorg/aqualink-app/pull/457
However we still have (for compatibility reasons) the old endpoint with which you can fetch any reef-application unrestricted (provided that you guess the hashed id or the uid correctly)
Anything else that should be done on this @ericboucher ?

ericboucher:
I think we can probably remove the old endpoints then. I'll check with the team that they are not used anymore but if you can prepare a quick PR to remove them that would be great!

PR #623, opened 2021-08-17
Delete deprecated reef-application endpoints

avalmas-programize:
Delete:
- GET /reef-applications/:id : Gets application based on a hashId and the uid
- PUT /reef-applications/:id : Update reefApplication based on a hashId

Also there was a functionality that allowed users to edit some reef column through the above PUT endpoint. This is no longer available and thus deleted.

### DIFF
--- a/packages/api/src/reef-applications/reef-application.spec.ts
+++ b/packages/api/src/reef-applications/reef-application.spec.ts
@@ -51,7 +51,7 @@ export const reefApplicationTests = () => {
     expect(rsp.status).toBe(404);
   });
 
-  it('PUT /:hashId/reefs/:id updates a reef application using reefId', async () => {
+  it('PUT /:appId/reefs/:id updates a reef application using reefId', async () => {
     const updatedReefApplicationFoundingSource = 'New found source';
     mockExtractAndVerifyToken(reefManagerFirebaseUserMock);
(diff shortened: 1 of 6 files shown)


ericboucher (approved):

ericboucher:
@tleraj out of caution, let's make sure that this doesn't break anything in the frontend

packages/api/src/reef-applications/reef-application.spec.ts
```diff
@@ -51,46 +51,6 @@ export const reefApplicationTests = () => {
...
-        'updatedAt',
-        'user',
-        'reef',
-        'uid',
-        'id',
-      ),
-    });
-  });
-
   it('PUT /:hashId/reefs/:id updates a reef application using reefId', async () => {
```
ericboucher, line 54: should we also delete this test? @avalmas-programize or @tleraj could you have another look at this spec file and see if other tests should be removed or updated?

tleraj: This endpoint is being used on the frontend, so I think we should keep it.

ericboucher: Then the question is, why is it being used in the frontend? There might be more frontend changes needed to fix #437

ericboucher (approved):

PR #626, opened 2021-09-07
Remove xlsx

ericboucher:


ericboucher:
Indeed... I thought it was used in the excel imports for the Stanford Palau data. But if it's not used anymore we can get rid of it.

PR #631, opened 2021-10-22
Switch from peakWavePeriod to meanWavePeriod

ericboucher:
Sofar has transitioned from using peak to using meanPeriod as the default for wave. And stopped reporting peakPeriod on wave models.

As such, we are making the same transition.

This PR aims to make the transition in the frontend and the backend and rename columns with more accurate names to separate these two values cleanly. A few things to do still:

- [x] rename wavePeriod to wavePeakPeriod
- [x] add new columns for waveMeanPeriod
- [x] add new metric for wave_mean_period in timeseries


packages/website/src/common/SiteDetails/Waves/index.tsx
```diff
@@ -194,10 +194,10 @@ const Waves = ({ liveData, classes }: WavesProps) => {
                 DIRECTION
               </Typography>
               <Grid container alignItems="baseline">
-                {waveDirection?.value && (
+                {waveMeanDirection?.value && (
                   <img
                     style={{
-                      transform: `rotate(${waveDirection?.value + 180}deg)`,
+                      transform: `rotate(${waveMeanDirection?.value + 180}deg)`,
```
tleraj, line 200: Maybe use classes with props in order to avoid inline styles?

packages/website/src/layout/App/index.tsx
```diff
@@ -58,6 +63,7 @@ function App() {
                 <Route exact path="/drones" component={Drones} />
                 <Route exact path="/faq" component={Faq} />
                 <Route exact path="/terms" component={Terms} />
+                <Redirect from="/reefs/:id" to="/sites/:id" />
```
tleraj, line 66: Why not `Route`?

tleraj (commented):
Looks good!! Just some minor comments

packages/website/src/layout/App/index.tsx
ericboucher: what do you mean?I actually want a redirect from /reefs/184 to /sites/184

tleraj: Actually this is my bad, because I didn't see the `<Route path="/sites" component={SiteRoutes} />` below. All good :slightly_smiling_face:

tleraj (approved):

PR #634, opened 2021-11-11
Fix negative wind and add lost spotter status

### LINKED ISSUE #625: Add new category for invasive species
Let's add a new category with

key: “invasive-species”
value: “sign of damage from invasive species”

ericboucher:
- Fix negative wind issue
- Add "lost" status for spotters
- Add invasive-species category (Fixes #625)


tleraj (approved):
Pretty standard changes. Let's merge :slightly_smiling_face:

PR #635, opened 2021-11-22
Allow development without file upload configuration (GCS Bucket)

ericboucher:
Allow development without file upload configuration

### DIFF
--- a/packages/api/src/uploads/file.decorator.ts
+++ b/packages/api/src/uploads/file.decorator.ts
@@ -56,6 +56,10 @@ export const AcceptFile = (
     ),
     acl: 'publicread',
   };
+  if (!config.bucket){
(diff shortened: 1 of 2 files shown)


packages/api/src/uploads/file.decorator.ts
```diff
@@ -56,6 +56,10 @@ export const AcceptFile = (
     ),
     acl: 'publicread',
   };
+  if (!config.bucket){
```
tleraj, line 59:
```suggestion
  if (!config.bucket) {
```

tleraj (commented):
I think that should do it :wink:

tleraj (approved):
Didn't know there was a `NOT_IMPLEMENTED` status :eyes:. Looks good :slightly_smiling_face:

PR #636, opened 2021-11-26
Add sonde metrics and create sonde data import file

ericboucher:


packages/api/src/time-series/metrics.entity.ts
```diff
@@ -34,9 +52,9 @@ export class Metrics {
...
 
   @ApiProperty({ example: 'Metric Description' })
-  @Column({ nullable: false })
+  @Column({ nullable: true })
   description: string;
 
-  @Column({ type: 'enum', enum: Units, nullable: false })
+  @Column({ type: 'enum', enum: Units, nullable: true })
   units: Units;
 }
```
tleraj, line 60: Why do we allow `null` values for this column? Shouldn't we extend the `Units` enum so that it has units for the new sonde metrics? If they're not trivial then we can leave as is :slightly_smiling_face:

packages/api/src/utils/uploads/upload-sonde-data.ts
```diff
@@ -0,0 +1,201 @@
...
+      const newKey = newKeys[key] || key;
+      return { [newKey]: obj[key] };
+    }
+    return {};
+  });
+  return Object.assign({}, ...keyValues);
+}
+
+function ExcelDateToJSDate(dateSerial: number) {
+  return new Date(Math.round((dateSerial - 25569) * 86400 * 1000));
```
tleraj, line 55: Let's assign these numbers to constants at the top. I get that 86400 are the seconds in a day, but what about 25569? :stuck_out_tongue:

packages/website/src/common/Chart/MultipleSensorsCharts/AnalysisCard.tsx
```diff
@@ -246,11 +246,11 @@ const styles = (theme: Theme) =>
       width: "auto",
     },
     lightFont: {
-      fontWeight: 300,
+      fontWeight: 250,
```
tleraj, line 249: `fontWeight` does not accept any number as a value. It only accepts 100, 200, 300, ...900. I'm assuming that this change did not cause any change in the UI. It won't show an error though.

packages/website/src/services/siteServices.ts
```diff
@@ -48,13 +48,18 @@ const getSiteTimeSeriesData = ({
   end,
   metrics,
   hourly,
-}: TimeSeriesDataRequestParams) =>
-  requests.send<TimeSeriesDataResponse>({
+}: TimeSeriesDataRequestParams) => {
+  // TODO - Move the site condition and adapt to prod.
+  const filterMetrics = metrics.length > 0 && siteId !== "1637";
```
tleraj, line 53: What do you think about adding this `siteId` into `.env`? Not necessary, just to keeps things visible

tleraj (commented):
Looks quite good. Just some comments.

packages/api/src/time-series/metrics.entity.ts
ericboucher: Some data doesn't have units, like a count or an alert level for example

packages/website/src/services/siteServices.ts
ericboucher: This is just temporary, ultimately we will have a config at the site level

packages/website/src/common/Chart/MultipleSensorsCharts/AnalysisCard.tsx
ericboucher: Are you sure?

In earlier versions of the font-weight specification, the property accepts only keyword values and the numeric values 100, 200, 300, 400, 500, 600, 700, 800, and 900; non-variable fonts can only really make use of these set values, although fine-grained values (e.g. 451) will be translated to one of these values for non-variable fonts using the Fallback weights system.

packages/api/src/utils/uploads/upload-sonde-data.ts
ericboucher: It's an Excel constant 🤷 
I'll see if I can make it a bit clearer what's happening

packages/api/src/utils/uploads/upload-sonde-data.ts
```diff
@@ -0,0 +1,196 @@
...
+interface Repositories {
+  siteRepository: Repository<Site>;
+  surveyPointRepository: Repository<SiteSurveyPoint>;
+  timeSeriesRepository: Repository<TimeSeries>;
+  sourcesRepository: Repository<Sources>;
+}
+
+const logger = new Logger('ParseSondeData');
+
+const metricsMapping: { [name: string]: Metric } = {
```
tleraj, line 23: I personally prefer `Record<string, Metric>` instead of `{ [name: string]: Metric }`, but totally up to you :slightly_smiling_face:.

tleraj (commented):
Looks good!! Just a minor comment, but it's good to go :slightly_smiling_face:

tleraj (approved):
:ok_hand:

PR #643, opened 2021-12-15
Flip wave directions

ericboucher:
Following a conversation with Sofar, we realized that there had been a misunderstanding on conventions used to represent wave directions.

We are switching back to using the meteorological convention for all metrics.


tleraj (approved):
:ok_hand:

PR #662, opened 2022-01-19
Fix issue with multiple source_id for latest_data

ericboucher:
If a site had multiple sources with the same type (eg. a spotter was replaced), we ended up with two data points for a few metrics.

This new MV query fixes it but is a bit slower. We should investigate how to speed things up eventually.


tleraj (approved):
:ok_hand:

PR #664, opened 2022-01-26
Faster hasSondeData check

### LINKED ISSUE #661: Heat stress alert level box is slow to load
Because of the recent changes allowing new type of cards on the reef details page, the heat stress alert level card is slow to load and we can see a grey card for a few seconds on normal sites. We should load it by default and replace it with another one.

And/or find a way to make the check happen faster.

Screen Shot 2022-01-19 at 3 54 00 PM

ericboucher:
Fixes #661


packages/api/src/sites/dto/live-data.dto.ts
```diff
@@ -1,8 +1,10 @@
 import { TimeSeriesValueDto } from '../../time-series/dto/time-series-value.dto';
+import { LatestData } from '../../time-series/latest-data.entity';
 import { SofarLiveData } from '../../utils/sofar.types';
```
tleraj, line 3: On my editor, I see a dependency cycle error. Specifically:

For line 2:
```
Dependency cycle via ../sites/sites.entity:11
```

For line 3:
```
Dependency cycle via ../time-series/latest-data.entity:1=>../sites/sites.entity:11
```

Is that your case as well? If not, then it might have something to do with my local settings.

packages/website/src/store/Sites/types.ts
```diff
@@ -56,6 +56,7 @@ export type ExclusionDateResponse = {
 
 export interface LiveData {
   site: { id: number };
+  latestData: any[];
```
tleraj, line 59: Think we can do better than `any[]`?

tleraj (commented):
Looks good!! Just some comments, but other than these I think it's good to go.

PR #667, opened 2022-01-27
Sonde Upload Enhancements

tleraj:
The purpose of this PR is to introduce the following enhancements:
 - [x] Refresh materialized view on data upload
 - [x] Display new data on site page after data upload is complete
 - [x] Allow survey point creation on upload page
 - [x] Display latest data on water sampling card

### DIFF
--- a/packages/api/migration/1643364520677-DataUploadsAddCreatedUpdatedColumns.ts
+++ b/packages/api/migration/1643364520677-DataUploadsAddCreatedUpdatedColumns.ts
@@ -1,0 +1,16 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class DataUploadsAddCreatedUpdatedColumns1643364520677 implements MigrationInterface {
+    name = 'DataUploadsAddCreatedUpdatedColumns1643364520677'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "data_uploads" ADD "created_at" TIMESTAMP NOT NULL DEFAULT now()`);
+        await queryRunner.query(`ALTER TABLE "data_uploads" ADD "updated_at" TIMESTAMP NOT NULL DEFAULT now()`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "data_uploads" DROP COLUMN "updated_at"`);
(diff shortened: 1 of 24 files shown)


packages/website/src/routes/SiteRoutes/UploadData/NewPointDialog.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { useState } from "react";
```
ericboucher, line 1: This feels quite duplicative of the functionality we already have in other places, no?

tleraj: In https://github.com/aqualinkorg/aqualink-app/pull/667/commits/0d4696678467dc133bc073631c94c3eb3df898ef I reused this new component to the other place we had a new survey point dialog, which was in the survey media upload page :wink:

packages/api/migration/1643364520677-DataUploadsAddCreatedUpdatedColumns.ts
```diff
@@ -0,0 +1,16 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class DataUploadsAddCreatedUpdatedColumns1643364520677 implements MigrationInterface {
+    name = 'DataUploadsAddCreatedUpdatedColumns1643364520677'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "data_uploads" ADD "created_at" TIMESTAMP NOT NULL DEFAULT now()`);
```
ericboucher, line 7: it's nice to have create_at, although for the feature we are implementing, we likely want to use "max_date" and not created_at to take the most recent data, not the most recently created file.

For updated_at, you would need to add a specific trigger to make it useful, no?

tleraj: As of https://github.com/aqualinkorg/aqualink-app/pull/667/commits/9b4b690e4f2bb00395259f54ece288dbb3bccf7d we use `max_date`. Also the `@UpdateDateColumn` decorator provides the desired functionality, so no triggers are needed.

ericboucher: Hmm, ok, so it means typeorm will handle it, but we won't see the change if it happens outside of the API, correct?

ericboucher (approved):

Issue #628, opened 2021-09-13
Adding exclusion period should delete data


ericboucher:
As a site admin, when I add an exclusion date, I would like to see it affect the charts.

In other words, the corresponding data should be filtered out from the time_series table or deleted.

ericboucher:
@imskr thanks for proposing your help!

imskr:
Thanks @ericboucher

imskr:
@ericboucher while setting up the project locally and running `yarn start` i'm getting following error:

```
➜ yarn start
yarn run v1.22.11
$ lerna run start -- --parallel
lerna notice cli v3.22.1
lerna info versioning independent
lerna info Executing command in 2 packages: "yarn run start --parallel"
lerna ERR! yarn run start --parallel exited 1 in 'api'
lerna ERR! yarn run start --parallel stdout:
$ yarn run start:prod --parallel
$ node dist/src/main.js --parallel
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run start --parallel stderr:
/home/skr/Github/aqualink-app/node_modules/multer-google-storage/lib/index.js:48
            throw new Error('You have to specify bucket for Google Cloud Storage to work.');
            ^

Error: You have to specify bucket for Google Cloud Storage to work.
```

Do I have to setup GCP to get this running locally?

ericboucher:
Hmm, good catch @imskr, let me add a quick fix for that!

imskr:
Thanks @ericboucher

imskr:
@ericboucher I'm still getting the same error on 1dev-no-upload` branch

ericboucher:
@imskr have you tried the updated master branch? It should work find but let me know if it doesn't and we can investigate :)

PR #668, opened 2022-01-31
Delete time series data when adding exclusion periods

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/628.
With this implementation, we delete the corresponding time series data whenever a user adds an exclusion period. Also, we take the existing exclusion periods into consideration whenever we run the spotter backfill script, so that it does not backfill data for these periods.

### Designs
| dialog | dates completed | `ADD PERIOD` button clicked |
|---|---|---|
| Screenshot_20220204_165336 | Screenshot_20220204_165406 | Screenshot_20220204_165419 |

### DIFF
--- a/packages/api/src/sites/dto/spotter-data.dto.ts
+++ b/packages/api/src/sites/dto/spotter-data.dto.ts
@@ -1,6 +1,6 @@
 import { TimeSeriesValueDto } from '../../time-series/dto/time-series-value.dto';
 
 export class SpotterDataDto {
-  topTemperature: TimeSeriesValueDto[];
-  bottomTemperature: TimeSeriesValueDto[];
+  topTemperature?: TimeSeriesValueDto[];
+  bottomTemperature?: TimeSeriesValueDto[];
 }
--- a/packages/api/src/sites/dto/spotter-data.dto.ts
+++ b/packages/api/src/sites/dto/spotter-data.dto.ts
@@ -1,6 +1,6 @@
 import { TimeSeriesValueDto } from '../../time-series/dto/time-series-value.dto';
 
 export class SpotterDataDto {
-  topTemperature?: TimeSeriesValueDto[];
-  bottomTemperature?: TimeSeriesValueDto[];
+  topTemperature: TimeSeriesValueDto[];
+  bottomTemperature: TimeSeriesValueDto[];
 }
--- a/packages/api/src/sites/sites.service.ts
+++ b/packages/api/src/sites/sites.service.ts
@@ -24,7 +24,7 @@ import {
   getRegion,
   getTimezones,
   handleDuplicateSite,
-  filterSpotterDataByDate,
+  filterMetricDataByDate,
   getConflictingExclusionDates,
   hasHoboDataSubQuery,
   getLatestData,
@@ -354,10 +354,10 @@ export class SitesService {
     );
 
     return {
-      topTemperature: filterSpotterDataByDate(topTemperature, exclusionDates),
-      bottomTemperature: filterSpotterDataByDate(
-        bottomTemperature,
+      topTemperature: filterMetricDataByDate(exclusionDates, topTemperature),
+      bottomTemperature: filterMetricDataByDate(
         exclusionDates,
+        bottomTemperature,
       ),
     };
   }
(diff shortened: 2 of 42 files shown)


ericboucher:
Let's add a warning popup for confirmation - to make sure users are aware that data will actually be deleted.

packages/api/src/sites/sites.service.ts
```diff
@@ -399,11 +405,44 @@ export class SitesService {
...
+        `Exclusion period [${startDate}, ${endDate}] already exists`,
+      );
+    }
+
+    // We assume that a specific spotter corresponds to only one source.
+    const source = await this.sourceRepository.findOne({
+      where: { sensorId: site.sensorId, type: SourceType.SPOTTER },
+    });
+
+    if (!source) {
```
ericboucher, line 423: If no source is found, it could mean that no data has been uploaded yet but just the spotter just got added. So let's just return early instead, and send back a warning, not an error?

packages/api/src/utils/spotter-time-series.ts
```diff
@@ -77,6 +79,49 @@ const getSpotterSources = (
...
+    exclusionDatesRepository.find({ where: { sensorId: source.sensorId } }),
+  );
+
+/**
+ * Filters out spotter data whose timestamp falls into any exclusion date interval.
+ * @param data The spotter data to filter
+ * @param exclusionDates An array of exclusion dates
+ * @returns The filtered spotter data
+ */
+const excludeSpotterData = (
```
ericboucher, line 102: I have a feeling this function already exists somewhere / or similar. It's great to create util functions, make sure to leverage them in old code as well.

packages/api/src/sites/sites.service.ts
tleraj: On second thoughts, we could delete the data only if the corresponding source exists. Otherwise, just add the exclusion period on the `exclusion_dates` table, and when the spotter data will be backfilled, this period will be taken into account.
I adopted this logic in https://github.com/aqualinkorg/aqualink-app/pull/668/commits/d04a1466377668b996688b5c8c8a55b002a5f26b.

packages/api/src/utils/spotter-time-series.ts
tleraj: Leveraged a pre-existing util in https://github.com/aqualinkorg/aqualink-app/pull/668/commits/c09780cd71b53ab21a48e269956a22bcaf19a871.

packages/api/src/sites/dto/spotter-data.dto.ts
```diff
@@ -1,6 +1,6 @@
 import { TimeSeriesValueDto } from '../../time-series/dto/time-series-value.dto';
 
 export class SpotterDataDto {
-  topTemperature: TimeSeriesValueDto[];
-  bottomTemperature: TimeSeriesValueDto[];
+  topTemperature?: TimeSeriesValueDto[];
```
ericboucher, line 4: is this change needed? Why not return empty lists when needed instead of making it optional?

packages/website/src/store/Sites/selectedSiteSlice.ts
```diff
@@ -213,6 +215,29 @@ const selectedSiteSlice = createSlice({
       }
       return state;
     },
+    excludeSourceData: (
```
ericboucher, line 218: Why do we need to do this in the frontend?

ericboucher (commented):
Nice work @tleraj, a few open questions but we're getting there

packages/website/src/store/Sites/selectedSiteSlice.ts
tleraj: When we make a request to delete time series data, we do not want to immediately to make a new request to fetch (a possibly huge amount) of the updated data, in order to view them on the frontend, so we handle the already stored data on redux store.

packages/api/src/sites/dto/spotter-data.dto.ts
ericboucher: For spotters we should always get the same format, so I think we should keep it as it was with non optional arrays.

packages/website/src/store/Sites/selectedSiteSlice.ts
ericboucher: Let's dispatch a new requests, excluding data should be pretty rare and happen at the beginning of a spotter's life

Issue #638, opened 2021-12-02
ApiTimeSeriesResponse should not return source which have 0 data


ericboucher:
Our `ApiTimeSeriesResponse` should be more sparse and only return objects if there is data. And let the frontend check if the source_type it's looking for exists.

At the moment, the API is returning a lot of unnecessary objects full of empty lists.

Issue #639, opened 2021-12-05
Return time-series grouped by metric/source instead of source/metric


ericboucher:
As we start to gather more and more data, we time-series API should return data grouped by metric type first instead of source. This will make it easier to display data coming from multiple sources on the same charts.

Linked to #638

tleraj:
I think I can take this :slightly_smiling_face:

PR #669, opened 2022-02-01
Do not return sources with no data

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/638.
This PR is a prework for https://github.com/aqualinkorg/aqualink-app/issues/639.
If a source metric has no data, we simply do not return it instead of an empty array. Likewise, if we have a source with no data for each metric, then we do not return the source at all.

## TODOs
 - [x] Do not return sources with no data
 - [x] Fix API tests

### DIFF
--- a/packages/api/src/utils/time-series.utils.ts
+++ b/packages/api/src/utils/time-series.utils.ts
@@ -27,16 +27,13 @@ interface TimeSeriesGroupable {
   source: SourceType;
 }
 
-type TimeSeriesResponse<T> = Record<SourceType, Record<Metric, T[]>>;
+type TimeSeriesResponse<T> = Record<SourceType, Partial<Record<Metric, T[]>>>;
 
 // TODO: Revisit the response structure and simplify when we have more metrics and sources available
 export const emptyMetricsSourcesObject = Object.values(SourceType).reduce(
   (root, key) => ({
     ...root,
-    [key]: Object.values(Metric).reduce(
-      (sources, source) => ({ ...sources, [source]: [] }),
-      {},
-    ),
+    [key]: {},
   }),
   {},
 ) as TimeSeriesResponse<TimeSeriesGroupable>;
--- a/packages/api/src/utils/time-series.utils.ts
+++ b/packages/api/src/utils/time-series.utils.ts
@@ -27,17 +27,11 @@ interface TimeSeriesGroupable {
   source: SourceType;
 }
 
-type TimeSeriesResponse<T> = Record<SourceType, Partial<Record<Metric, T[]>>>;
+type TimeSeriesResponse<T> = Partial<
+  Record<SourceType, Partial<Record<Metric, T[]>>>
+>;
 
 // TODO: Revisit the response structure and simplify when we have more metrics and sources available
-export const emptyMetricsSourcesObject = Object.values(SourceType).reduce(
-  (root, key) => ({
-    ...root,
-    [key]: {},
-  }),
-  {},
-) as TimeSeriesResponse<TimeSeriesGroupable>;
-
 export const groupByMetricAndSource = <T extends TimeSeriesGroupable>(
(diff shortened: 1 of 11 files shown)


packages/api/src/utils/time-series.utils.ts
```diff
@@ -27,20 +27,11 @@ interface TimeSeriesGroupable {
   source: SourceType;
 }
 
-type TimeSeriesResponse<T> = Record<SourceType, Record<Metric, T[]>>;
+type TimeSeriesResponse<T> = Partial<
+  Record<SourceType, Partial<Record<Metric, T[]>>>
+>;
 
 // TODO: Revisit the response structure and simplify when we have more metrics and sources available
```
ericboucher, line 34: Can this TODO be removed?

tleraj: Maybe we could keep it until we start working on https://github.com/aqualinkorg/aqualink-app/issues/639

packages/website/src/common/SiteDetails/WaterSampling/index.tsx
```diff
@@ -92,21 +91,24 @@ const WaterSamplingCard = ({
...
-          uploadHistory.find(
-            ({ surveyPoint }) => surveyPoint.id.toString() === pointId
-          ) || {};
-        if (from && to) {
+        const {
+          minDate: from,
+          maxDate: to,
+          surveyPoint,
+        } = head(uploadHistory) || {};
+        if (from && to && typeof surveyPoint?.id === "number") {
```
ericboucher, line 99: The max_date and min_date are enforced not null in the DB, so is the survey point. So maybe check one out of the 3 but no need to check the 3?

packages/website/src/common/SiteDetails/WaterSampling/index.tsx
```diff
@@ -77,11 +82,10 @@ const WaterSamplingCard = ({ siteId }: WaterSamplingCardProps) => {
   const classes = useStyles();
   const [minDate, setMinDate] = useState<string>();
   const [maxDate, setMaxDate] = useState<string>();
-  const [pointId, setPointId] = useState<number>();
-  const [pointName, setPointName] = useState<string>();
+  const [point, setPoint] = useState<SurveyPoint>();
   const [sondeData, setSondeData] = useState<TimeSeriesData["sonde"]>();
   const meanValues = calculateSondeDataMeanValues(sondeData);
-  const isPointNameLong = pointName ? pointName.length > 24 : false;
+  const isPointNameLong = point?.name ? point.name.length > 24 : false;
```
ericboucher, line 88: how about: `(point?.name?.length || 0 ) > 24`

ericboucher: You could also create and use this right below:

const surveyPointDisplayName = `${isPointNameLong ? "" : "Survey point:"} ${point?.name || point.id}`

packages/website/src/common/SiteDetails/WaterSampling/index.tsx
```diff
@@ -85,7 +85,21 @@ const WaterSamplingCard = ({ siteId }: WaterSamplingCardProps) => {
...
+  const surveyPointDisplayName =
+    point?.name || isPointIdValid
+      ? `${isPointNameLong ? "" : " Survey point:"} ${point?.name || point?.id}`
+      : undefined;
+  const viewUploadButtonLink =
+    minDate && maxDate && isPointIdValid
+      ? `/sites/${siteId}${requests.generateUrlQueryParams({
+          start: minDate,
+          end: maxDate,
+          surveyPoint: point?.id,
```
ericboucher, line 99: surveyPoint: isPointIdValid ? point.id | undefined

ericboucher (approved):

PR #670, opened 2022-02-03
Group time series by metric/source

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/639.
Until now, we grouped time series data first by source and then by metric. Since we want to display data coming from multiple sources on the same charts, with this implenentation we first group time series data by metric and then by source.


packages/api/src/utils/time-series.utils.ts
```diff
@@ -99,7 +134,7 @@ export const getDataRangeQuery = (
 ): Promise<TimeSeriesRange[]> => {
   const surveyPointCondition = surveyPointId
     ? `(source.survey_point_id = ${surveyPointId} OR source.survey_point_id is NULL)`
-    : 'source.survey_point_id is NULL';
+    : '1=1';
```
ericboucher, line 137: @tleraj Is this enough? We need to consider the getDataRangeQuery as a whole and how we want it to behave. Can you clarify what this will return in each case and if that matches what we need to make decisions in the frontend?

tleraj: So on the `/site:id` page, where we request data from the `/api/time-series/sites/{{site_id}}/range` endpoint, we fetch for each metric the range of the survey point with the most recent max date, so the response looks something like this:

```json
    "salinity": {
        "sonde": {
            "surveyPoint": {
                "id": 2,
                "name": "Point #1"
            },
            "data": [
                {
                    "minDate": "2021-09-30T08:15:00.000Z",
                    "maxDate": "2021-09-30T10:30:00.000Z"
                }
            ]
        }
    },
    "conductivity": {
        "sonde": {
            "surveyPoint": {
                "id": 36,
                "name": "Point #2"
            },
            "data": [
                {
                    "minDate": "2021-09-30T08:15:00.000Z",
                    "maxDate": "2021-09-30T10:30:00.000Z"
                }
            ]
        }
    }
```

and the results on the chart look like this:

| salinity | conductivity |
|---|---|
| Screenshot_20220303_114836 | Screenshot_20220303_114859 |


On the other hand, on the `/sites/:siteId/points/surveyPointId` page, where we only fetch time series data and ranges regarding the selected survey point.

However, the only use we have now for the time series data ranges on the frontend, is to declare if the site has hobo data, and if so, enable the `MAX` range button. But on the future, if we would like to display for each metric chart its available range, this info is in our hands now.

Let me know if we want to change the range display logic on the frontend (e.g. for each metric chart display the available range as we do on the `TEMPERATURE ANALYSIS` chart).

ericboucher: Yes, we should enable it for all metrics, not just for hobo data

tleraj: After https://github.com/aqualinkorg/aqualink-app/pull/685/commits/8e74ae2995134dc08f635fafdf9c8337788c7604, the result looks like this:

Screenshot_20220303_191344

ericboucher: The range values seem odd, no? Why is it stuck at 03/03/2022?

ericboucher (approved):

PR #675, opened 2022-02-08
Fetch survey point with the latest data for each metric

tleraj:
On the site page, for each metric, we need to fetch the survey point time series data for the survey point with the most recent data.

### DIFF
--- a/packages/api/src/utils/time-series.utils.ts
+++ b/packages/api/src/utils/time-series.utils.ts
@@ -1,4 +1,4 @@
-import _, { omit } from 'lodash';
+import _, { last, omit, isNull } from 'lodash';
 import { IsNull, Repository } from 'typeorm';
 import { Site } from '../sites/sites.entity';
 import { SourceType } from '../sites/schemas/source-type.enum';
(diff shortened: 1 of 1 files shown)


packages/api/src/utils/time-series.utils.ts
```diff
@@ -61,7 +74,7 @@ export const getDataQuery = (
 
   const surveyPointCondition = surveyPointId
     ? `(source.survey_point_id = ${surveyPointId} OR source.survey_point_id is NULL)`
-    : 'source.survey_point_id is NULL';
+    : `1=1`;
```
ericboucher, line 77: I simplified the logic a bit using `1=1`

packages/api/src/utils/time-series.utils.ts
```diff
@@ -99,7 +116,7 @@ export const getDataRangeQuery = (
 ): Promise<TimeSeriesRange[]> => {
   const surveyPointCondition = surveyPointId
     ? `(source.survey_point_id = ${surveyPointId} OR source.survey_point_id is NULL)`
-    : 'source.survey_point_id is NULL';
+    : '1=1';
```
ericboucher, line 119: you also need to consider the getDataRangeQuery. It might need more changes than what I did here

PR #682, opened 2022-02-16
Fix firebase spotter backfill error

tleraj:
The purpose of this PR is to fix the following error that occured on the spotter data backfill function on firebase:

```
A firebase error has occurred on firebase function scheduledSpotterTimeSeriesUpdate:
TypeError: Cannot read properties of undefined (reading 'sensorId')
```

This error occurs because there might be spotters with no exclusion dates, thus the

```typescript
exclusionDatesRepository.find({ where: { sensorId: source.sensorId } })
```
will return `[]`.
With the following approach, if a spotter has no exclusion dates, we simply return its data, otherwise we exclude the data that should be excluded.

### DIFF
--- a/packages/api/src/utils/spotter-time-series.ts
+++ b/packages/api/src/utils/spotter-time-series.ts
@@ -1,5 +1,5 @@
 import { Logger } from '@nestjs/common';
-import { has, times } from 'lodash';
+import { get, times } from 'lodash';
 import moment from 'moment';
 import { Connection, In, IsNull, Not, Repository } from 'typeorm';
 import Bluebird from 'bluebird';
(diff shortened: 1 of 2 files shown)


packages/api/src/utils/spotter-time-series.ts
```diff
@@ -180,11 +179,19 @@ export const addSpotterData = async (
             return DEFAULT_SPOTTER_DATA_VALUE;
           }
 
-          const sensorExclusionDates = sensorToExclusionDates[site.sensorId];
+          const sensorExclusionDates = has(
+            sensorToExclusionDates,
+            site.sensorId,
+          )
+            ? sensorToExclusionDates[site.sensorId]
+            : undefined;
```
ericboucher, line 187: I think you can simply use a `get` default undefined

ericboucher (approved):

Issue #683, opened 2022-02-17
Make the data upload more flexible


ericboucher:
We would like to adapt our data upload function to accept more formats and files.

In particular, we need to:
- [ ] accept CSV data
- [ ] ingest meteorological data

To do so, we should implement a regex mapping from column names to aqualink db implementation and units.

```
  // Meteorological Data
  'Pressure, mbar': Metric.PRESSURE,
  'Rain, mm': Metric.PRECIPITATION,
  'Temp, °C': Metric.TOP_TEMPERATURE,
  'RH, %': Metric.RH,
  'Wind Speed, m/s': Metric.WIND_SPEED,
  'Gust Speed, m/s': Metric.WIND_GUST_SPEED,
  'Wind Direction, ø': Metric.WIND_DIRECTION,
```

PR #685, opened 2022-02-21
Allow csv file uploads

tleraj:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/683.
We introduce the following new metrics (a migration for this is coming):
```typescript
PRESSURE = 'pressure',
PRECIPITATION = 'precipitation',
RH = 'rh',
WIND_GUST_SPEED = 'wind_gust_speed',
AIR_TEMPERATURE = 'air_temperature'
```
We use the same parsing function for csv files from `node-xlsx`, as we did for xlsx files. The only difference is on the timestamp generation, since meteorological data files have a different column format: `Date Time GMT +XX:XX`
We also introduce a stronger mimetype typing, `Mimetype`, so that we only accept specific file types.
In addition, we tweak the header row finding util function, refactored some already existing utils and created some others.

### TODOs
 - [x] Add metrics enum update migration
 - [x] Frontend work to allow csv files selection
 - [ ] ~~Investigate on using regexps to recognize sheet columns~~ Will be investigated in a future PR


ericboucher:
@tleraj  we should also introduce a new type of temperature: AIR_TEMPERATURE

Indeed, met temp data should not be mixed up with water temp data

packages/api/scripts/upload-sonde-data.ts
```diff
@@ -50,23 +51,28 @@ async function run() {
...
   const connection = await createConnection(config);
 
+  const isCsv = filePath.endsWith('.csv');
+
   logger.log('Uploading sonde data');
-  await uploadSondeData(
+  await uploadTimeSeriesData(
     filePath,
     last(filePath.split('/')),
+    isCsv
```
ericboucher, line 70: I think this logic should happen within uploadTimeSeriesData

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -71,6 +90,24 @@ export const fileFilter: MulterOptions['fileFilter'] = (
...
 };
 
+export const requiredHeadersForMimetype = (mimetype: Mimetype): string[] => {
+  switch (mimetype) {
+    case 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet':
+      return ['Date (MM/DD/YYYY)', 'Time (HH:mm:ss)', 'Time (Fract. Sec)'];
+    case 'text/csv':
+      return ['Date Time'];
+    default:
+      return [];
```
ericboucher, line 100: These are not linked to the mimetype. This is just the case with the data we are getting at the moment, but certainly not linked to the mimetype as we could have a csv or an xlsx with other requirements.

packages/website/src/routes/SiteRoutes/UploadData/Selectors.tsx
```diff
@@ -27,6 +27,7 @@ type EnhancedSelectOption = SelectOption & { disabled?: boolean };
 
 const SENSOR_TYPES: EnhancedSelectOption[] = [
   { id: 0, name: "sonde", label: "Sonde data" },
+  { id: 1, name: "metlog", label: "Meteorological data" },
   // TODO: Enable these options once the API can support various sensors data upload.
```
ericboucher, line 31: can we remove this todo?

packages/website/src/routes/SiteRoutes/UploadData/Selectors.tsx
```diff
@@ -27,6 +27,7 @@ type EnhancedSelectOption = SelectOption & { disabled?: boolean };
 
 const SENSOR_TYPES: EnhancedSelectOption[] = [
   { id: 0, name: "sonde", label: "Sonde data" },
+  { id: 1, name: "metlog", label: "Meteorological data" },
```
ericboucher, line 30: is it normal to have overlapping ids (1)?

packages/website/src/constants/metlogConfig.ts
```diff
@@ -0,0 +1,85 @@
...
+  | "top_temperature"
+  | "pressure"
+  | "precipitation"
+  | "rh"
+  | "wind_gust_speed"
+>;
+
+export const metlogConfig: Record<MetlogMetricsKeys, MetlogConfig> = {
+  top_temperature: {
+    title: "Top Temperature",
```
ericboucher, line 25: This should be "Air Temperature"

packages/website/src/constants/metlogConfig.ts
```diff
@@ -0,0 +1,85 @@
...
+  visibility: string;
+  order: number;
+  convert?: number;
+}
+
+export type MetlogMetricsKeys = Extract<
+  MetricsKeys,
+  | "wind_direction"
+  | "wind_speed"
+  | "top_temperature"
```
ericboucher, line 16: air_temperature

packages/website/src/routes/SiteRoutes/UploadData/Selectors.tsx
tleraj: No, my bad :stuck_out_tongue:

packages/website/src/routes/SiteRoutes/UploadData/Selectors.tsx
tleraj: It's just there to remind us that we should enable these options once our API supports it. But I can remove it :slightly_smiling_face:

packages/website/src/routes/SiteRoutes/UploadData/DropZone.tsx
```diff
@@ -11,8 +11,8 @@ const GREY_COLOR = grey[500];
 const MAX_SIZE_MB = 10;
 const ACCEPTED_TYPES = [
   // TODO: Add these once our API supports them.
-  // { extension: "csv", mimeType: "text/csv" },
   // { extension: "xls", mimeType: "application/vnd.ms-excel" },
```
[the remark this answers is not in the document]
tleraj: Do we have any sample files so that I can test it?

packages/api/src/utils/uploads/upload-sheet-data.ts
tleraj: I removed the mimetype dependency in https://github.com/aqualinkorg/aqualink-app/pull/685/commits/80145d833927c57a31d5dca4256456b63f8c3d86. Let me know what you think :slightly_smiling_face:

packages/website/src/routes/SiteRoutes/UploadData/Selectors.tsx
ericboucher: Maybe rephrase it then?

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -43,22 +43,51 @@ const metricsMapping: Record<string, Metric> = {
   'Temp °C': Metric.BOTTOM_TEMPERATURE,
   'Battery V': Metric.SONDE_BATTERY_VOLTAGE,
   'Cable Pwr V': Metric.SONDE_CABLE_POWER_VOLTAGE,
+  'Pressure, mbar (LGR S/N: 21067056, SEN S/N: 21062973)': Metric.PRESSURE,
```
ericboucher, line 46: So one of the reasons we need to use regex is because in the next files, the S/N will probably be different. So we need a better mapping logic that does not depend on `(LGR S/N: 21067056, SEN S/N: 21062973)`

tleraj:
Currently, there is a bug when uploading meteorological data, like the ones in the `metlog240921_021021.csv`. With https://github.com/aqualinkorg/aqualink-app/pull/685/commits/1fec824df9db332fc0c85642325b0c2a6ef6c348, we do not look for exact matches with the `metricsMapping` dict, but we use a `csvHeader.toLowerCase().startsWith(metricKey.toLowerCase())` logic, so the following `renameKeys` util in https://github.com/aqualinkorg/aqualink-app/blob/1fec824df9db332fc0c85642325b0c2a6ef6c348/packages/api/src/utils/uploads/upload-sheet-data.ts#L116-L125

does not rename the keys properly.

ericboucher:
@tleraj maybe move the renaming at the very beginning and with the new mapping logic. So that after you can simply use the final metrics names to do all the other checks. What do you think?

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -43,22 +43,53 @@ const metricsMapping: Record<string, Metric> = {
   'Temp °C': Metric.BOTTOM_TEMPERATURE,
   'Battery V': Metric.SONDE_BATTERY_VOLTAGE,
   'Cable Pwr V': Metric.SONDE_CABLE_POWER_VOLTAGE,
+  'Pressure, mbar': Metric.PRESSURE,
+  'Rain, mm': Metric.PRECIPITATION,
+  'Temp, °C': Metric.AIR_TEMPERATURE,
```
ericboucher, line 48: You have two keys with the exact same name, maybe the mapping needs to be by source?

ericboucher: @tleraj

tleraj: Fixed in https://github.com/aqualinkorg/aqualink-app/pull/685/commits/a7021651e8987a7c15bfb81b12f5e1c20a8348d8

ericboucher (approved):

PR #690, opened 2022-03-11
Improve CSV Date Parse

tleraj:
The purpose of this PR is to improve the date parsing logic for data uploads.
With this approach, we do not parse csv data as raw, so the dates are parsed as strings, which are manipulated to generate a timestamp for each sheet row.

### TODOs
 - [x] Handle possible timezones
 - [x] end to end testing from the frontend
 - [x] add a note regarding the order of overlapping column names


packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -139,26 +139,11 @@ const renameKeys = (
...
   const hourSerial = dataObject['Time (HH:mm:ss)'];
   const seconds = dataObject['Time (Fract. Sec)'];
-  // Extract EXCEL date and hours from serials:
-  const date = ExcelDateToJSDate(dateSerial);
-  const hours = (hourSerial * 24) % 24;
-  const hoursFloor = Math.floor(hours);
-  const decimalHours = hours - hoursFloor;
-  date.setHours(hoursFloor);
-  date.setMinutes(Math.round(decimalHours * 60));
+  const date = new Date(`${dateSerial} ${hourSerial} UTC`);
```
ericboucher, line 146: This is not a dateSerial anymore, just a date string. Now that we use string all the time we might have a better way to take care of the multi column date

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -172,9 +189,17 @@ const getTimestampFromCsvDateString = (dataObject: any) => {
     return undefined;
   }
 
-  const timezone = last(dateKey.split(', '));
+  const timezoneKeyword = TIMEZONE_KEYWORDS.find((keyword) =>
```
ericboucher, line 192: I think you can use a "simple" regex like this, since for our purpose, GMT/UTC are the same thing:

`[+-][0-9]{2}:[0-9]{2}\b` and then use Date(`${dataObject[dateKey]} UTC${timezone}`)

ericboucher (approved):

PR #697, opened 2022-03-29
Make data upload date columns case insensitive

ericboucher:
Make data upload date columns case insensitive and other improevements


packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -73,12 +73,27 @@ const IGNORE_HEADER_KEYS = [
   'Site Name',
   'Time (HH:mm:ss)',
   'Time (Fract. Sec)',
-];
+].map((key) => key.toLowerCase());
```
ericboucher, line 76: @tleraj is there a smarter way to do this?

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -190,6 +212,7 @@ const getTimestampFromDateString = (dataObject: any) => {
   );
 
   if (!dateKey) {
+    // TODO - Can we raise an error here instead? Is it safe?
```
ericboucher, line 215: @tleraj wdyt?

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -445,6 +468,7 @@ export const uploadTimeSeriesData = async (
       'timestamp',
     );
 
+    // TODO - Why do we have this check on surveyPoint?
```
ericboucher, line 471: @tleraj can you explain why we have this check? And document if needed?

packages/api/src/utils/uploads/upload-sheet-data.ts
tleraj: Not sure what you mean. We can store the array as lowercase in the first place, i.e. 

```typescript
const IGNORE_HEADER_KEYS = [
  '#',
  'site name',
  'time (hh:mm:ss)',
  'time (fract. sec)'
];
```

packages/api/src/utils/uploads/upload-sheet-data.ts
tleraj: Yeah, sounds good!! :+1:

packages/api/src/utils/uploads/upload-sheet-data.ts
tleraj: Not sure what you mean. We could store this array to lowercase in the first place, i.e.
```typescript
const IGNORE_HEADER_KEYS = [
  '#',
  'site name',
  'time (hh:mm:ss)',
  'time (fract. sec)'
]
```

packages/api/src/utils/uploads/upload-sheet-data.ts
tleraj: On the frontend, we need to specify a survey point when we upload sonde or metlog data. If this is not necessary, we can remove this if statement.

ericboucher: I think it should be necessary on the frontend for now, but not on the backend

Issue #701, opened 2022-04-04
Save files to google cloud during data uploads


ericboucher:
When a user updates a dataset using the "data upload" feature, we should keep a version of the file in google storage for reference.

This will help with debugging, and in case we need to re-import the data

PR #707, opened 2022-04-14
Save data uploads to gcloud

echaidemenos:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/701

### DIFF
--- a/packages/api/src/data-uploads/data-uploads.entity.ts
+++ b/packages/api/src/data-uploads/data-uploads.entity.ts
@@ -44,6 +44,9 @@ export class DataUploads {
   @Column('character varying', { array: true, nullable: false })
   metrics: Metric[];
 
+  @Column({ nullable: true })
+  gCloudLocation: string;
+
   @CreateDateColumn()
   createdAt: Date;
 
--- a/packages/api/src/data-uploads/data-uploads.entity.ts
+++ b/packages/api/src/data-uploads/data-uploads.entity.ts
@@ -45,7 +45,7 @@ export class DataUploads {
   metrics: Metric[];
 
   @Column({ nullable: true })
-  gCloudLocation: string;
+  fileLocation: string;
 
   @CreateDateColumn()
   createdAt: Date;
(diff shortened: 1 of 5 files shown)


packages/api/src/data-uploads/data-uploads.entity.ts
```diff
@@ -44,6 +44,9 @@ export class DataUploads {
   @Column('character varying', { array: true, nullable: false })
   metrics: Metric[];
 
+  @Column({ nullable: true })
+  gCloudLocation: string;
```
ericboucher, line 48: Let's make it cloud agnostic and call it something like "fileLocation" or "fileStorageLocation"

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -442,6 +443,16 @@ export const uploadTimeSeriesData = async (
       );
     }
 
+    // Initialize google cloud service, to be used for media upload
+    const googleCloudService = new GoogleCloudService();
+
+    const gCloudPath = await googleCloudService.uploadFile(
```
ericboucher, line 449: I would like this to happen before processing, so let's move the following function call below the upload:

```
const results = findSheetDataWithHeader(
      fileName,
      workSheetData,
      sourceType,
      mimetype,
    );
```

packages/api/src/google-cloud/google-cloud.service.ts
```diff
@@ -23,7 +23,7 @@ export class GoogleCloudService {
 
   constructor(
     @InjectRepository(SurveyMedia)
-    private surveyMediaRepository: Repository<SurveyMedia>,
+    private surveyMediaRepository?: Repository<SurveyMedia>,
```
ericboucher, line 26: Why do we need this change?

echaidemenos: Because this dependency is used by only one function of the service, and someone using the service may not necessary need it. Maybe a cleaner solution could be a derived class from `GoogleCloudService` implementing `findDanglingFiles`

ericboucher (approved):

Issue #703, opened 2022-04-07
Loading site page is slow


ericboucher:
Let's understand which queries are taking too long and optimize our API calls accordingly.

Especially if the data we are waiting for is not necessary to start rendering.

echaidemenos:
The app, groups multiple request and waits them all to start rendering. One of them is `livedata` taking most of the time, because it makes multiple requests to an external API. I will de-group it since its only a small part of the data shown and it will load separately

PR #708, opened 2022-04-19
Separate liveData requests for faster page load

echaidemenos:
https://github.com/aqualinkorg/aqualink-app/issues/703

Splits `liveData` from `Site`, allowing components that don't need the the result of the time demanding liveData request to render.

### DIFF
--- a/packages/website/src/common/SiteDetails/index.tsx
+++ b/packages/website/src/common/SiteDetails/index.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import {
   createStyles,
   Grid,
@@ -11,7 +11,7 @@ import {
 import classNames from "classnames";
 import times from "lodash/times";
 
-import { useSelector } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import Map from "./Map";
 import FeaturedMedia from "./FeaturedMedia";
 import Satellite from "./Satellite";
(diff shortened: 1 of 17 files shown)


packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -44,33 +49,36 @@ const SiteDetails = ({
 }: SiteDetailsProps) => {
   const classes = useStyles();
   const theme = useTheme();
+  const liveData = useSelector(liveDataSelector);
+  const liveDataLoading = useSelector(siteLiveDataLoadingSelector);
```
ericboucher, line 53: could we just assume that the data is loading while `liveData` it's undefined?

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -185,7 +193,11 @@ const SiteDetails = ({
         {cards.map((Component, index) => (
           <Grid key={index.toString()} item xs={12} sm={6} md={3}>
             <div className={classes.card}>
-              <LoadingSkeleton variant="rect" height="100%" loading={isLoading}>
+              <LoadingSkeleton
+                variant="rect"
+                height="100%"
+                loading={isLoading || liveDataLoading}
```
ericboucher, line 199: won't this make all the components wait until liveData has loaded?

ericboucher: It's ok if they all depend on the liveData but I am not sure that's the case?

echaidemenos: these skeletons are for the 4 cards under the map. They all depend on `liveData`
https://github.com/aqualinkorg/aqualink-app/blob/4ab6418c03fbc18ebef2405168ee7351c704c190/packages/website/src/common/SiteDetails/index.tsx#L60-L81

ericboucher: Indeed, thanks for checking

packages/website/src/common/SiteDetails/index.tsx
ericboucher: This works ok, but we are having the same problem as #709 I think. We need to make sure to reset these values to undefined.

As it is know, navigating to a new site keeps the cards up with the value from the site you are navigating from, which is not ideal.

echaidemenos:
this commit will also resolve https://github.com/aqualinkorg/aqualink-app/issues/709

ericboucher (approved):

Issue #712, opened 2022-04-20
Data upload is very slow


ericboucher:
Uploading a file from the frontend is very slow, even for a small file of 100 lines or so.

Let's analyze where time is spent during an upload and investigate ways to make it smarter.

ericboucher:
The main factor was that we were waiting for a full refresh of the "latest_data" materialized view, which could take up to a minute to complete.

To mitigate that we:
- tweaked the material view to use indices and run a lot faster
- not block the thread while the view is refreshing

PR #713, opened 2022-04-21
Fix data upload is very slow

echaidemenos:
https://github.com/aqualinkorg/aqualink-app/issues/712

Uploading 85kb file, for me, would take on average 2 seconds, of which, on average, 1 second was for the upload to gCloud. I added an async alternative for uploading files to gCloud, to skip the waiting time.

### DIFF
--- a/packages/api/src/google-cloud/google-cloud.service.ts
+++ b/packages/api/src/google-cloud/google-cloud.service.ts
@@ -33,6 +33,27 @@ export class GoogleCloudService {
     });
   }
 
+  private getDestination(
+    filePath: string,
+    type: string,
+    dir: string,
+    prefix: string,
+  ): string {
+    const folder = `${this.STORAGE_FOLDER}/${dir}/`;
+    const basename = path.basename(filePath);
+    return getRandomName(folder, prefix, basename, type);
+  }
+
+  public uploadFileAsync(
+    filePath: string,
+    type: string,
+    dir: string = 'surveys',
+    prefix: string = 'site_hobo_image',
+  ): string {
+    this.uploadFile(filePath, type, dir, prefix);
+    return this.getDestination(filePath, type, dir, prefix);
+  }
+
   public async uploadFile(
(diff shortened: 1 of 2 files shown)


packages/api/src/google-cloud/google-cloud.service.ts
```diff
@@ -33,6 +33,27 @@ export class GoogleCloudService {
...
+    return getRandomName(folder, prefix, basename, type);
+  }
+
+  public uploadFileAsync(
+    filePath: string,
+    type: string,
+    dir: string = 'surveys',
+    prefix: string = 'site_hobo_image',
+  ): string {
+    this.uploadFile(filePath, type, dir, prefix);
```
ericboucher, line 53: How do you guarantee that this is the correct destination if it is random?

We should probably do something like this instead:
```
const destination = this.getDestination(filePath, type, dir, prefix);
this.uploadFile(filePath, type, destination);
return destination
```

No?

echaidemenos: oh yes, my bad!

packages/api/src/google-cloud/google-cloud.service.ts
```diff
@@ -50,24 +54,29 @@ export class GoogleCloudService {
...
   ): string {
-    this.uploadFile(filePath, type, dir, prefix);
-    return this.getDestination(filePath, type, dir, prefix);
+    const dest = this.getDestination(filePath, type, dir, prefix);
+    this.uploadFile(filePath, dest);
+    return dest;
   }
 
-  public async uploadFile(
+  public async uploadFileSync(
```
ericboucher, line 62: should it be called `uploadFileAsync`?

echaidemenos: my reasoning for the names here is the same as fs.readFileSync. They are both meant to be blocking

ericboucher (approved):

PR #714, opened 2022-04-22
Use latestData instead of liveData for cards

ericboucher:
Fetching liveData is too slow and can take up to a few seconds. This is mostly due to the response time of the Sofar API and out of our control.

Since we already save most of the data on an hourly basis in latestData, let's try to use that instead.

As a quick win to make things look faster, I removed the skeleton loading for cards when the rest of the site data is ready.

@echaidemenos, could you take over and:
- split the API endpoint for latestData out of liveData
- use latestData inputs in cards


echaidemenos (approved):

packages/website/src/common/SiteDetails/Satellite/__snapshots__/index.test.tsx.snap
```diff
@@ -349,7 +349,7 @@ exports[`renders as expected 1`] = `
                 >
                   Last data received
                    
-                  5 min. ago
+                  24 days ago
```
ericboucher, line 352: You need to mock these values in tests
Otherwise they will keep changing and CI will fail (as it is doing now [failed tests](https://app.circleci.com/pipelines/github/aqualinkorg/aqualink-app/4020/workflows/c074ba53-1c85-4641-8aac-a3b64fe0fd0c/jobs/5331)

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -76,33 +95,81 @@ const SiteDetails = ({
     };
   }, [dispatch]);
 
+  useEffect(() => {
+    if (latestData) {
+      // Data for Satellite
+      // not sure if here 'dhw' is the one
+      const degreeHeating = latestData.find((x) => x.metric === "dhw");
```
echaidemenos, line 94: Is this the correct value here or another metric should be added here @ericboucher ?

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -76,33 +95,81 @@ const SiteDetails = ({
...
+      setDataForSensor({
+        topTemperature: latestDataToSofarValue(topTemperature),
+        bottomTemperature: latestDataToSofarValue(bottomTemperature),
+      });
+
+      // Data for Wave
+      const waveHeight = latestData.find(
+        (x) => x.metric === "significant_wave_height"
+      ); // not sure if here 'significant_wave_height' is the one
+      const waveMeanDirection = latestData.find(
```
echaidemenos, line 121: same here

packages/website/src/common/SiteDetails/Waves/index.test.tsx
```diff
@@ -3,12 +3,42 @@ import { render } from "@testing-library/react";
...
+  const waveMeanDirection = mockLatestData.find(
+    (x) => x.metric === "wave_mean_direction"
+  );
+  const waveMeanPeriod = mockLatestData.find(
+    (x) => x.metric === "wave_mean_period"
+  );
+  const windSpeed = mockLatestData.find((x) => x.metric === "wind_speed");
+  const windDirection = mockLatestData.find(
+    (x) => x.metric === "wind_direction"
+  );
```
ericboucher, line 28: This seems very repetitive. We already have functions to go to snake_case to camelCase when the names are same. Otherwise you could use a mapping.

Have a look at functions such as `mapMetrics` in `helpers.ts`

packages/website/src/common/SiteDetails/Waves/index.tsx
```diff
@@ -288,7 +288,7 @@ const useStyles = makeStyles((theme: Theme) =>
 );
 
 interface WavesProps {
-  liveData: LiveData;
+  data: WavesDataProps;
```
ericboucher, line 291: Why not replace `liveData` by `latestDataParsed` or something like that? Is it necessary to create these intermediate types for Props?

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -23,7 +23,12 @@ import Surveys from "./Surveys";
 import CardWithTitle from "./CardWithTitle";
 import { Value } from "./CardWithTitle/types";
 import CombinedCharts from "../Chart/CombinedCharts";
-import type { Site } from "../../store/Sites/types";
+import type {
+  Site,
+  SatelliteDataProps,
+  SensorDataProps,
+  WavesDataProps,
```
ericboucher, line 30: See my comment above. This seems unnecessary

packages/website/src/common/SiteDetails/index.tsx
ericboucher: dhw is degreeHeatingWeeks, so to get degreeHeatingDays you need to multiply by 7

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -60,36 +75,93 @@ const SiteDetails = ({
...
       dispatch(unsetLiveData());
+      dispatch(unsetLatestData());
     };
   }, [dispatch]);
 
+  useEffect(() => {
+    if (latestData) {
+      // Data for Satellite
+      // not sure if here 'dhw' is the one
+      const degreeHeating = latestData.find((x) => x.metric === "dhw");
```
ericboucher, line 94: See my comment about WaveProps and such. It seems that we are adding a lot of complexity that is not necessary.

Instead, create a parser for latestData to make it more similar to the liveData we had and use that instead.

ericboucher: And make sure that you select the data by timestamp, for the same metric that is

packages/website/src/common/Chart/MultipleSensorsCharts/index.tsx
```diff
@@ -120,11 +120,11 @@ const MultipleSensorsCharts = ({
   );
 
   const hasSondeData = Boolean(
```
ericboucher, line 122: Make sure this gets resolved early enough to decide on what metrics to fetch from the backend

packages/website/src/common/SiteDetails/index.tsx
ericboucher: They can come from different sources, so no

packages/api/src/utils/spotter-time-series.ts
```diff
@@ -236,3 +207,161 @@ export const addSpotterData = async (
...
+          }
+        }),
+      );
+    },
+    { concurrency: 4 },
+  );
+
+  // Update materialized view
+  logger.log('Refreshing materialized view latest_data');
+  await connection.query('REFRESH MATERIALIZED VIEW latest_data');
```
echaidemenos, line 366: Why should we wait for this? in `upload-sheet-data.ts` we changed it so it is non blocking

echaidemenos (approved):

echaidemenos (approved):

packages/website/src/store/Sites/helpers.ts
```diff
@@ -318,3 +320,29 @@ export const timeSeriesRequest = async (
     !attachDirection ? end : maxDate,
   ];
 };
+
+export const parseLatestData = (data: LatestData[]): LatestDataASSofarValue => {
+  if (!data || data.length === 0) return {};
+
+  // Coping, sorting and filtering are done to keep the latest timestamp for each Metric
```
ericboucher, line 327: actually we want to prioritize spotter value when possible.

We will prioritize spotter values over forecast values when possible. If we had no spotter data for 12 hours,  use other sources

Issue #717, opened 2022-04-27
Add HOBO data uploads


ericboucher:
Leveraging the work that was done for Stanford, we would like to accept HOBO data in our upload module

In other words, take the relevant part of the code in `/api/src/utils/uploads/upload-hobo-data.ts` and move it to `/api/src/utils/uploads/upload-sheet-data.ts`.

Finally, remove the "coming soon" mention in the data upload module

PR #718, opened 2022-04-28
Enable HOBO data uploads

echaidemenos:
This PR closes https://github.com/aqualinkorg/aqualink-app/issues/717

Enables Hobo data option for Sensor type at the upload data page

### DIFF
--- a/packages/api/src/time-series/time-series.module.ts
+++ b/packages/api/src/time-series/time-series.module.ts
@@ -1,14 +1,9 @@
 import { Module } from '@nestjs/common';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { DataUploads } from '../data-uploads/data-uploads.entity';
-import { Region } from '../regions/regions.entity';
 import { SiteSurveyPoint } from '../site-survey-points/site-survey-points.entity';
-import { HistoricalMonthlyMean } from '../sites/historical-monthly-mean.entity';
 import { Site } from '../sites/sites.entity';
 import { Sources } from '../sites/sources.entity';
-import { SurveyMedia } from '../surveys/survey-media.entity';
-import { Survey } from '../surveys/surveys.entity';
-import { User } from '../users/users.entity';
 import { EntityExists } from '../validations/entity-exists.constraint';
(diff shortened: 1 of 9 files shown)


ericboucher:
@echaidemenos I should have given more details.

This will use the same data upload module. So we are ONLY interested in the temperature data. We won't be uploading survey images at this point.

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -15,6 +16,7 @@ import { SourceType } from '../../sites/schemas/source-type.enum';
 import { DataUploads } from '../../data-uploads/data-uploads.entity';
 import { getSite, getSiteAndSurveyPoint } from '../site.utils';
 import { GoogleCloudService } from '../../google-cloud/google-cloud.service';
+import { parseCSV } from './upload-hobo-data';
```
ericboucher, line 19: @echaidemenos instead of doing this, we should add the "hobo" functionality to this file and use the structure already in place. Ultimately, the `upload-hobo-data` file should be deprecated and this file should accept more and more data sources / formats

echaidemenos: logic in `upload-hobo-data.ts` is only used by upload hobo data script, which creates sites, TimeSeries etc. Its purpose is so specific, so i am not sure why we need to merge these 2 files

ericboucher: We don't need to merge these two files. We need to use a common parseCSV logic for all files. So we should not import this function in here

packages/api/src/time-series/time-series.module.ts
```diff
@@ -19,6 +24,11 @@ import { TimeSeriesService } from './time-series.service';
       SiteSurveyPoint,
       Sources,
       DataUploads,
+      User,
+      Survey,
+      SurveyMedia,
```
ericboucher, line 29: @echaidemenos can you remove all the changes concerning surveys? Since we will only be uploading temp for now?

ericboucher (approved):

ericboucher (approved):

Issue #689, opened 2022-03-11
Add date selector to url for site and survey point pages


ericboucher:
When the user selects a new time range to display the data for, we would like to have the URL change and include the start=...&end=... to make the page view shareable.

In reverse, loading a site page with start and end parameters should load the set startDate and endDate and fetch the data for that period.

PR #719, opened 2022-04-28
Add date start & end query parameter to url

echaidemenos:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/689
Adds query parameters to the url of the site page, so the view can be shareable
Screenshot from 2022-04-28 17-00-14

### DIFF
--- a/packages/api/migration/1649334332499-AddForeignKeyToTimeSeriesReferencingDataUploads.ts
+++ b/packages/api/migration/1649334332499-AddForeignKeyToTimeSeriesReferencingDataUploads.ts
@@ -1,0 +1,31 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddForeignKeyToTimeSeriesReferencingDataUploads1649334332499
+  implements MigrationInterface
+{
+  name = 'AddForeignKeyToTimeSeriesReferencingDataUploads1649334332499';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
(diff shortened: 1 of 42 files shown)


ericboucher:
And that also works on survey_point pages when we have one?

echaidemenos:
> And that also works on survey_point pages when we have one?

Yes. Both pages use the same component

ericboucher (approved):

ericboucher:
Likely because of timezone, when we go to http://localhost:3000/sites/1225?start=2021-05-04&end=2022-05-05 the dates that appear are one day off. Not a huge deal but let's see if we can find an easy fix

Screen Shot 2022-05-03 at 4 37 28 PM

ericboucher (changes requested):
Let's see if we can resolve the 1 day offset due to timezone. Note that the charts are presented in the sites specific timezone

echaidemenos:
> Likely because of timezone, when we go to http://localhost:3000/sites/1225?start=2021-05-04&end=2022-05-05 the dates that appear are one day off. Not a huge deal but let's see if you can find an easy fix
> 
> Screen Shot 2022-05-03 at 4 37 28 PM

It looks like the date picker registers it as the next date if time is after 21:59:59.999. For example 2022-02-04T21:59:59.999Z will show 2022-02-04 but 2022-02-04T22:00:00.000Z will show 22-02-05 regardless the time zone. We can keep like it is now, or we an trim the time off so the same date will show on both the date-picker and url

ericboucher (approved):

PR #727, opened 2022-05-09
Ocean Monitoring - text and image update

ericboucher:
- Update the landing page
- Update site registration page


packages/website/src/common/Chart/MultipleSensorsCharts/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Box, Container, makeStyles, Theme } from "@material-ui/core";
 import moment from "moment";
 import { camelCase, isNaN, isNumber, sortBy } from "lodash";
 import { useDispatch, useSelector } from "react-redux";
-import { utcToZonedTime } from "date-fns-tz";
+import { utcToZonedTime, zonedTimeToUtc } from "date-fns-tz";
```
ericboucher, line 7: @echaidemenos could you test the changes in this commit when you get a chance? I've done some QA but it would be nice to get your thoughts as well.

echaidemenos: If i paste a link like this `http://localhost:3000/sites/2?start=2022-04-14&end=2022-05-14` the date will shift from  4-14 to 4-13

ericboucher: Not for me, it might be because the sites in your database don't have a timezone set?
The previous implementation created shifts for me so that's why I tried this

echaidemenos (approved):

PR #733, opened 2022-05-24
Send message to slack channel when a new site is created

PysarenkoDS-BWT:


packages/api/src/sites/sites.service.ts
```diff
@@ -51,6 +51,8 @@ import {
 import { getTimeSeriesDefaultDates } from '../utils/dates';
 import { SourceType } from './schemas/source-type.enum';
 import { TimeSeries } from '../time-series/time-series.entity';
+import { sendSlackMessage, SlackMessage } from '../utils/slack.utils';
+import { functionsConfig } from '../../functionsConfig';
```
ericboucher, line 55: The functionsConfig are meant for the firebase functions. Since this is running in the API server, I think it is better to import it from process.env directly to keep things separated

    token: process.env.SLACK_BOT_TOKEN,
    channel: process.env.SLACK_BOT_CHANNEL

ericboucher (approved):

PR #748, opened 2022-06-01
Merge forecastData and latestData in frontend

ericboucher:
The logic in SiteDetails is too complicated and not giving the expected results, @echaidemenos this is a quick fix but please take over to impement the following changes.

API:
- add the source field to the forecast_data

Frontend:
- get latestData
- get forecastData (add the source field)
- merge the two into latestData
- keep the logic we had before to prioritize spotter data

### DIFF
--- a/packages/website/src/store/Sites/helpers.ts
+++ b/packages/website/src/store/Sites/helpers.ts
@@ -352,11 +352,11 @@ export const parseLatestData = (
   // eslint-disable-next-line fp/no-mutating-methods
   const sorted = filtered.sort((x, y) => {
     // if spotter data is available and, use it.
-    if (x.source === "spotter") {
+    if (x.source === "spotter" && y.source !== "spotter") {
       return +1;
     }
 
-    if (y.source === "spotter") {
+    if (y.source === "spotter" && x.source !== "spotter") {
       return -1;
     }
 


echaidemenos (approved):

packages/api/src/wind-wave-data/wind-wave-data.service.ts
```diff
@@ -15,19 +14,12 @@ export class WindWaveService {
...
-    const getSofarValue = (metric: WindWaveMetric) => {
-      const forecast = result.find((x) => x.metric === metric);
-      return { timestamp: forecast?.timestamp, value: forecast?.value };
-    };
-    return {
+    return result.map((item) => ({
+      timestamp: item.timestamp,
+      value: item.value,
+      source: item.source,
+      metric: item.metric,
```
ericboucher, line 21: maybe just deconstruct `item`?

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -91,52 +90,30 @@ const SiteDetails = ({
...
-        validMetrics.includes(x.metric)
+    if (forecastData && latestData) {
+      const combinedArray = [...forecastData, ...latestData];
+      const formatted = combinedArray.map((x) => ({
+        timestamp: x.timestamp,
+        value: x.value,
+        source: x.source,
+        metric: x.metric,
+      }));
+      const spotterValidityLimit = 12 * 60 * 60 * 1000; // 12 hours
```
ericboucher, line 101: @echaidemenos I don't think we need this anymore. All of this anymore and everything should be handled in `parseLatestData`. What do you think?

packages/api/migration/1654159594680-AddSourceToForecastData.ts
```diff
@@ -0,0 +1,43 @@
...
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "forecast_data" DROP CONSTRAINT "FK_2319fce6c102102ca2f4dae0349"`,
+    );
+    await queryRunner.query(
+      `DROP INDEX "public"."IDX_2319fce6c102102ca2f4dae034"`,
+    );
+    await queryRunner.query(`DROP TABLE "forecast_data"`);
+    await queryRunner.query(
+      `CREATE TABLE "forecast_data" ("id" SERIAL NOT NULL, "timestamp" TIMESTAMP NOT NULL, "value" double precision NOT NULL, "metric" "public"."metrics_metric_enum" NOT NULL, "created_at" TIMESTAMP NOT NULL DEFAULT now(), "updated_at" TIMESTAMP NOT NULL DEFAULT now(), "site_id" integer NOT NULL, CONSTRAINT "one_row_per_site_per_metric" UNIQUE ("site_id", "metric"), CONSTRAINT "PK_95ee099ba4892af55de59d5bc86" PRIMARY KEY ("id"))`,
```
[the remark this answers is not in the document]
echaidemenos: Thanks for the comment. The reason was that in undo migration, the less strict constraint failed. But you are right i will just select and delete rows that fail the constraint and then change it!

Issue #753, opened 2022-06-06
Add an index on "site_id" in the sketchfab table


ericboucher:
The  `site_id` column in the sketchfab table should be indexed since since we are using it as a child of "site".

I thought it would have been indexed automatically as a foreign key... So it might be worth checking some of our other foreign key columns and make sure that they have the necessary indexes.

PR #754, opened 2022-06-07
Add indexes to foreign keys

echaidemenos:
This PR adds indexes to a lot of foreign  key columns. It also renames `wind-wave-data.entity.ts` to `forecast-data.entity.ts`.
@ericboucher let me know if there is any to add or remove.

The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/753

### DIFF
--- a/packages/api/migration/1654608487703-AddIndexes.ts
+++ b/packages/api/migration/1654608487703-AddIndexes.ts
@@ -1,0 +1,107 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddIndexes1654608487703 implements MigrationInterface {
+  name = 'AddIndexes1654608487703';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_5ae17cee9e339b189cf3e220e5" ON "site_audit" ("site_id") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_4f925485b013b52e32f43d430f" ON "collection" ("user_id") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_1ed4a7be821c5819d382e4b72e" ON "data_uploads" ("site_id") `,
+    );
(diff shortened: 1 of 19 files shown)


packages/api/src/wind-wave-data/forecast-data.entity.ts
```diff
@@ -15,13 +15,13 @@ import { WindWaveMetric } from './wind-wave-data.types';
 
 @Entity()
 @Unique('one_row_per_site_per_metric_per_source', ['site', 'metric', 'source'])
-@Index(['site'])
```
ericboucher, line 18: just curious, what's the difference?

packages/api/scripts/hindcast-wind-wave-data.ts
```diff
@@ -3,7 +3,7 @@ import yargs from 'yargs';
 import { configService } from '../src/config/config.service';
 import { Site } from '../src/sites/sites.entity';
 import { addWindWaveData } from '../src/utils/hindcast-wind-wave';
-import { ForecastData } from '../src/wind-wave-data/wind-wave-data.entity';
```
ericboucher, line 6: Any need to change the table name? Or it should work as expected?

packages/api/migration/1654608487703-AddIndexes.ts
```diff
@@ -0,0 +1,107 @@
...
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_2e5de2284644a41979436d3820" ON "historical_monthly_mean" ("site_id") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_ff7a13fae129f15180a3e36f9e" ON "site" ("region_id") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_2d56028954ec00388e734fb266" ON "site" ("stream_id") `,
+    );
```
ericboucher, line 45: I think these are not needed since we will likely use this relation the other way round

ericboucher (approved):

Issue #711, opened 2022-04-20
Optimize liveData function


ericboucher:
At the moment, the liveData query is quite slow. Let's investigate how to make it faster. Or are we limited by Sofar's API?

ericboucher:
To make it faster, we should switch to leveraging hindcast endpoints instead of forecast. See similar changes in #722

PR #756, opened 2022-06-09
Replace live_data request with spotter_position

echaidemenos:
The purpose of this PR is to close https://github.com/aqualinkorg/aqualink-app/issues/711

This PR removes unnecessary FE request to `live_data` endpoint. It also creates a new `spotter_position` endpoint for getting the latest spotter coordinates and Deployed status.

### DIFF
--- a/packages/api/src/sites/sites.controller.ts
+++ b/packages/api/src/sites/sites.controller.ts
@@ -110,10 +110,8 @@ export class SitesController {
   @ApiParam({ name: 'id', example: 1 })
   @Public()
   @Get(':id/spotter_position')
-  findSpotterPosition(
-    @Param('id', ParseIntPipe) id: number,
-  ): Promise<SofarLiveDataDto> {
-    return this.sitesService.findLiveData(id);
+  findSpotterPosition(@Param('id', ParseIntPipe) id: number) {
+    return this.sitesService.findSpotterPosition(id);
   }
 
   @ApiNestNotFoundResponse('No site was found with the specified id')
--- a/packages/api/src/sites/sites.controller.ts
+++ b/packages/api/src/sites/sites.controller.ts
@@ -106,6 +106,17 @@ export class SitesController {
   }
 
   @ApiNestNotFoundResponse('No site was found with the specified id')
+  @ApiOperation({ summary: 'Returns spotter position for the specified site' })
+  @ApiParam({ name: 'id', example: 1 })
+  @Public()
+  @Get(':id/spotter_position')
+  findSpotterPosition(
+    @Param('id', ParseIntPipe) id: number,
+  ): Promise<SofarLiveDataDto> {
+    return this.sitesService.findLiveData(id);
+  }
+
+  @ApiNestNotFoundResponse('No site was found with the specified id')
   @ApiOperation({ summary: 'Returns latest data for the specified site' })
   @ApiParam({ name: 'id', example: 1 })
   @Public()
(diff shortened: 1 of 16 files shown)


packages/api/src/sites/sites.controller.ts
```diff
@@ -105,6 +105,17 @@ export class SitesController {
...
 
+  @ApiNestNotFoundResponse('No site was found with the specified id')
+  @ApiOperation({ summary: 'Returns spotter position for the specified site' })
+  @ApiParam({ name: 'id', example: 1 })
+  @Public()
+  @Get(':id/spotter_position')
+  findSpotterPosition(
+    @Param('id', ParseIntPipe) id: number,
+  ): Promise<SofarLiveDataDto> {
+    return this.sitesService.findLiveData(id);
```
ericboucher, line 116: Does it make sense to return the full `this.sitesService.findLiveData(id);` ?

ericboucher: Or this should be `findSpotterPosition`?

packages/api/src/sites/sites.service.ts
```diff
@@ -351,6 +354,47 @@ export class SitesService {
     };
   }
 
+  async findSpotterPosition(id: number): Promise<{
```
ericboucher, line 357: I don't think this logic belongs to the backend. The backend should simply return a position with a timestamp. And the api users define how they want to use it.

So maybe this endpoint returns:
- position
- timestamp
- site lat/long
- site status

packages/api/src/workers/dailyData.ts
```diff
@@ -45,6 +45,7 @@ export async function getDegreeHeatingDays(
   endOfDate: Date,
   maxMonthlyMean: number | null,
 ): Promise<ValueWithTimestamp | undefined> {
+  console.time(`--> getDegreeHeatingDays for ${latitude}, ${longitude}`);
```
ericboucher, line 48: 👣

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -228,7 +232,9 @@ const SiteDetails = ({
           {site && !isSketchFabView && (
             <Map
               siteId={site.id}
-              spotterPosition={liveData?.spotterPosition}
+              spotterPosition={
+                hastSpotterData ? spotterPosition?.spotterPosition : null
```
ericboucher, line 236: There is a typo hastSpotterData -> hasSpotterData

packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -95,15 +95,19 @@ const SiteDetails = ({
...
-      const hasSpotter = Boolean(parsedData.bottomTemperature);
       const hasSonde = Boolean(parsedData.salinity);
 
       setHasSondeData(hasSonde);
-      setHasSpotterData(hasSpotter);
       setLatestDataAsSofarValues(parsedData);
     }
   }, [forecastData, latestData]);
 
+  useEffect(() => {
```
ericboucher, line 105: I think we can keep the same logic as before for `hasSpotter`

packages/website/src/store/Sites/types.ts
```diff
@@ -369,7 +369,14 @@ export interface SitesListState {
 export interface SelectedSiteState {
   draft: SiteUpdateParams | null;
   details?: Site | null;
-  liveData?: LiveData | null;
+  spotterPosition?: {
+    spotterPosition?: {
```
ericboucher, line 373: maybe name it `position`? It seems redundant to name the child object `spotterPosition` again

packages/api/src/sites/sites.service.ts
```diff
@@ -351,6 +354,43 @@ export class SitesService {
...
+      return {
+        timestamp: undefined,
+        isDeployed,
+        spotterPosition: undefined,
+      };
+
+    const spotterRaw = await getSpotterData(sensorId);
+    const spotterData = spotterRaw
+      ? {
+          longitude:
```
ericboucher, line 374: let's change lat/long to be numbers since we already have the timestamp in the spotterPosition object

ericboucher (approved):

ericboucher (approved):

Issue #691, opened 2022-03-15
Use image thumbnails for faster image load time


ericboucher:
Some survey and site images are very big. Let's investigate ways to use image thumbnails instead.

Maybe this is a feature offered by google storage by default?

PR #764, opened 2022-06-15
Create resized survey images

echaidemenos:
The purpose of this PR is to resolve https://github.com/aqualinkorg/aqualink-app/issues/691

This PR:
- Adds a script that gets all image urls from `survey_media` table and uploads new resized versions. Example usage `yarn resize-survey-images -s 512`. The resized images urls would be in the form of `https://storage.googleapis.com/reef_media_programize/be-dev/surveys/site-image-24cc936cc50ce8c8.jpg=s512`
- Makes uploading a survey picture saves it's resized version. The width must be set with `SURVEY_IMAGE_RESIZE` env variable.
- `/site/:siteId` route requests first for resized survey pictures, if they exists, to reduce loading time. `/sites/:siteId/survey_details/:surveyId` shows the original media.


packages/api/src/surveys/surveys.service.ts
```diff
@@ -19,10 +19,19 @@ import { GoogleCloudService } from '../google-cloud/google-cloud.service';
 import { Site } from '../sites/sites.entity';
 import { getFileFromURL } from '../utils/google-cloud.utils';
 import { getSite } from '../utils/site.utils';
+import { getImageData, resize } from '../../scripts/utils/image';
+import { validateMimetype } from '../uploads/mimetypes';
 
 @Injectable()
 export class SurveysService {
   private logger: Logger = new Logger(SurveysService.name);
+  private readonly SURVEY_IMAGE_RESIZE = Number(
```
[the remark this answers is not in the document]
ericboucher: Also, what happens if the uploaded image is smaller than our thumbnail size?

packages/website/src/common/SiteDetails/Surveys/SurveyCard/index.tsx
```diff
@@ -60,11 +62,18 @@ const SurveyCard = ({
               <Link to={`/sites/${siteId}/survey_details/${survey.id}`}>
                 <CardMedia
                   className={classes.cardImage}
+                  component="img"
                   image={
                     isShowingFeatured
-                      ? survey.featuredSurveyMedia?.url
+                      ? `${survey.featuredSurveyMedia?.url}=s${REDUCED_IMAGE_SIZE}`
```
ericboucher, line 68: this should not happen here. The api could return "url" and "thumbnailUrl", which you could use in the frontend.

The frontend should not need to know anything about the reduced_image_size

packages/api/src/surveys/surveys.service.ts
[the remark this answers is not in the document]
echaidemenos: > Also, what happens if the uploaded image is smaller than our thumbnail size?

we check if image is smaller before resizing and uploading. So in some cases the original image will be kept only

packages/api/src/surveys/surveys.service.ts
```diff
@@ -25,9 +25,7 @@ import { validateMimetype } from '../uploads/mimetypes';
 @Injectable()
 export class SurveysService {
   private logger: Logger = new Logger(SurveysService.name);
-  private readonly SURVEY_IMAGE_RESIZE = Number(
-    process.env.SURVEY_IMAGE_RESIZE,
-  );
+  private readonly surveyImageResizeWidth = 512;
```
ericboucher, line 28: Maybe add a comment above to explain what this number is used for?

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
```diff
@@ -53,12 +54,18 @@ const FeaturedMedia = ({
...
+    const thumbnailUrl = getThumbnailLink(featuredImage);
     return (
       <Link to={`/sites/${siteId}/survey_details/${surveyId}`}>
         <CardMedia
           className={classes.card}
           style={{ height: "100%" }}
-          image={featuredImage}
+          image={thumbnailUrl}
+          component="img"
+          onError={(e: any) => {
```
ericboucher, line 65: This still does not feel right. I think the backend should be the one sending the thumbnail link when it exists, so the survey object should return both the imageUrl and thumbnailUrl (when it exists).

packages/website/src/routes/HomeMap/SiteTable/SelectedSiteCard/__snapshots__/index.test.tsx.snap
```diff
@@ -12,7 +12,7 @@ exports[`renders as expected 1`] = `
         color="textSecondary"
         variant="h5"
       >
-        Selected Site
+        Featured Site
```
ericboucher, line 15: why did this change happen?

packages/website/src/common/SiteDetails/FeaturedMedia/index.tsx
ericboucher: So the frontend should just need to do:

- thumbnailUrl || imageUrl 

-> no need for error handling
-> no need for `getThumbnailLink`

ericboucher (changes requested):
Better. But following up on my previous comments, the frontend should not be required to know the formatting for a thumbnail url. This url should ideally be returned dircetly by the API

packages/website/src/routes/HomeMap/SiteTable/SelectedSiteCard/__snapshots__/index.test.tsx.snap
echaidemenos: a change in my .env caused this. I will bring it back

packages/website/src/utils/helpers.tsx
```diff
@@ -0,0 +1,6 @@
+export const getThumbnailLink = (url: string) => {
+  const s = url.split("/");
+  // eslint-disable-next-line fp/no-mutation
```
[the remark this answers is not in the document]
ericboucher: Agreed. Though I think we should delete this function completely @echaidemenos ;)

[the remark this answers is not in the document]
echaidemenos: thanks for the tip, i will change to this in places i use similar login, because this function will be removed

packages/api/migration/1655476789970-AddThumbnailUrlToSurveyMedia.ts
```diff
@@ -0,0 +1,25 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddThumbnailUrlToSurveyMedia1655476789970
+  implements MigrationInterface
+{
+  name = 'AddThumbnailUrlToSurveyMedia1655476789970';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "survey_media" RENAME COLUMN "url" TO "original_url"`,
```
ericboucher, line 10: Maybe we can keep url, and simply add the thumbnail_url, wdyt?

packages/api/scripts/resize-survey-images.ts
```diff
@@ -0,0 +1,137 @@
...
+    if ((imageDetails.width || 0) <= size) {
+      return 'skipped';
+    }
+    const resizedBuffer = await resize(imageBuffer, size);
+
+    const prefixed = `thumbnail-${rest.slice(-1)}`;
+    const modified = [...rest.slice(0, -1), prefixed];
+
+    // Upload file to google cloud
+    const destination = modified.join('/');
```
ericboucher, line 51: let's do this in one line

packages/api/src/surveys/surveys.service.ts
```diff
@@ -57,6 +65,46 @@ export class SurveysService {
...
+    if ((imageData.width || 0) <= this.surveyImageResizeWidth)
+      return { originalUrl };
+    const resizedImage = await resize(file.buffer, this.surveyImageResizeWidth);
+    // remove 'https://' from the string
+    const trimmed = originalUrl.substring(8);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [_domain, bucket, ...rest] = trimmed.split('/');
+    const prefixed = `thumbnail-${rest.slice(-1)}`;
+    const modified = [...rest.slice(0, -1), prefixed];
+    const destination = modified.join('/');
```
ericboucher, line 98: a lot of this seems duplicative of what's in the upload script file, maybe you could extract this as a utility function?

ericboucher (approved):
Great work @echaidemenos 

- Let's switch back to "url" instead of "originalUrl"
- Let's dedup the code a bit between the script and the service file

ericboucher (approved):

Issue #759, opened 2022-06-10
Analyze Stormglass data for our sites


ericboucher:
We gained temporary access to the stormglass API. We would like to analyze the temperature information for our sites.

As a first step, we could create a script to pull all the wave/wind data for each of our site and compare to sofar/gfs. The file (CSV) would contain the following info:

- wind/wave from our current sources
- boolean flag "in_sofar_wave_model_zone"
- wind/wave from stormglass ai
- ?wind/wave from all stormglass sources?
- name of the source selected by the stormglass ai
- site name
- site region
- link to aqualink.org site page

ericboucher:
Ping @ericboucher to get the API key

PR #766, opened 2022-06-23
Add script to create csv for stormglass/sofar comparison

echaidemenos:
This PR closes https://github.com/aqualinkorg/aqualink-app/issues/759

It adds a script `yarn stormglass-csv` that gets all sofar and stormglass wind/wave data and put them into a csv for comparison.
Note that `point_deviation_in_degrees` and `point_deviation_in_meters` refers to distance between the exact point and sofar nearest point. Also resolutions about stormglass can be found [here](https://docs.stormglass.io/#/sources).


ericboucher:
@echaidemenos please resolve the conflicts. We will merge this PR so that it can be useful in the future

ericboucher (approved):

PR #768, opened 2022-06-27
Change the logic of upload-sheet-data functions 

PysarenkoDS-BWT:
In sourceItems we describe all of table headers and by interface SourceItem

`
interface SourceItem {
  metric?: Metric;
  time?: TimeType;
  ignore?: boolean;
  convertFn?: (arg: number) => number;
}
`

here we provide next data 

1. metric - column name via we save data
2. ignore - if we don't need to save data
3. convertFn - provide function for converting value if it is needed
4. time - mark specific column for timestamp can be "Timestamp" - if one column date, "Time" and "Date" - if 2 columns date

Here we need discribe all interesting us data in 1 place - `sourceItems` object


ericboucher:
Great start. I'll do a more thorough review tomorrow. But let's implement actual tests by uploading a few sample files and adding a test script that processes them

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -114,6 +42,70 @@ const ACCEPTED_FILE_TYPES = [
...
+
+interface SourceItem {
+  metric?: Metric;
+  time?: TimeType;
+  ignore?: boolean;
+  convertFn?: (arg: number) => number;
+}
+
+const sourceItems: Record<SourceType, Record<string, SourceItem>> = {
+  [SourceType.SONDE]: {
```
ericboucher, line 60: I think we can revamp how this works and instead make the metrics mapping source agnostic and have a list of "ignored columns".

Then if a column is not ignored and not parsed to a metric, then it would give a warning.

packages/api/src/utils/uploads/upload-sheet-data.test.ts
```diff
@@ -2678,7 +2678,13 @@ test('Test convertData function', async () => {
       metric: 'sonde_cable_power_voltage',
       source: {},
     },
-  ].map((item) => ({ ...item, timestamp: new Date(item.timestamp) }));
+  ].map((item) => {
+    const date = new Date(item.timestamp);
+    const offset = date.getTimezoneOffset() / 60;
+    date.setHours(date.getHours() + offset + 3);
```
ericboucher, line 2684: why are we adding 3 here?

packages/api/src/utils/uploads/sonde_data.csv
```diff
@@ -0,0 +1,30 @@
+Kor MEASUREMENT DATA FILE EXPORT,,,,,,,,,,,,,,,,,,,,
```
ericboucher, line 1: Let's create sample test files for all the different data files I shared with you @PysarenkoDS-BWT

packages/api/src/utils/uploads/upload-sheet-data.ts
ericboucher: We will do this later.

packages/api/src/utils/uploads/upload-sheet-data.test.ts
```diff
@@ -0,0 +1,2690 @@
...
+import { SourceType } from '../../sites/schemas/source-type.enum';
+import { Sources } from '../../sites/sources.entity';
+
+test('Test getFilePathData function', async () => {
+  const data = await getFilePathData(
+    'src/utils/uploads/sonde_data.csv',
+    SourceType.SONDE,
+  );
+
+  expect(data).toStrictEqual({
```
ericboucher, line 11: let's remove this part and focus on testing the output data

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -565,7 +484,7 @@ export const uploadTimeSeriesData = async (
     console.timeEnd(`Loading into DB ${fileName}`);
     logger.log('loading complete');
 
-    refreshMaterializedView(repositories);
+    await refreshMaterializedView(repositories);
```
ericboucher, line 487: @PysarenkoDS-BWT why did you add this await?

ericboucher (approved):

PR #770, opened 2022-07-01
Add endpoint to get raw Sofar Spotter

PysarenkoDS-BWT:
Create an endpoint `/api/sensor-data` to pass raw Sofar Spotter data to users

For it we must provide next query params
id - sensorId (SPOT-1446) required
start - date of starting of data (2022-05-27) not required 
end - date of ending data (2022-05-27) not required


ericboucher (commented):
Just a few thoughts:
- why not add this as a module instead of a controller+service?
- the tests don't seem very useful as they are, no?

packages/api/src/sensor-data/sensor-data.service.ts
```diff
@@ -1,9 +1,9 @@
 import { Injectable } from '@nestjs/common';
-import { getSpotterData } from '../utils/sofar';
+import { sofarSensor } from '../utils/sofar';
 
 @Injectable()
 export class SensorDataService {
-  get(sensorId: string, endDate?: Date, startDate?: Date) {
-    return getSpotterData(sensorId, endDate, startDate);
+  get(sensorId: string, endDate?: string, startDate?: string) {
```
ericboucher, line 6: I think start and endDate are inverted between the function and its call, no?

ericboucher:
@PysarenkoDS-BWT looks like there is a small type mismatch and then we should be good to go

Issue #777, opened 2022-09-15
Add Google Tag Manager


ericboucher:
Add Google Tag Manager scripts to the frontend. Ping @ericboucher for the GTM ID.

On standard HTML sites, these are the head and body scripts:

- head
```
<!-- Google Tag Manager -->
<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({‘gtm.start’:
new Date().getTime(),event:‘gtm.js’});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!=‘dataLayer’?‘&l=‘+l:‘’;j.async=true;j.src=
’[https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f)](https://www.googletagmanager.com/gtm.js?id=%27+i+dl;f.parentNode.insertBefore(j,f));
})(window,document,‘script’,‘dataLayer’,‘GTM-ID_TO_REPLACE’);</script>
<!-- End Google Tag Manager -->
```

- body
```
<!-- Google Tag Manager (noscript) -->
<noscript><iframe src=“https://www.googletagmanager.com/ns.html?id=GTM-ID_TO_REPLACE”
height=“0" width=“0” style=“display:none;visibility:hidden”></iframe></noscript>
<!-- End Google Tag Manager (noscript) -->
```

PR #783, opened 2022-10-03
Add google tag manager

echaidemenos:
This PR resolves https://github.com/aqualinkorg/aqualink-app/issues/777

Adds google tag manager. Provide GTM ID in `REACT_APP_GA_TAG_MANAGER_ID` environment variable.


ericboucher:
Curious if we should use a package instead to make sure it works well throughout the app. Could you deploy this to the programize endpoint and make sure it's being run on all pages?

See https://medium.com/finnovate-io/integrating-google-tag-manager-with-a-react-app-5a8584ee2251

ericboucher (approved):

Issue #790, opened 2022-10-24
Add bottom pressure to line graphs


ericboucher:
The Duxbury site has been updated to SPOT-30810C and comes with bottom pressure. We would like to add it as a metric in our database and display it as a line graph when available.

PR #794, opened 2022-11-03
Add bottom pressure metric and line graphs

echaidemenos:
This PR resolves https://github.com/aqualinkorg/aqualink-app/issues/790

- adds new metric `barometric_pressure_bottom`
- renames metrics `barometric_pressure` and `barometric_pressure_diff` to `barometric_pressure_top` and `barometric_pressure_top_diff`
- adds bottom barometric pressure line graph if data are available

image

### DIFF
--- a/packages/website/src/common/SiteDetails/Sensor/index.tsx
+++ b/packages/website/src/common/SiteDetails/Sensor/index.tsx
@@ -69,8 +69,8 @@ const Sensor = ({ depth, id, data, classes }: SensorProps) => {
   const {
     topTemperature,
     bottomTemperature,
-    barometricPressure,
-    barometricPressureDiff,
+    barometricPressureTop,
+    barometricPressureTopDiff,
   } = data;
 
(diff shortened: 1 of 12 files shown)


packages/website/src/common/SiteDetails/Sensor/index.tsx
```diff
@@ -163,27 +163,29 @@ const Sensor = ({ depth, id, data, classes }: SensorProps) => {
               alignItems: "flex-end",
             }}
           >
-            {barometricPressure && (
+            {barometricPressureTopDiff && (
```
ericboucher, line 166: you could test for `barometricPressureTopDiff?.value` directly maybe?

echaidemenos: optional chaining operator is not needed, since we check for `barometricPressureTopDiff`. I will remove it from below

ericboucher (approved):

Issue #776, opened 2022-09-01
Update the featured site card to use the same data as the table


ericboucher:
Update the data used by the feature site card to match the data from the table. See an example of discrepancy below:

<img width="858" alt="Screen Shot 2022-09-01 at 11 22 11 AM" src="https://user-images.githubusercontent.com/16843267/187879871-013926ee-e4ea-4a0b-a273-521d5bac3426.png">

PR #795, opened 2022-11-04
Fix featured card data do not match table data

echaidemenos:
This PR resolves https://github.com/aqualinkorg/aqualink-app/issues/776

Data in card where from `daily_data`, while in the table they are from `collectionData`, which in reality are `latest_data`.
To resolve the issue, we updated `sites/:id` endpoint to include the same `collectionData` as the `sites/` endpoint and use them instead of `daily_data`.

### DIFF
--- a/packages/api/src/sites/sites.service.ts
+++ b/packages/api/src/sites/sites.service.ts
@@ -241,12 +241,18 @@ export class SitesService {
 
     const videoStream = await this.checkVideoStream(site);
 
+    const mappedSiteData = await getCollectionData(
+      [site],
+      this.latestDataRepository,
+    );
+
     return {
       ...site,
       surveys,
       historicalMonthlyMean,
       videoStream,
       applied: site.applied,
+      collectionData: mappedSiteData[site.id],
     };
   }
 
--- a/packages/website/src/routes/HomeMap/SiteTable/SelectedSiteCard/CardContent.tsx
+++ b/packages/website/src/routes/HomeMap/SiteTable/SelectedSiteCard/CardContent.tsx
@@ -18,9 +18,7 @@ import classNames from "classnames";
 import { Alert } from "@material-ui/lab";
 import Chart from "../../../../common/Chart";
 import { formatNumber } from "../../../../helpers/numberUtils";
-import { sortByDate } from "../../../../helpers/dates";
 import { Site } from "../../../../store/Sites/types";
-import { degreeHeatingWeeksCalculator } from "../../../../helpers/degreeHeatingWeeks";
 import { getSiteNameAndRegion } from "../../../../store/Sites/helpers";
 import { standardDailyDataDataset } from "../../../../common/Chart/MultipleSensorsCharts/helpers";
 import {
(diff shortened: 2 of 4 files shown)


packages/website/src/routes/HomeMap/SiteTable/SelectedSiteCard/CardContent.tsx
```diff
@@ -44,8 +44,20 @@ const SelectedSiteCardContent = ({
   const isTablet = useMediaQuery(theme.breakpoints.down("sm"));
   const sortedDailyData = sortByDate(site?.dailyData || [], "date");
   const dailyDataLen = sortedDailyData.length;
-  const { maxBottomTemperature, satelliteTemperature, degreeHeatingDays } =
-    (dailyDataLen && sortedDailyData[dailyDataLen - 1]) || {};
+  const {
```
ericboucher, line 47: Hmm, how about grabbing latestData directly then and using that?
I don't think we need the fall back logic and we can completely switch to using latestData instead, no?

packages/api/src/sites/sites.service.ts
```diff
@@ -241,12 +241,18 @@ export class SitesService {
...
+      this.latestDataRepository,
+    );
+
     return {
       ...site,
       surveys,
       historicalMonthlyMean,
       videoStream,
       applied: site.applied,
+      collectionData: mappedSiteData[site.id],
```
ericboucher, line 255: I think we can call this `latestData` instead

packages/website/src/routes/HomeMap/SiteTable/SelectedSiteCard/CardContent.tsx
echaidemenos: We can skip the fall back logic, but daily data are still used in the chart.

ericboucher: yes

ericboucher (approved):

packages/website/src/routes/HomeMap/SiteTable/SelectedSiteCard/CardContent.tsx
```diff
@@ -42,10 +40,8 @@ const SelectedSiteCardContent = ({
   const classes = useStyles({ imageUrl, loading });
   const theme = useTheme();
   const isTablet = useMediaQuery(theme.breakpoints.down("sm"));
-  const sortedDailyData = sortByDate(site?.dailyData || [], "date");
-  const dailyDataLen = sortedDailyData.length;
-  const { maxBottomTemperature, satelliteTemperature, degreeHeatingDays } =
-    (dailyDataLen && sortedDailyData[dailyDataLen - 1]) || {};
+  const { bottomTemperature, satelliteTemperature, dhw } =
```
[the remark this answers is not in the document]
ericboucher: In this context it's obvious (for us), but we could rename it to degreeHeatingWeeks

Issue #793, opened 2022-11-03
Add hui upload script


ericboucher:
We would like to add a new sensor type "HUI"

Already started in #773 but needs significant changes

PR #797, opened 2022-11-14
Add HUI upload script

echaidemenos:
Resolves https://github.com/aqualinkorg/aqualink-app/issues/793

Adds script for uploading HUI data in the format of the following example file: [hui-example.xlsx](https://github.com/aqualinkorg/aqualink-app/files/10013753/hui-example.xlsx)  
(please not there is no `QA issues or comments` column)

The script, generally, parses the data from the file, using using [turfs](https://www.npmjs.com/package/@turf/clusters-dbscan)` implementation of [DBSCAN](https://en.wikipedia.org/wiki/DBSCAN) groups them into sites and site survey points. Lastly it creates site, site survey points and sources, if they do not exist, and uploads time series data.

Also this PR refactors `upload-sheed-data.ts` to make it more re-usable.


packages/api/migration/1668517874752-AddHUIDataSource.ts
```diff
@@ -0,0 +1,103 @@
...
+    );
+    await queryRunner.query(`DROP TYPE "public"."sources_type_enum"`);
+    await queryRunner.query(
+      `ALTER TYPE "public"."sources_type_enum_old" RENAME TO "sources_type_enum"`,
+    );
+
+    await queryRunner.query(
+      `CREATE MATERIALIZED VIEW "latest_data"
+                AS SELECT
+                    DISTINCT ON (metric, type, site_id, survey_point_id)
```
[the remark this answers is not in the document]
echaidemenos: Good point it is a bit unclear. But according to the [documentation](https://www.postgresql.org/docs/current/sql-select.html#SQL-DISTINCT), since there is an `ORDER BY` it will always select the one with the latest `timestamp`, in our case.

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -171,7 +195,20 @@ const getTimeStamp = (
   timezone?: string,
 ) => {
   const isArray = Array.isArray(index);
-  if (isArray) return new Date(`${item[index[0]]} ${item[index[1]]}`);
+  if (
+    isArray &&
+    typeof item[index[0]] === 'string' &&
+    typeof item[index[1]] === 'string'
```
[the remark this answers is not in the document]
echaidemenos: For example: 
```
const a = [];
const b = [];
console.log(typeof b[a[42]])
```
would result in `'undefined'`.
But i am gonna change the if body bellow where i am doing:
```
date.getDate() + item[index[0]]
```
which would result in `NaN,` to 
```
date.getDate() + (item[index[0]] || 0)
// also here 
SECONDS_IN_DAY * (item[index[1]] || 0)
```

packages/api/migration/1668517874752-AddHUIDataSource.ts
```diff
@@ -0,0 +1,103 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddHUIDataSource1668517874752 implements MigrationInterface {
+  name = 'AddHUIDataSource1668517874752';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "data_uploads" ALTER COLUMN "site_id" DROP NOT NULL`,
```
ericboucher, line 8: Why would we have data with null site_id?

packages/api/migration/1668517874752-AddHUIDataSource.ts
```diff
@@ -0,0 +1,103 @@
...
+    );
+    await queryRunner.query(
+      `ALTER TABLE "data_uploads" ALTER COLUMN "survey_point_id" DROP NOT NULL`,
```
ericboucher, line 11: Same here?

packages/api/migration/1668517874752-AddHUIDataSource.ts
```diff
@@ -0,0 +1,103 @@
...
+    await queryRunner.query(`DROP TYPE "public"."metrics_metric_enum"`);
+    await queryRunner.query(
+      `ALTER TYPE "public"."metrics_metric_enum_old" RENAME TO "metrics_metric_enum"`,
+    );
+
+    await queryRunner.query(
+      `CREATE TYPE "public"."sources_type_enum_old" AS ENUM('gfs', 'hobo', 'metlog', 'noaa', 'sofar_wave_model', 'sonde', 'spotter')`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "forecast_data" ALTER COLUMN "source" DROP DEFAULT`,
```
ericboucher, line 69: do we need to make any changes to forecast_data?

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -144,14 +168,14 @@ export const fileFilter: MulterOptions['fileFilter'] = (
 const headerMatchesKey = (header: string, key: string) =>
   header.toLowerCase().startsWith(key.toLowerCase());
 
-async function refreshMaterializedView(repositories: Repositories) {
+export const refreshMaterializedView = async (
```
ericboucher, line 171: while we're at it, it would be a good idea to extract this as a util that can be re-used throughout the code base. It could accept either a repository or a connection object since that's how we're using it elsewhere.

Eg. `connection.query('REFRESH MATERIALIZED VIEW latest_data');` in `sst-time-series.ts`

packages/api/migration/1668517874752-AddHUIDataSource.ts
echaidemenos: Because HUI files contain multiple sites. We count change the relation to one to many. I think this information (site-file relation) is only used to list data upload in sites, in order to delete them. The data delete would still work fine, but the file would not be listed in the FE.

packages/api/migration/1668517874752-AddHUIDataSource.ts
echaidemenos: psql complains altering column `source` without  dropping the default value first. It is later restored:
```
await queryRunner.query(
      `ALTER TABLE "forecast_data" ALTER COLUMN "source" SET DEFAULT 'sofar_wave_model'`,
    );
```

ericboucher: Ok, thinking about it a bit more, should it even have a default value? It could be misleading if we forget to set it explicitly and then assign "sofar_wave_model". What do you think?

echaidemenos: Yes that sounds good. forecast_data are being set only in `hindcast-wind-wave-data.ts` where source is explicitly set. I am removing the default value in this migration.

packages/api/src/utils/time-series.utils.ts
```diff
@@ -253,3 +253,12 @@ export const insertSiteDataToTimeSeries = (
     .onConflict('ON CONSTRAINT "no_duplicate_data" DO NOTHING')
     .execute();
 };
+
+export const refreshMaterializedView = async (repository: Repository<any>) => {
+  const hash = (Math.random() + 1).toString(36).substring(7);
```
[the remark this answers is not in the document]
echaidemenos: Yes this is not a hash. It's purpose is to give a unique id to `console.time`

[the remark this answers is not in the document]
ericboucher: Maybe we could create a small util and name it "randomId" or something like that?

ericboucher (approved):

PR #798, opened 2022-11-23
Fix typeorm auto-generated migrations

echaidemenos:
Fixes the issue where auto-generating a migration using typeorm would create an unusable migration file. 

schema before: [aqualink_before.txt](https://github.com/aqualinkorg/aqualink-app/files/10074645/aqualink_before.txt)
schema after: [aqualink_after.txt](https://github.com/aqualinkorg/aqualink-app/files/10074643/aqualink_after.txt)

suggestions:
- add a step to circle ci checking consistency between database state and the definitions in typescript, by expecting no result from `yarn typeorm migration:generate -n MigrationName`


packages/api/src/time-series/time-series.entity.ts
```diff
@@ -15,7 +15,9 @@ import { Metric } from './metrics.entity';
 
 @Entity()
 @Unique('no_duplicate_data', ['metric', 'source', 'timestamp'])
-@Index(['metric', 'source', 'timestamp'])
+// https://github.com/typeorm/typeorm/issues/3336
+// CREATE INDEX "IDX_cb2f3e83c09f83e8ce007ffd6f" ON "time_series" ("metric", "source_id", "timestamp" DESC)
+@Index('IDX_time_series_metric_source_timestamp_DESC', { synchronize: false })
```
ericboucher, line 20: Can you explain a bit more what's happening here?

packages/api/src/sites/video-streams.entity.ts
```diff
@@ -19,7 +19,13 @@ export class VideoStream {
   @Column()
   ownerEmail: string;
 
-  @Column('point')
+  @Column({
+    type: 'geometry',
+    spatialFeatureType: 'Point',
+    unique: true,
+    srid: 4326,
+    nullable: false,
```
ericboucher, line 27: I think this should probably be nullable for now

packages/api/src/time-series/time-series.entity.ts
echaidemenos: `synchronize: false` tells typeorm to not try to change `IDX_time_series_metric_source_timestamp_DESC`. We want that because it is not yet possible to create index with `DESC` using typeorm.

packages/api/migration/1669109147094-FixTypeorm.ts
```diff
@@ -0,0 +1,679 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class FixTypeorm1669109147094 implements MigrationInterface {
+  name = 'FixTypeorm1669109147094';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // wrapping migration with DROP and CREATE of the materialized view.
+    await queryRunner.query(
+      `DELETE FROM "typeorm_metadata" WHERE "type" = $1 AND "name" = $2 AND "schema" = $3`,
```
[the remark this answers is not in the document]
echaidemenos: This format is auto-generated by `typeorm`, so i think we would like to keep this consistent

ericboucher (approved):

Issue #800, opened 2022-11-24
Get temperature data for older surveys


ericboucher:
At the moment, we use daily data to get temperature information for surveys. But this does not work if the survey is older than 3 months.

Here is an example - https://aqualink.org/sites/3031/survey_details/294

We should find a convenient way to grab temperature data for this date, either as a separate query or as part of the survey data query.

PR #803, opened 2022-12-01
Fix no satellite temperature for older surveys

echaidemenos:
Resolves https://github.com/aqualinkorg/aqualink-app/issues/800

Use api's temperature in `ObservationBox`, as we do in surveys timeline, instead of daily data.
Change `GET` `/sites/:id/surveys/:id` to include `temperature` field as we do in `GET` `/sites/:id/surveys`

### DIFF
--- a/packages/website/src/common/Chart/MultipleSensorsCharts/index.tsx
+++ b/packages/website/src/common/Chart/MultipleSensorsCharts/index.tsx
@@ -512,7 +512,6 @@ const MultipleSensorsCharts = ({
           className={classes.button}
         />
       </div>
-      {console.log(tempAnalysisDatasets)}
       <ChartWithCard
         id="temperature"
         range={range}


packages/website/src/routes/Surveys/View/SurveyDetails.tsx
```diff
@@ -85,11 +85,11 @@ const SurveyDetails = ({ site, survey, classes }: SurveyDetailsProps) => {
           )}
         </Grid>
       )}
-
       <Grid item xs={12} md={6} lg={3}>
         <ObservationBox
           depth={site.depth}
           date={survey?.diveDate}
+          satelliteTemperature={survey?.temperature}
```
ericboucher, line 92: I thought we agreed we would keep referencing to satelliteData explicitly here and not survey.temperature?

ericboucher: Maybe return a separate column instead, `survey.satelliteTemperature`?

packages/api/test/mock/surveys.mock.ts
```diff
@@ -7,7 +7,7 @@ import { siteManagerUserMock } from './user.mock';
 export const californiaSurveyOne: DeepPartial<Survey> = {
   comments: 'California Survey One',
   diveDate: moment().subtract(6, 'days').toISOString(),
-  temperature: 22,
+  temperature: null,
```
ericboucher, line 10: I was not expecting this to become null?

ericboucher (approved):

packages/website/src/common/Chart/MultipleSensorsCharts/index.tsx
```diff
@@ -512,6 +512,7 @@ const MultipleSensorsCharts = ({
           className={classes.button}
         />
       </div>
+      {console.log(tempAnalysisDatasets)}
```
ericboucher: 👣

Issue #802, opened 2022-12-01
Investigate a few bugs with collections


ericboucher:
- The https://aqualink.org/collections/heat-stress collection seems to be missing. Though it is referenced in https://aqualink.org/tracker

- Public collections don't seem to be accesible by id. Eg. minderoo has id=1, yet it is not accessible through https://aqualink.org/collections/1, only through https://aqualink.org/collections/minderoo

- Should we only create a collection when users actually add a site? And we could call it "Person's Name Collection" instead of "My Collection".

Issue #755, opened 2022-06-09
Simplify Collection Creation


ericboucher:
At the moment, we automatically create a collection for every user in our database. Which creates a lot of "noise".
We should:

- create a collection when the user saves their first site
- make the "no site" page more appealing and show where people should click, and what their dashboard will look like
- the default name for the collection should be "FirstName's Dashboard"

The current version of our "no site page":
<img width="1728" alt="Screen Shot 2022-06-09 at 11 22 12 AM" src="https://user-images.githubusercontent.com/16843267/172813674-6c8f9f85-c421-4313-990b-5e5915aa1fde.png">

PR #804, opened 2022-12-02
Improve Collections

echaidemenos:
The purpose of this PR is to resolve https://github.com/aqualinkorg/aqualink-app/issues/802 and https://github.com/aqualinkorg/aqualink-app/issues/755

This PR fixes:
- [heat-stress](https://aqualink.org/collections/heat-stress) collection not being accessible

Changes: 
- Collections are created when a user tries to add their first site. This change does not affect the user's flow.
- Default collection name now is: "User Name's collection"

Adds:
- It is now possible to access public collection throw their id
- The [/tracker](https://aqualink.org/tracker) page is now also displayed when a user either has no collection or its empty with an extra banner

Screenshot from 2022-12-09 13-16-25

Additionally we can run the following SQL code to remove unused collections:
```
delete from collection c2
	where c2.id not in  
	(select c.id from collection c inner join collection_sites_site css on c.id = css.collection_id)
	and c2."name" = 'My Dashboard'
```

### DIFF
--- a/packages/api/src/collections/collections.service.ts
+++ b/packages/api/src/collections/collections.service.ts
@@ -10,7 +10,7 @@ import { Repository, MoreThanOrEqual, In } from 'typeorm';
 import { Collection, DynamicCollection } from './collections.entity';
 import { Sources } from '../sites/sources.entity';
 import { LatestData } from '../time-series/latest-data.entity';
-import { User } from '../users/users.entity';
+import { AdminLevel, User } from '../users/users.entity';
 import { hasHoboDataSubQuery } from '../utils/site.utils';
 import { CreateCollectionDto } from './dto/create-collection.dto';
 import { FilterCollectionDto } from './dto/filter-collection.dto';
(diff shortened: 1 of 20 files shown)


packages/api/src/collections/collections.service.ts
```diff
@@ -40,8 +40,19 @@ export class CollectionsService {
...
 
-  create(createCollectionDto: CreateCollectionDto): Promise<Collection> {
-    const { name, isPublic, siteIds, userId } = createCollectionDto;
+  create(
+    createCollectionDto: CreateCollectionDto,
+    user?: User,
+  ): Promise<Collection> {
+    const { name, isPublic, siteIds, userId: idFromDTO } = createCollectionDto;
+
+    if (idFromDTO && user?.adminLevel !== AdminLevel.SuperAdmin) {
```
ericboucher, line 49: Are you sure this should be `&&`?
We also want normal users to create collections, no? Or am I missing something?

packages/website/src/routes/Dashboard/Content.tsx
```diff
@@ -27,9 +27,7 @@ const Content = () => {
   }
 
   if (collection?.sites.length === 0) {
-    return (
-      <FullScreenMessage message="There are no sites in your dashboard. Add sites to your dashboard to monitor multiple locations in a single view." />
-    );
+    return <Tracker shouldShowNav={false} />;
```
ericboucher, line 30: Would it be easy to add a banner at the top saying something like: `You have not saved any sites yet. Follow the instructions on this page and come back to your dashboard after saving a few sites!`

ericboucher (approved):

Issue #806, opened 2022-12-07
Improve data downloads


ericboucher:
As a user of the application, I would like to download all available data for my spotter/site. At the moment, we provide the following metrics:

<img width="731" alt="Screenshot 2022-12-03 at 7 29 00 PM" src="https://user-images.githubusercontent.com/16843267/206194885-37348fa2-cd29-446f-aad8-03a074fe30f5.png">

In addition, we would like to allow users to download the rest of the data available through their spotter (when available):

- wave period
- wave direction
- wind direction
- barometric pressure

And with a checkbox, allow them to download all the additional data (eg. wave data from GFS)

PR #810, opened 2022-12-12
CSV data downloads improvements

echaidemenos:
This PR resolve https://github.com/aqualinkorg/aqualink-app/issues/806

This PR adds 3 check boxes with the following options:
- Additional Data: Downloads all available metrics from all available sources for the current time window.
- All Dates: Ignores the dates from the time picker and downloads data from all available dates
- Hourly Data: Checked by default. If unchecked, data will not be averaged per hour

image
Screenshot from 2022-12-15 15-52-20
Screenshot from 2022-12-19 11-45-48

### DIFF
--- a/packages/website/src/common/Chart/MultipleSensorsCharts/DownloadCSVDialog.tsx
+++ b/packages/website/src/common/Chart/MultipleSensorsCharts/DownloadCSVDialog.tsx
@@ -37,13 +37,13 @@ const DownloadCSVDialog = ({
   const classes = useStyles();
   const [additionalData, setAdditionalData] = React.useState(false);
   const [allDates, setAllDates] = React.useState(false);
-  const [hourly, setHourly] = React.useState(true);
+  const [hourly, setHourly] = React.useState(false);
 
   const handleClose = (shouldDownload: boolean) => {
     if (shouldDownload) {
       onClose(true, additionalData, allDates, hourly);
     } else {
-      onClose(false, false, false, false);
+      onClose(false, false, false, true);
     }
   };
 
@@ -51,7 +51,7 @@ const DownloadCSVDialog = ({
     if (open) {
       setAdditionalData(false);
       setAllDates(false);
-      setHourly(true);
+      setHourly(false);
     }
   }, [open]);
 
@@ -90,6 +90,7 @@ const DownloadCSVDialog = ({
             <Tooltip
               title="Download all additional data available from all sources for this site"
               placement="top"
+              arrow
             >
               <FormControlLabel
                 control={
@@ -105,6 +106,7 @@ const DownloadCSVDialog = ({
             <Tooltip
               title="Download data for all available dates"
               placement="top"
+              arrow
             >
               <FormControlLabel
                 control={
@@ -117,10 +119,7 @@ const DownloadCSVDialog = ({
                 label="All Dates"
               />
             </Tooltip>
-            <Tooltip
-              title="If unchecked, Data will not be averaged per hour"
-              placement="top"
-            >
+            <Tooltip title="Average data per hour" placement="top" arrow>
               <FormControlLabel
                 control={
                   <Checkbox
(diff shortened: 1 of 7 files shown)


packages/website/src/common/Chart/MultipleSensorsCharts/DownloadCSVDialog.tsx
```diff
@@ -58,15 +78,60 @@ const DownloadCSVDialog = ({
...
+                    onChange={() => setAllDates(!allDates)}
+                  />
+                }
+                label="All Dates"
+              />
+            </Tooltip>
+            <Tooltip title="Download hourly data" placement="top">
+              <FormControlLabel
+                control={
+                  <Checkbox
```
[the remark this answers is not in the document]
echaidemenos: I will add this as the tooltip description so it is more explanatory. Thanks!

ericboucher:
Nice work @echaidemenos
Can you add a screenshot of the notification banner maybe? Thanks :)

ericboucher:
Let's add an (i) information tooltip next to each toolbox name so that users know what they mean exactly.

For hourly data, are we really averaging by hour or picking one data point per hour?

ericboucher:
@echaidemenos as suggested in the linked issue, can we maybe add to the bullet list that it downloads wind and wave direction as well as wave period? Those three are the last metrics that the spotter collects.

echaidemenos:
> @echaidemenos as suggested in the linked issue, can we maybe add to the bullet list that it downloads wind and wave direction as well as wave period? Those three are the last metrics that the spotter collects.

I added a list with data that may be there. barometric data will already show on the existing list, because there is a chart for it

packages/website/src/common/Chart/MultipleSensorsCharts/DownloadCSVDialog.tsx
```diff
@@ -15,9 +19,22 @@ export interface DownloadCSVDialogProps {
...
+  onClose: (
+    shouldDownload: boolean,
+    additionalData: boolean,
+    allDates: boolean,
+    hourly: boolean
+  ) => void;
 }
 
+const additionalDataLabels = [
+  "wave_mean_period_spotter",
```
ericboucher, line 31: Could we also make the standard data columns be in snake_case? At the moment they are a mix o camelCase and snake_case for some reason. Or change these additional data labels if its way easier, but at least be consistent in our naming conventions ;)

packages/website/src/common/Chart/MultipleSensorsCharts/DownloadCSVDialog.tsx
```diff
@@ -15,9 +19,22 @@ export interface DownloadCSVDialogProps {
...
+  "wind_direction_spotter",
+  "wave_mean_direction_spotter",
+  "dhw_noaa",
+  "temp_alert_noaa",
```
ericboucher, line 35: Let's add a bullet point showing `"additional site-specific data"`

packages/website/src/common/Chart/MultipleSensorsCharts/DownloadCSVDialog.tsx
echaidemenos: I will change them to snake_case, since that is how they are printed in the csv

packages/website/src/common/Chart/MultipleSensorsCharts/DownloadCSVDialog.tsx
```diff
@@ -143,22 +143,26 @@ const DownloadCSVDialog = ({
...
+                <li>
+                  additional site-specific data:
+                  <ul>
+                    {additionalDataLabels.map((x) => (
+                      <li key={x}>{x}</li>
+                    ))}
+                  </ul>
+                </li>
+              )}
             </ul>
```
ericboucher, line 164: This is not what I meant. I meant that in the list of additional data, we should also add another list item that says `additional site-specific data when available`. For example in the case of sites with HOBO data, users would receive this data as well in the CSV. Correct?

ericboucher (approved):

PR #812, opened 2022-12-15
Reduce number of requests in scheduledSSTTimeSeriesUpdate

echaidemenos:
The purpose of this PR is to solve an issue where not all sites would get SST updates daily from the functions.
We suspect that sofar API fails to answer the big amount of the functions' requests.
This PR combines the 4 per site requests to 1 per site, by combining the start and end API parameters.
It also cleans up the functions logs so they are more useful.

Also since the sofar hindcast gives only one value per day, we could reduce the `scheduledSSTTimeSeriesUpdate` function calls from 1 per hour to one per day.

### DIFF
--- a/packages/api/scripts/backfill-sofar-time-series.ts
+++ b/packages/api/scripts/backfill-sofar-time-series.ts
@@ -84,7 +84,6 @@ async function run() {
   return fn(
     parsedSiteIds,
     days,
-    connection,
     // Fetch all needed repositories
     {
       siteRepository: connection.getRepository(Site),
--- a/packages/api/src/utils/sst-time-series.ts
+++ b/packages/api/src/utils/sst-time-series.ts
@@ -1,8 +1,8 @@
 import { Logger } from '@nestjs/common';
 import Bluebird from 'bluebird';
-import { Connection, In, Repository } from 'typeorm';
+import { In, Repository } from 'typeorm';
 import { Point } from 'geojson';
-import { isNil, times } from 'lodash';
+import { flatten, groupBy, isNil, times } from 'lodash';
 import moment from 'moment';
 
 import { Site } from '../sites/sites.entity';
@@ -49,13 +49,11 @@ const getSites = (siteIds: number[], siteRepository: Repository<Site>) => {
  * and save all above metrics to time_series table
  * @param siteIds The siteIds for which to perform the update
  * @param days How many days will this script need to backfill (1 = daily update)
- * @param connection An active typeorm connection object
  * @param repositories The needed repositories, as defined by the interface
  */
 export const updateSST = async (
   siteIds: number[],
   days: number,
-  connection: Connection,
   repositories: Repositories,
 ) => {
   const { siteRepository, timeSeriesRepository, sourceRepository } =
@@ -72,6 +70,8 @@ export const updateSST = async (
     }),
   );
 
+  logger.log(`Back-filling ${sources.length} sites`);
+
   await Bluebird.map(
     sources,
     async (source) => {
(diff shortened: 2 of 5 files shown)


packages/api/src/utils/sst-time-series.ts
```diff
@@ -82,89 +82,105 @@ export const updateSST = async (
...
-              getLatestData(sstFiltered)?.value,
-              // Calculate degree heating days
-              latestDhw && latestDhw.value * 7,
-            );
-
-            // Calculate the sstAnomaly
-            const sstAnomaly = sstFiltered
+      const endDate = moment().format();
+      const startDate = moment()
+        .subtract(days - 1, 'd')
```
ericboucher, line 87: I think we can take:
- `.endOf('day)` for the endDate and
- and simply `.subtract(days, 'd')` for startDate

packages/api/src/utils/sst-time-series.ts
```diff
@@ -82,89 +82,105 @@ export const updateSST = async (
...
-              site.maxMonthlyMean,
-              getLatestData(sstFiltered)?.value,
-              // Calculate degree heating days
-              latestDhw && latestDhw.value * 7,
-            );
-
-            // Calculate the sstAnomaly
-            const sstAnomaly = sstFiltered
+      const endDate = moment().format();
+      const startDate = moment()
```
ericboucher, line 86: Is there a limit at how many days the SofarAPI will return?

packages/api/src/utils/sst-time-series.ts
```diff
@@ -82,89 +82,105 @@ export const updateSST = async (
...
+        alert: alertLevel,
+      };
+
+      if (!data.sst.length || data.sst.length === 0) {
+        console.error(`No Hindcast data available for site '${site.id}'`);
+      }
+
       return Bluebird.map(
-        data,
+        [data],
```
ericboucher, line 183: Do we still need the `bluebird.map` this change makes it feel like we don't?

packages/api/scripts/backfill-sofar-time-series.ts
```diff
@@ -84,7 +84,6 @@ async function run() {
   return fn(
     parsedSiteIds,
     days,
-    connection,
```
ericboucher, line 87: Why this change?

ericboucher (approved):
Looks great overall, thanks for looking into this! I left a few small comments for potential improvements inline

packages/api/src/utils/sst-time-series.ts
echaidemenos: `endDate` must be in the past, so `.endOf('day')` wont work.
Ok about `startDate`.

packages/api/src/utils/sst-time-series.ts
echaidemenos: turns out that there is a 7 day limit. I will have to keep some of the old logic to take that into account for the backfill script. The function will run OK since it only need 4 days

packages/api/src/utils/sst-time-series.ts
```diff
@@ -82,89 +82,105 @@ export const updateSST = async (
...
       );
 
+      const data = {
+        sst: sstFiltered,
+        dhw: dhwFiltered,
+        sstAnomaly: anomaly,
+        alert: alertLevel,
+      };
+
+      if (!data.sst.length || data.sst.length === 0) {
```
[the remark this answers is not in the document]
echaidemenos: `sofarHindcast` already does error handling. This is in case the values are `null`, which is the case of the point not being in the availability zone.

packages/api/scripts/backfill-sofar-time-series.ts
echaidemenos: It was left there unneeded form a previous PR. It is not being used so I removed it

packages/api/src/utils/sst-time-series.ts
```diff
@@ -82,89 +82,105 @@ export const updateSST = async (
...
+            site.maxMonthlyMean,
+            latest?.value,
+            // Calculate degree heating days
+            dhw && dhw.value * 7,
+          );
+          if (!alert) return undefined;
+          return {
+            value: alert,
+            timestamp: latest?.timestamp,
+          } as ValueWithTimestamp;
```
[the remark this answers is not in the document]
echaidemenos: Thanks a lot for the detailed comment! You are right setting `if (!latest) return undefined;` is the better approach here.
`if (!alert) return undefined;` actually implies also `latest` not being nullish, since they are linked by the logic above

packages/api/src/utils/sst-time-series.ts
```diff
@@ -82,89 +82,105 @@ export const updateSST = async (
...
+          startDate,
+          endDate,
+        ),
+      ]);
+
+      // Filter out null values
+      const sstFiltered = filterSofarResponse(SofarSSTRaw);
+      const dhwFiltered = filterSofarResponse(sofarDegreeHeatingWeekRaw);
+
+      const getDateNoTime = (x?: string) => new Date(x || '').toDateString();
```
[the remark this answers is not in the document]
echaidemenos: Thanks for the comment. Yes this is the desired behavior. `Invalid Date` is still a `Date` object and `toDateString` it's function. 
Although since you noted that i noticed that there is a chance `groupedDHWFiltered` and `groupedSSTFiltered` getting the key `Invalid Date`. I will make sure to have that key and its properties removed.

packages/api/src/utils/sst-time-series.ts
ericboucher: Ah I see, I thought endDate could be an input as well. Then it makes sense.

PR #834, opened 2023-04-11
Make .prettierrc common for api and website

echaidemenos:
This PR makes `.prettierrc` common for both api and website packages. The inspiration behind this it to keep the styling consistent between components


echaidemenos:
@ericboucher should we continue with this?

.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+8a264fd6bf9e6099c245e612ae3826c4441f803d
```
ericboucher, line 1: Maybe add a quick comment for future ref?

ericboucher (approved):

Issue #407, opened 2021-01-06
Add a script to update "deployed" status of buoys regularly


ericboucher:
After #406 is complete:

Create a simple function to update the `deployed` status (eg. during the daily update)
A site is considered deployed if it received spotter data within the past two days and the last location of the spotter is within 50km of the site location

Issue #620, opened 2021-08-12
Retire daily data


avalmas-programize:
The `daily_data` table is slowly getting replaced by the new `time_series` table.
However there are still many instances where `daily_data` are used instead of the `time_series` in the backend.
We need to clear out those first and then make sure that the front-end uses only `daily_data`.
Finally we should stop the dailyDataUpdate and retire the `daily_data`.
We can keep a backup of the data of the table, before removing it, in case we might need them.

PR #838, opened 2023-04-18
Add daily buoy status check

echaidemenos:
Resolves https://github.com/aqualinkorg/aqualink-app/issues/407

This PR adds a new cloud function, checking if any data ware received recently (in 2 days period) and sent slack a notification with the site ids whose spotters received no data. 

Also adds a check in `addSpotterData` for the distance between the spotter and the site. 
At the moment `daily_data` will still update without considering the new check, but since https://github.com/aqualinkorg/aqualink-app/issues/620 will be soon implemented and the change will not brake any feature, this should not be a problem


packages/api/src/utils/spotter-time-series.ts
```diff
@@ -152,9 +157,34 @@ export const addSpotterData = async (
...
+            endDate,
+            startDate,
+          ).then((data) => excludeSpotterData(data, sensorExclusionDates));
+
+          if (
+            !skipDistanceCheck &&
+            spotterData?.latitude?.length &&
+            spotterData?.latitude[0].value &&
+            spotterData?.longitude?.length &&
+            spotterData?.longitude[0].value
```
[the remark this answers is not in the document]
ericboucher: This can probably be simplified quite a bit imho

packages/api/src/utils/spotter-time-series.ts
```diff
@@ -152,9 +157,34 @@ export const addSpotterData = async (
...
+          ) {
+            // Check if spotter in within a specified distance from the site, else don't return any data.
+            const dist = distance(
+              (site.polygon as Point).coordinates,
+              [spotterData.longitude[0].value, spotterData.latitude[0].value],
+              { units: 'kilometers' },
+            );
+            if (dist > MAX_DISTANCE_FROM_SITE) {
+              logger.warn(
+                `Spotter is over ${MAX_DISTANCE_FROM_SITE}km from site ${site.id}`,
```
ericboucher, line 181: please indicate in the message that the data is not saved

packages/api/src/utils/spotter-time-series.ts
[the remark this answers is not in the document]
echaidemenos: indeed they are not, i will be removing `spotterData?.longitude[0].value`

packages/api/scripts/backfill-sofar-time-series.ts
```diff
@@ -38,6 +38,12 @@ const { argv } = yargs
     describe: 'The sites that should be backfilled with spotter data',
     type: 'array',
   })
+  .option('c', {
```
ericboucher, line 41: I'm being pick but I feel like `c` would mean "check" and not "skip check"
Maybe change this to `nc` "no-check"? Wdyt?

packages/api/src/workers/check-buoys-status.ts
```diff
@@ -0,0 +1,78 @@
...
+          text: `Spotters of site(s) ${diff.join(
+            ', ',
+          )} have not received data for more than 2 days!`,
+        },
+      },
+    ],
+  };
+
+  // Log message in stdout
+  logger.log(messageTemplate);
```
ericboucher, line 64: Should this be a logger.warn or logger.error?

echaidemenos: There is no "error" with the function if there are such sites. The "error" will appear in slack, so i will make it a warning. Does that make sense @ericboucher ?

ericboucher (approved):

Issue #850, opened 2023-05-04
Connect non-aqualink spotters


ericboucher:
As the owner of a Sofar spotter, I would like my data to be uploaded to Aqualink.

If there isn't an Aqualink spotter configured, we can add the spotterId and SofarAPIToken fields in the edit site details section and display. We could indicate that we only support SofarOcean spotters for now but they can reach out to us if they would like to connect other live buoys.

We should log changes to these fields in the `site_audit` table 

In the backend, let's find a place for tokens. It could be a new column in `sites` or `sources`, or a new table altogether. For now, I think a one-to-one relationship is ok. We can change this in the future if we have many organizations with many spotters.
 
Then during data fetching, we simply grab the associated token if there is one, or default to the Aqualink token. It might be worth adding better error-handling around `access forbidden` errors. Maybe send a Slack message?

PR #851, opened 2023-05-04
Add option to use different sofar tokens for spotters

echaidemenos:
Resolves https://github.com/aqualinkorg/aqualink-app/issues/850

In summary this PR adds `sofar_api_token` column to `public.sites` and uses it instead of env variable `SOFAR_API_TOKEN`, when available.

`site_audit` table is updated with every change of `sofar_api_token` and a slack notification is sent in case 403 or 401 is received from Sofar API, as stated in the issue 

`sofar_api_token` can not be obtained from the API, it can only be updated.

Add new UI for updating this field as shown bellow, in `EDIT SITE DETAILS` menu

Screenshot from 2023-05-04 19-32-44
Screenshot from 2023-05-04 19-32-50


packages/api/src/sites/dto/update-site.dto.ts
```diff
@@ -68,4 +68,18 @@ export class UpdateSiteDto {
...
   readonly streamId?: number;
+
+  @ApiProperty({ example: 'SPOT-1742' })
+  @IsOptional()
+  @IsString()
+  @IsNotEmpty()
+  @MaxLength(100)
+  readonly sensorId?: string;
+
+  @ApiProperty({ example: 'jl3Xr1kZeqDqs7KAiktXOyr3PlB5Ip' })
```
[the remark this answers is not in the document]
ericboucher: Yes

packages/api/migration/1683214698387-AddSofarAPITokenToSite.ts
```diff
@@ -0,0 +1,131 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddSofarAPITokenToSite1683214698387 implements MigrationInterface {
+  name = 'AddSofarAPITokenToSite1683214698387';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "site" ADD "sofar_api_token" character varying`,
```
ericboucher, line 8: is this protected to only be read by admins?

ericboucher: Ok great, can we just rename this to `spotter_api_token` instead to make it more future-proof?

packages/api/src/utils/hindcast-wind-wave.ts
```diff
@@ -50,7 +49,11 @@ export const addWindWaveData = async (
 ) => {
   logger.log('Fetching sites');
   // Fetch all sites
-  const sites = await getSites(siteIds, false, repositories.siteRepository);
+  const sites = await repositories.siteRepository.find({
```
ericboucher, line 52: Why not update add the functionality you need in `getSites` and use this instead?

packages/api/functionsConfig.ts
```diff
@@ -25,6 +25,9 @@ export const functionsConfig = {
   database: {
     url: process.env.DATABASE_URL,
   },
+  general: {
+    deploy_env: process.env.DEPLOY_ENV,
```
ericboucher, line 29: does the environment variable already exist?

packages/api/cloud-functions/index.ts
```diff
@@ -149,11 +149,12 @@ exports.scheduledSpotterTimeSeriesUpdate = functions
   .retryConfig({ retryCount: 2 })
   .onRun(async () => {
     process.env.SOFAR_API_TOKEN = functions.config().sofar_api.token;
+    const isNotProd = functions.config().general.deploy_env !== 'prod';
```
ericboucher, line 152: If you're only using this here, let's call it: `skipDistanceCheck` directly for clarity and add a quick comment

packages/api/src/utils/hindcast-wind-wave.ts
echaidemenos: This was the definition of `getSites`
```
export const getSites = (
  siteIds: number[],
  hasSensorOnly: boolean = false,
  siteRepository: Repository<Site>,
) => {
  return siteRepository.find({
    where: {
      ...(siteIds.length > 0 ? { id: In(siteIds) } : {}),
      ...(hasSensorOnly ? { sensorId: Not(IsNull()) } : {}),
    },
  });
};
```
It was used 2 times, and with different parameters. So in reality it was adding complexity to code readability, by having to keep in mind one more function, when it can just be replaced with a straightforward (and also) single call. The extra parameter in this PR would further increase the semantic difference between the only 2 appearances of that function.

packages/api/functionsConfig.ts
ericboucher: This won't work as cloud functions don't use the environment variables that are set locally. They need to be set manually for each environment with the `config:cloud-functions` command line. Could we use another environment variable such as FirebaseId or BACKEND_BASE_URL ?

ericboucher (approved):

Issue #845, opened 2023-04-26
Control that the API and DB are both running with the same TZ setting


ericboucher:
Failing local tests on https://github.com/aqualinkorg/aqualink-app/pull/833 made it apparent that we have to be very careful with how we handle dates.

Our tests started failing when the DB and Nest were not using the same TZ (timezone) setting. Ideally, we would want to make this less of an issue and find out why this happened.

As an alternative fix, we could control on start that they are both using the same TZ setting, or force both to use `TZ=UTC`

See https://www.appsloveworld.com/postgresql/100/237/typeorm-find-methods-returns-wrong-timestamp-time

PR #853, opened 2023-05-12
Set timezone to UTC in tests

echaidemenos:
Resolves https://github.com/aqualinkorg/aqualink-app/issues/845

The timezone offset in both the actual nest app and the scripts is 0, but that is not the case in test. That was causing the tests to fail locally for anyone with no UTC time, while the app would appear to be working as expected. This change sets the timezone offset to 0 (UTC), for everyone, to match the nest app behavior.

### DIFF
--- a/packages/api/test/global-setup.js
+++ b/packages/api/test/global-setup.js
@@ -1,0 +1,3 @@
+module.exports = async () => {
+  process.env.TZ = 'UTC';
+};
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -142,23 +142,6 @@
     "webpack-cli": "^3.3.12",
     "yargs": "^15.4.1"
   },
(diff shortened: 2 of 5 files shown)


packages/api/test/global-setup.js
```diff
@@ -0,0 +1,3 @@
+module.exports = async () => {
```
ericboucher, line 1: Is there a way to enforce this setup for the API and DB as well?

echaidemenos: @ericboucher turns out it is already set for the API in the `.env` https://github.com/aqualinkorg/aqualink-app/blob/c53627e6891e6ed6fa56ef6430c9ff75e1560a82/packages/api/.env.example#L19

ericboucher (approved):

PR #863, opened 2023-05-29
Update condition where sonde card appears 

echaidemenos:
This PR updates the condition where the sonde `WaterSamplingCard` card appears.
Changes the condition from "there is `salinity` in latest data" to "at least n # of `sondeMetrics` are in latest data"


Caesarh97 (commented):
This looks good to me.

ericboucher (approved):

PR #866, opened 2023-06-06
Remove webpack for functions deployment

echaidemenos:
This PR removes the need of `webpack` for the firebase functions deployment.
Webpack 4 is not supported by node 18 and webpack 5 introduced some issues to our already "hacky" building solution.
Now the build is more straightforward using just `tsc`.


firebase.json
```diff
@@ -0,0 +1,26 @@
+{
+  "functions": {
+    "source": ".",
+    "codebase": "default",
+    "ignore": [
```
ericboucher, line 5: Could we use an "incldue" logic instead?

ericboucher:
It seems odd that we actually have to move everything to the top level. Was there no way to keep this in an independent folder?

packages/api/.env.example
```diff
@@ -5,9 +5,9 @@ POSTGRES_HOST=localhost
 POSTGRES_PORT=54321
 POSTGRES_DATABASE=postgres
 TEST_POSTGRES_DATABASE=test_ovio
-FIREBASE_PROJECT_ID=
-FIREBASE_PRIVATE_KEY=
-FIREBASE_CLIENT_EMAIL=
+FB_PROJECT_ID=
```
ericboucher, line 8: why this change?

echaidemenos: This was done due to this note: https://firebase.google.com/docs/functions/config-env?gen=2nd#reserved-names
but it looks like Firebase is not complaining, maybe because the `.env` is under `./packages/api` and not under `./` (where `firebase.json` is located).
I will be reverting the change.

ericboucher: firebase is not using the .env, we are setting the environment variables through `config:cloud-functions`

packages/api/cloud-functions/index.ts
```diff
@@ -14,10 +14,47 @@ import { sendSlackMessage } from '../src/utils/slack.utils';
 import { checkBuoysStatus } from '../src/workers/check-buoys-status';
 import { dataSourceOptions } from '../ormconfig';
 
-// We have to manually import all required entities here, unfortunately - the globbing that is used in ormconfig.ts
-// doesn't work with Webpack. This declaration gets processed by a custom loader (`add-entities.js`) to add import
-// statements for all the entity classes.
-// import-all-entities
+import * as SiteAuditEntity from '../src/audit/site-audit.entity';
```
ericboucher, line 17: Can't we keep using a formulae like we had in `[add-entities.js](https://github.com/aqualinkorg/aqualink-app/pull/866/files#diff-7f237b5a6c3cf6ad00f2b425ac060371215e0c3f6f6ec9e59395ada53b867d37)`?

packages/website/package.json
```diff
@@ -91,7 +91,7 @@
     "@types/immutable": "^3.8.7",
     "@types/jest": "^24.0.0",
     "@types/lodash": "^4.14.149",
-    "@types/node": "^20.2.3",
+    "@types/node": "^18.2.3",
```
ericboucher, line 94: I think [18.16.16](https://www.npmjs.com/package/@types/node/v/18.16.16) is the latest version

packages/api/cloud-functions/index.ts
ericboucher: I'd much rather have it automated and not have to think about exporting entities here whenever we add a new one.

firebase.json
echaidemenos: unfortunately there isn't

packages/api/cloud-functions/index.ts
echaidemenos: I have added some per-prossesing, almost identical to what we did before, let me know what you think

package.json
```diff
@@ -46,7 +47,7 @@
     "typescript": "^4.5.5"
   },
   "engines": {
-    "node": "18.x",
+    "node": "18",
```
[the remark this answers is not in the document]
echaidemenos: Hello! That is not true as far as i know and tested. `18` will match any node 18 version, just as `18.x` would do. The change here is because firebase requires the simple `18` format: https://firebase.google.com/docs/functions/manage-functions?gen=2nd#set_nodejs_version

PR #868, opened 2023-06-08
Fix functions would not read SOFAR_API_TOKEN

echaidemenos:
Fixes a bug where in some cases functions would not read `SOFAR_API_TOKEN`.

The issue was that `export const { SOFAR_API_TOKEN } = process.env;` in `constants.ts` would run before `process.env.SOFAR_API_TOKEN = functions.config().sofar_api.token;` in `cloud-functions/index.ts` causing imported `SOFAR_API_TOKEN` to be `undefined`.

### DIFF
--- a/packages/api/src/utils/constants.ts
+++ b/packages/api/src/utils/constants.ts
@@ -8,8 +8,7 @@ try {
 export const envName = process.env.NODE_ENV || 'development';
 export const isTestEnv = envName === 'test';
 
-// Sofar API urls and token
-export const { SOFAR_API_TOKEN } = process.env;
+// Sofar API urls
 export const SOFAR_MARINE_URL =
   'https://api.sofarocean.com/marine-weather/v1/models/';
 export const SOFAR_WAVE_DATA_URL = 'https://api.sofarocean.com/api/wave-data';
@@ -55,7 +54,3 @@ export const sofarVariableIDs = {
 };
 
 export const STORM_GLASS_BASE_URL = 'https://api.stormglass.io/v2';
-
-export const { SLACK_BOT_TOKEN, SLACK_BOT_CHANNEL } = !isTestEnv
-  ? process.env
-  : { SLACK_BOT_TOKEN: undefined, SLACK_BOT_CHANNEL: undefined };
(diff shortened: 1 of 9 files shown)


.github/workflows/build_website.yml
```diff
@@ -61,8 +61,8 @@ jobs:
         working-directory: packages/website
         run: yarn build
 
-      - name: Deploy on surge
-        if: steps.build_website.outcome == 'success'
+      - name: Deploy pull request on surge
+        if: steps.build_website.outcome == 'success' && github.event.pull_request.number != null
```
echaidemenos, line 65: @ericboucher just added this change here to (hopefully) skip deploying on surge when merging to master. We could also split the workflow and create one for deploying and not run it when pushing to master. Tell me what you think and if we want this

packages/api/src/utils/constants.ts
```diff
@@ -8,8 +8,12 @@ try {
 export const envName = process.env.NODE_ENV || 'development';
 export const isTestEnv = envName === 'test';
 
-// Sofar API urls and token
-export const { SOFAR_API_TOKEN } = process.env;
+// Environment variables (especially those used by cloud-functions)
+// should NOT be exported from here (eg. "export const { SOFAR_API_TOKEN } = process.env"),
+// since it will interfere with the way they are set in cloud-functions,
+// causing them to be undefine.
```
ericboucher, line 14: `undefined`

.github/workflows/build_website.yml
ericboucher: Not sure if it's necessary but it probably doesn't hurt

ericboucher (approved):

.github/workflows/build_website.yml
echaidemenos: next steps after deploying on surge is the PR commenter

.github/workflows/build_website.yml
```diff
@@ -61,8 +61,8 @@ jobs:
...
         uses: dswistowski/surge-sh-action@v1
```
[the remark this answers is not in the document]
ericboucher: I opened issue #870

Issue #871, opened 2023-06-13
Remove relative file paths on imports


echaidemenos:
Consider removing relative file paths on imports by using a config file.
https://www.amitmerchant.com/avoid-relative-import-paths-in-type-script/

as stated in: https://github.com/aqualinkorg/aqualink-app/pull/869#discussion_r1227318255

PR #873, opened 2023-06-13
Remove relative file paths on imports

echaidemenos:
resolves https://github.com/aqualinkorg/aqualink-app/issues/871

Basically adds:
```
"compilerOptions": {
    ...
    "baseUrl": "./src"
  }
```
to `tsconfig.json` to enable us to import directly from `src` instead of using relative paths 
(eg. `../../../store/Survey/surveySlice` => `store/Survey/surveySlice`)


echaidemenos:
@ericboucher I canceled [pullrequest](https://github.com/apps/pullrequest) review due to the PR size.

Issue #861, opened 2023-05-24
Get SeaSurfaceTemperature from spotters


ericboucher:
Most spotters also register sea surface temperature. We would like to download it as well and save it into our database. When possible, we should display it as well in the frontend.

I think we can use the latest data call and just enable SST.
https://docs.sofarocean.com/spotter-and-smart-mooring/spotter-data/latest-data
![image](https://github.com/aqualinkorg/aqualink-app/assets/16843267/c51bf5f7-783e-4eaa-9368-4f2570c751c1)

PR #874, opened 2023-06-19
Add spotter surface_temperature

echaidemenos:
resolves https://github.com/aqualinkorg/aqualink-app/issues/861

- adds `surface_temperature` metric
- swaps `sofarSensor` with the more light `sofarLatest`, when looking for spotter position 
- show spotter surface temperature to spotter card and add relative chart, if the metric is available

Screenshot from 2023-06-23 17-00-30

Screenshot from 2023-06-23 16-54-42


ericboucher:
Let's add some tests, `SPOT-1576` should have access to this data.

PR #881, opened 2023-06-22
Fix graphs' data would not show for selected dates

echaidemenos:
This PR changes the condition of showing graph data from: "source exists in latest data" to "source has data for selected dates range"

Also changes the latest data query to include sonde and hui data for 180 days (90 days before and no hui)

### DIFF
--- a/packages/api/migration/1687443367647-ExtendLatestDataInterval.ts
+++ b/packages/api/migration/1687443367647-ExtendLatestDataInterval.ts
@@ -1,0 +1,49 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ExtendLatestDataInterval1687443367647
+  implements MigrationInterface
+{
+  name = 'ExtendLatestDataInterval1687443367647';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DELETE FROM "typeorm_metadata" WHERE "type" = $1 AND "name" = $2 AND "schema" = $3`,
+      ['MATERIALIZED_VIEW', 'latest_data', 'public'],
+    );
+    await queryRunner.query(`DROP MATERIALIZED VIEW "latest_data"`);
+
+    await queryRunner.query(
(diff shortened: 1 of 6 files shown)


packages/api/migration/1687443367647-ExtendLatestDataInterval.ts
```diff
@@ -0,0 +1,49 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
ericboucher, line 1: Let's add a comment explaining that we are extending it to 180 days

packages/api/src/time-series/latest-data.entity.ts
```diff
@@ -33,7 +33,10 @@ import { Metric } from './metrics.enum';
         .where("timestamp >= current_date - INTERVAL '7 days'")
         // Look a bit further in the past for sonde data
         .orWhere(
-          "type IN ('sonde') AND (timestamp >= current_date - INTERVAL '90 days')",
+          "type IN ('sonde') AND (timestamp >= current_date - INTERVAL '180 days')",
+        )
+        .orWhere(
+          "type IN ('hui') AND (timestamp >= current_date - INTERVAL '180 days')",
```
ericboucher, line 39: do we have other types of data that could end up being needed for the cards?

packages/api/migration/1687443367647-ExtendLatestDataInterval.ts
echaidemenos: something like:
```
// Extending latest data range for some sources, to show water sampling cards in more cases
```
is ok?

packages/api/src/time-series/latest-data.entity.ts
echaidemenos: maybe the newly added `sheet_data` and `metlog`. Should we add them?

ericboucher: yes maybe sheet_data to be consistent?

packages/api/migration/1687443367647-ExtendLatestDataInterval.ts
ericboucher: water sampling card is specific to sonde data. Maybe something like this?

`// Extending the latest data range for custom sources, to show the additional metrics card in more cases`

packages/api/src/time-series/latest-data.entity.ts
echaidemenos: Ok i am adding `sheet_data`

ericboucher (approved):

packages/website/src/common/Chart/MultipleSensorsCharts/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,869 @@
...
+                    <muipickersutilsprovider>
+                      <keyboarddatepicker
+                        class="DatePicker-textField-24"
+                        format="MM/dd/yyyy"
+                        helpertext=""
+                        inputprops="[object Object]"
+                        inputvariant="standard"
+                        keyboardbuttonprops="[object Object]"
+                        keyboardicon="[object Object]"
+                        maxdate="2023/06/28"
```
ericboucher, line 374: @echaidemenos this is changing every day and blocks tests. Please update the mocks to fix this

Issue #875, opened 2023-06-21
Failed to update site info when site has a spotter


ericboucher:
Trying to edit location for site https://aqualink.org/sites/3400 results in "something went wrong". It looks like the put endpoint is expecting a `sensorId` but the frontend is not passing it. The `sensorId` probably shouldn't be required and we should simply update the fields we are passing data for.

<img width="694" alt="Screenshot 2023-06-21 at 1 54 17 PM" src="https://github.com/aqualinkorg/aqualink-app/assets/16843267/7b191436-c5ae-45ee-a554-54137fd765dd">

PR #886, opened 2023-06-23
Fix could not update site info when site has a spotter

echaidemenos:
resolves https://github.com/aqualinkorg/aqualink-app/issues/875


ericboucher (approved):

packages/website/src/routes/SiteRoutes/Site/SiteInfo/EditForm/index.tsx
```diff
@@ -102,7 +102,7 @@ const EditForm = ({
         },
         name: siteName.value,
         depth: parseInt(siteDepth.value, 10),
-        sensorId: siteSensorId.value,
+        sensorId: siteSensorId.value || undefined,
```
ericboucher, line 105: I thought the error was happening on the API side, no? Aren't we already getting undefined here?

ericboucher: Maybe add a comment that we pass `undefined` instead of `""`?

Issue #880, opened 2023-06-22
Water sampling cards show no data


echaidemenos:
water sampling cards currently depend on latest data to pick up information. But when someone select a different date range with the date pickers, i believe it is confusing not to show eg. sonde data in water sampling card, but only show in graphs

PR #901, opened 2023-07-06
Refactor water sampling card

echaidemenos:
resolves https://github.com/aqualinkorg/aqualink-app/issues/880

This PR merges old water sampling card (SONDE) with HUI card.
Now SONDE and HUI cards show data from the latest sonde data upload.
SONDE card shows mean values, where HUI card shows geometric means.
The goal here is to make water sampling card more modular, in order to accept more metrics from different source, since we also enabled uploading any metric form any source here: https://github.com/aqualinkorg/aqualink-app/pull/854

For next steps, we could form some water sampling metric preference list and show the first 6 metrics that do exists in the latest data upload. We could also implement a switch or a per metric preference for choosing between geometric or arithmetic mean

(also fixes some visual bug in buoy observation card)


ericboucher (approved):

tsconfig.json
```diff
@@ -20,6 +20,7 @@
     "jsx": "react",
     "strictPropertyInitialization": false,
     "experimentalDecorators": true,
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "strictNullChecks": true
```
ericboucher, line 24: Where is this needed?

PR #906, opened 2023-07-20
Make survey_id not null in time-series

echaidemenos:
This PR makes `survey_id` not NULL in `time-series` table.

`time-series` rows with NULL values for `survey_id` are meaningless for the application, since they can not be related to any site.

### DIFF
--- a/packages/api/migration/1689848494071-MakeTimeSeriesSourceNotNull.ts
+++ b/packages/api/migration/1689848494071-MakeTimeSeriesSourceNotNull.ts
@@ -1,0 +1,46 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class MakeTimeSeriesSourceNotNull1689848494071
+  implements MigrationInterface
+{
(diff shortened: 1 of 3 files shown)


packages/api/migration/1689848494071-MakeTimeSeriesSourceNotNull.ts
```diff
@@ -6,6 +6,9 @@ export class MakeTimeSeriesSourceNotNull1689848494071
   name = 'MakeTimeSeriesSourceNotNull1689848494071';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
+    // lock time_series so we can safely execute DDL operations
+    await queryRunner.query('LOCK TABLE time_series IN ACCESS EXCLUSIVE MODE');
```
ericboucher, line 10: Just making sure this will automatically release access once the migration is done? No need for a release?

ericboucher (approved):

Issue #926, opened 2023-09-21
Alert level seems off


ericboucher:
Alert level seems off on some sites (no alert when the DHW is super high)

Eg. https://aqualink.org/sites/2904
<img width="1724" alt="Screenshot 2023-09-21 at 12 19 43 PM" src="https://github.com/aqualinkorg/aqualink-app/assets/16843267/b84de80e-02f6-4a72-b141-621eaface051">

ericboucher:
After investigation, I think it's because we are using `dailyData`.weeklyAlertLevel (and `dailyDaya` was having some issues that day). We could maybe use `latestData` instead.

ericboucher:
This will be partially resolved by #927 
However, we should still investigate why the dailyData values are wrong. It looks like the daily alert calculation is not working properly in the backend.

PR #927, opened 2023-09-21
Use latest data for coral bleaching alert level

ericboucher:
Linked to #926

### DIFF
--- a/packages/website/src/common/SiteDetails/index.tsx
+++ b/packages/website/src/common/SiteDetails/index.tsx
@@ -121,6 +121,7 @@ const SiteDetails = ({
   const { videoStream } = site || {};
 
   const cards =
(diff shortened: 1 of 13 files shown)


packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -121,6 +121,7 @@ const SiteDetails = ({
   const { videoStream } = site || {};
 
   const cards =
+    // should we test for `latestDataAsSofarValues` instead of `latestData`?
```
ericboucher, line 124: any thoughts on this @echaidemenos ?

echaidemenos: I believe yes we should. `latestDataAsSofarValues` should be defined if `latestData` is, but it looks like we left that check unchanged when we added the `latestDataAsSofarValues` variable

PR #935, opened 2023-10-12
Fix console warnings on tests 

echaidemenos:
resolves https://github.com/aqualinkorg/aqualink-app/issues/894

Fixes most console warnings, that appear when running the tests.

`Material-UI: The `fade` color utility was renamed to `alpha` to better describe its functionality.` cannot be straightforwardly resolved due to the outdated package `@material-ui/pickers` causing it.

TODO:
- [ ]  `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()`


packages/website/src/common/SurveyForm/index.tsx
```diff
@@ -61,7 +61,7 @@ function SurveyForm({ siteId, timeZone, onSubmit }: SurveyFormProps) {
   const [weather, setWeather] =
     useState<SurveyData['weatherConditions']>('calm');
   const itemsSize = isMobile ? 'small' : 'medium';
-  const iconSize = isMobile ? 'small' : 'default';
+  const iconSize = isMobile ? 'small' : 'medium';
```
ericboucher, line 64: If these two are the same maybe merge them?

ericboucher (approved):

packages/website/src/common/SurveyForm/index.tsx
echaidemenos: Yes didn't notice i will update it

ericboucher (approved):
@echaidemenos there are still errors when looking at the tests in
https://github.com/aqualinkorg/aqualink-app/actions/runs/6497769899/job/17647592203?pr=935

Issue #940, opened 2023-10-22
HOBO loggers. Display depth and add multiple to graph


Caesarh97:
Aqualink would like to investigate what our possibilities are with HOBO loggers. We would like to display the depth and add multiple loggers to the same graph. For example, on site [https://aqualink.org/sites/3071?end=2019-10-19&start=2018-09-18](https://aqualink.org/sites/3071?end=2019-10-19&start=2018-09-18), could we add all the HOBO loggers for all the “colonies” to the “temperature analysis” graph and display each HOBO logger and its depth? Imagine a case where we have multiple HOBO loggers deployed in different depths but at the same survey site. Could we display the loggers' depths similarly to the attached image? 

<img width="222" alt="Screenshot 2023-10-22 at 5 06 14 PM" src="https://github.com/aqualinkorg/aqualink-app/assets/113801743/46ffaf67-9fe4-484f-99eb-d03ee7a654dd">

Please let Caesar know if we're able to add these features to Aqualink. If we are able to add them, he will discuss it with the users who are interested in this before we need to spend too much time building it.

PR #941, opened 2023-10-25
Allow multiple sensor depths

echaidemenos:
resolves https://github.com/aqualinkorg/aqualink-app/issues/940

Changes the condition of uniqueness on `sources`, adding `depth`.
Shows multiple hobo loggers in the `TEMPERATURE ANALYSIS` chart. 

image


echaidemenos:
The chart may be too crowded now, maybe we should split them? Also the card becomes too crowded, for now i am only showing one hobo logger.

echaidemenos:
hello @Caesarh97! We made a test deployment for this PR [here](https://aqualink-app-programize.web.app/sites/1630) and uploaded some random sample data, how does it look?

The upload file had this form:
```
Temp,depth,Date,Time
19.6109769768861,1,10/08/2023,19:53
22.6625268511672,5,10/08/2023,19:53
20.0840630383378,10,,10/08/2023,19:53
...
```
Does this work out for the issue here?
It could also be 3 different upload files, the important part is to include the `depth` field

packages/website/src/common/Chart/MultipleSensorsCharts/index.tsx
```diff
@@ -563,11 +572,15 @@ const MultipleSensorsCharts = ({
...
+            data: timeSeriesDataRanges?.bottomTemperature?.find(
+              (x) => x.type === 'spotter',
+            )?.data,
           },
           {
             name: 'HOBO',
-            data: timeSeriesDataRanges?.bottomTemperature?.hobo?.data,
+            data: timeSeriesDataRanges?.bottomTemperature?.find(
+              (x) => x.type === 'hobo',
+            )?.data,
```
ericboucher, line 583: looks like we're running this piece of code multiple times. Maybe create a util and/ or get these values once?

ericboucher (approved):

PR #951, opened 2023-12-04
Fix could not create surveys

echaidemenos:
The response from `/upload` would be treated as text and not as a json. Probably a bug on `axios` had left the incorrect `responseType` slip away, till our resent update


ericboucher (approved):

ericboucher:
The changes in the test are due to missing mocked dates

PR #959, opened 2023-12-18
Monitoring improvements

echaidemenos:
This PR

- Adds `Created At` column at sites overview page, which by default sorts the table
- Improve `siteId` validation on requests
- Adds the ability to have descriptions on the monitoring cards and pages
- Site Metrics page now uses the same layout as the rest
- Sites Overview filters now immediately apply, no need for refresh.
- Option to download csv for site metrics
- Add options to download as csv data from MonitoringTable

Screenshot from 2023-12-18 14-01-17
Screenshot from 2023-12-19 17-02-22
Screenshot from 2023-12-21 13-16-48


packages/website/src/routes/MonitoringRoutes/Monitoring/index.tsx
```diff
@@ -3,7 +3,13 @@ import { colors } from 'layout/App/theme';
 import React from 'react';
 import { Link } from 'react-router-dom';
 
-const tiles = [
+interface Tile {
+  title: string;
+  link: string;
+  description?: string | React.JSX.Element;
```
echaidemenos, line 9: @ericboucher please feel free to suggest some descriptions, if needed :)

ericboucher: done

packages/website/src/store/User/helpers.ts
```diff
@@ -21,3 +21,22 @@ export const constructUserObject = async (
     token,
   };
 };
+
+const adminLevelToInt = (a: AdminLevel | undefined) => {
```
ericboucher, line 25: This seems like an overkill. I see two other options
1. use a list of allowed levels instead a `minAdminLevel` in the tile definition
2. use a list to define admin levels to simplify comparison:

```
export const AdminLevels = ['default', 'site_manager', 'super_admin'] as const; 
type AdminLevelType = typeof AdminLevels[number]; 
```

packages/api/src/monitoring/monitoring.service.ts
```diff
@@ -26,6 +30,15 @@ interface GetMetricsForSitesProps {
   endDate?: Date;
 }
 
+interface SiteMetrics {
+  date?: Date;
+  totalRequests: number;
+  registeredUserRequests: number;
+  siteAdminRequests: number;
+  timeSeriesRequests: number;
+  CSVDownloadRequests: number;
```
ericboucher, line 39: @echaidemenos is it really necessary to handle the CSV download in the backend?
Considering the relatively low amount of data, it might be easier to just add a download function to the table component directly and be able to save a table as CSV. What do you think?

packages/website/src/routes/MonitoringRoutes/SiteMetrics/index.tsx
```diff
@@ -150,134 +144,121 @@ function SiteMetrics() {
...
+
+  const getResult = React.useCallback(
+    async (token) =>
+      (
+        await monitoringServices.getMonitoringStats({
+          token,
+          ...(siteId && { siteIds: [siteId] }),
+          ...(spotterId && { spotterId }),
+          monthly,
+          start: startDate?.toISOString(),
```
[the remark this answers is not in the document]
echaidemenos: Yes i believe they are not intended to be optional, thanks for pointing it out!

packages/website/src/store/User/helpers.ts
echaidemenos: I believe it is more clear to specify the order of admin level directly, than use the index of the array defining them. I am changing the name of the function to `adminLevelOrderVal` and the `if` statements to `switch` so it is more clear

packages/api/src/utils/csv-utils.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      writeSync(fd, csvLines);
+    }
+
+    closeSync(fd);
+
+    const readStream = createReadStream(tempFileName);
+
+    res.set({
+      'Content-Disposition': `attachment; filename=${encodeURIComponent(
```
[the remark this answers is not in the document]
echaidemenos: thanks for checking this out! I am removing the `encodeURIComponent`, and also adding a check that the file name is valid

ericboucher (approved):

PR #992, opened 2024-05-06
Fix featured site data would be displayed after pressing back button

### LINKED ISSUE #991: Using the browser back arrow causes unexpected results
When using the website back arrow to go to the map and then using the browser back arrow to come back to the site, the webapp does not update properly and uses the "latestData" from the featured website.

Initial site
Screenshot 2024-05-06 at 6 55 15 PM

Site after we come back
Screenshot 2024-05-06 at 6 55 08 PM

Featured site
Screenshot 2024-05-06 at 6 55 28 PM

ericboucher:
Partially fixes #991


ericboucher:
@echaidemenos actually it still seems to be flashing for the customer.

PR #1005, opened 2024-07-17
Update HUI to show alert card instead of spotter card

### LINKED ISSUE #1004: Change default dashboard cards HUI
Hello!

Can you please make the Heat Stress Alert dashboard card visible instead of Buoy Observation on sites with HUI data? We'll be promoting their new dashboards in the coming week, so it would be awesome if we could show them from the best side possible. 

Thank you!

ericboucher:
Fixes #1004 

Screenshot 2024-07-17 at 4 44 46 PM


Caesarh97:
@ericboucher, Looks good to me. 

Thank you!

PR #1013, opened 2024-08-23
Update chartjs to latest (v4)

### LINKED ISSUE #994: Fix ghost hover points and upgrade Chartjs
There's an issue when you hover over the graphs on the dashboard. It works great until you hover over the circle that indicates a survey was conducted at that point in time. You can look at this site as an example. [https://aqualink.org/sites/3331](https://aqualink.org/sites/3331)

https://github.com/aqualinkorg/aqualink-app/assets/113801743/4fd9c60c-640f-44f0-83f5-bb73d7d5a3a8

yaodingyd:
fix #994 
/claim #994

This is to update chartjs to latest V4 and related dependencies to latest. There are a lot breaking changes from V2 -> V4, includes:

- [x] plugin interface change: (custom) tooltip
- [x] plugin interface change: annotation registration
- [x] x axis time adapter usage, replace moment with date-fns

Screenshot 2024-08-28 at 10 38 40 PM

there are some remaining type issues (all of them are related to using `any` to avoid typecheck error) but it is in a good state to review


packages/website/src/common/Chart/index.tsx
```diff
@@ -6,12 +6,16 @@ import React, {
   useState,
   memo,
 } from 'react';
-import { Line } from 'react-chartjs-2';
 import { useSelector } from 'react-redux';
+import { Line } from 'react-chartjs-2';
+import 'chart.js/auto';
+import { Tick } from 'chart.js';
+import 'chartjs-adapter-date-fns';
```
ericboucher, line 13: Looks like importing this like that is causing issues when running the tests.
We could try to import it in another file like you did for the annotation plugin, or maybe mock it during tests.

ericboucher:
@yaodingyd great work, thanks!

I noticed that we still have a "ghost" point when hovering on a survey. Eg. on site 3197. See the blue point over `June 1st` when hovering on the survey points.

Screenshot 2024-08-29 at 6 07 23 PM

packages/website/package.json
```diff
@@ -27,8 +27,9 @@
     "@testing-library/user-event": "^7.1.2",
     "@types/leaflet.locatecontrol": "^0.74.4",
     "axios": "^1.7.4",
-    "chart.js": "^2.9.3",
-    "chartjs-plugin-annotation": "^0.5.7",
+    "chart.js": "^4.0.0",
```
ericboucher, line 30: Let's use `^4.4.3` or similar directly no, wdyt?

yaodingyd:
@ericboucher can you take another look? I think the ghost hover issue is fixed


https://github.com/user-attachments/assets/22e81206-b195-4304-bb05-18249d4ae365

ericboucher (approved):
Thanks @yaodingyd, great work!!

PR #1028, opened 2024-09-20
Update react router usage to v6; fix minor issue

yaodingyd:
fix #1019
fix #1020 

/claim #1019
/claim #1020

1. The major change is to update react-router to v6, a lot of breaking changes includes
2.  use link `[highlights.aqualink.org/learn-more/get-started-with-aqualink#h.49zi97uifdjf](https://highlights.aqualink.org/learn-more/get-started-with-aqualink#h.49zi97uifdjf)` for dashboard
3. Update util to sort Sonda data to display the latest upload

Screenshot 2024-09-19 at 10 26 51 PM

Issue #1026, opened 2024-09-18
Add water quality sites in the map filter


Caesarh97:
On the /map page where we have a drop-down list where we filter sites, I’d like to add all the sites with water quality data. Would it be easy to add these? In this filter, all sites that have either {sonde} or {hui} can be added. Both are displaying water quality data and would fit in this filter. This issue is of low priority, so if it’s fast and easy, we could add it now. Otherwise, it can wait. 

<img width="516" alt="Screenshot 2024-09-17 at 4 15 05 PM" src="https://github.com/user-attachments/assets/b158db68-9a99-4ae4-a3fe-366171af8bc6">

yaodingyd:
Not sure if there's an easy way to do this. Right now we need to query `forcastData` and `latestData` two endpoints to find if there's any `hui` or `sonde` data, and these two endpoints require each specific site id, so we need to paginate through all sites and dispatch two endpoint for all of them to aggregate the collection of sites with water samplings. This seems rather efficient. 

I think to enable this feature, we would need to return
1)  water quality data to the top level `site` model, or
2) return ALL of latestData and forcastData
so we can process this in client side
not sure the implication/feasibility of doing this in api

ericboucher:
Indeed, I think we probably need to update the API `/sites` endpoint to achieve this. We could maybe add a boolean flag to the site if it has received any water quality data in the past X weeks. Where `X weeks` is the same metric we currently used on the site page to determined if we will display Water Quality data or not. Is that something you feel comfortable looking into @yaodingyd ?

yaodingyd:
sounds good, let me give it a go

yaodingyd:
@ericboucher  is there a way to populate local db with some mock data? I'm doing development under `api`, applied the migrations but need some data (ideally medium scale) to run it end-to-end

ericboucher:
@yaodingyd unfortunately we don't have a great way to do this at the moment, but I could create a small extract of our staging DB for you. Which tables would you need? You can email me: eric at ovio.org

PR #1031, opened 2024-09-22
Add cache layer to axios

yaodingyd:
Should be merged after https://github.com/aqualinkorg/aqualink-app/pull/1028 to validate  

When investigating #1026 I found out there is no existing cache layer on axios so there's a lot of duplicate calls:
 
For example when in site page, there are four identical calls to latest_data
Screenshot 2024-09-21 at 10 39 53 PM

with axios-cache-interceptor in place, we only have unique calls
Screenshot 2024-09-21 at 10 40 30 PM

the default cache TTL  is 5min which looks good. I don't think the backend data will change in a 5min window.


ericboucher:
Nice addition @yaodingyd. You might need to mock caching in the tests to turn the CI green.

yaodingyd:
tests are fixed

ericboucher (approved):

PR #1036, opened 2024-10-16
Improve upload sheet performance

K-Markopoulos:
## Issue
Upload sheet data flow is slow and process time grows exponentially as the size of the file grows.

Some benchmarks using upload-sonde-script:
|# Rows | Grouping |Converting|Total|
|---|---|---|---|
|10k|0s|3s|47s|
|100k|90s|5m|7m|
|200k|5m|26m|32m|

## Changes
- Remove array spreading for adding new elements. This resulted in allocating and copying the whole array on each iteration. Use `.push()` or pre allocate the array and fill in the elements.
- Remove constant calculations out of the main loop.

|# Rows | Grouping |Converting|Total|
|---|---|---|---|
|10k|0s|0s|28s|
|100k|0s|0.3s|30s|
|200k|1s|0.5s|30s|


packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -849,6 +864,7 @@ export const uploadTimeSeriesData = async ({
     }),
   );
 
+  logger.log('Refresh materialized view');
```
ericboucher, line 867: The refreshMaterializedView already has logging internally but using the console method. If you believe it is useful to switch to logger, we can do that there.

packages/api/src/utils/uploads/upload-sheet-data.ts
```diff
@@ -328,14 +343,25 @@ export const convertData = (
...
+      return acc;
+    }
+    return { ...acc, [i]: token };
+  }, {});
+  const results = Array(
+    workSheetData.length * Object.keys(metricHeadersMap).length,
+  );
+  let resultsIndex = 0;
 
+  logger.log('Converting data');
```
ericboucher, line 361: 👣 ?

packages/api/src/utils/uploads/upload-sheet-data.ts
K-Markopoulos: The console method inside the function is sufficient, I'll remove the log

packages/api/src/utils/uploads/upload-sheet-data.ts
K-Markopoulos: Removing the log

Issue #1038, opened 2024-10-22
Some values are not visible on "EDIT SITE DETAILS" 


Caesarh97:
When you have edited and saved either the "Status" or the "Add a custom API token", the data is not visible. The data gets saved properly to the database, but you can't see it in the dashboard view. For example, [this site](https://aqualink.org/sites/3445) (also see attached image) actually has its status as "deployed" and has the "Add a custom API token" checked, and the API token is added to the API textbox. 

Would it be possible to make this visible when the data has been entered and saved?

Please email admin@aqaulink.org for clarification if necessary. 

![Screenshot 2024-10-21 at 1 56 13 PM](https://github.com/user-attachments/assets/3b49b633-57aa-4d18-883b-6cfc276fa679)

PR #1043, opened 2024-10-23
Populate submitted values in site details

K-Markopoulos:
## Issue #1038 

Status and custom spotter api token values are not displayed in site details.

## Changes
- Add status initial value
- Change api to return the spotter api token masked (only 4 last digits) if it has custom token. Unfortunately we cannot add a new computed property in the Site entity, because the `spotterApiToken` is not selected and therefore accessible by default, so we select all site properties and whitelist the properties we want to return.
- On UI, show the masked token. User can provide a new one by clicking on the input field - then the old token is erased and the user can start typing the new token. If they leave the input without typing anything, the old token is filled back in. If the user wants to completely remove the custom token they uncheck the "Add custom api token" checkbox.

Note the logged `spotterApiToken`, when `null` the token is deleted, when `undefined` it's not changed and when it has value, the value is persisted.

[Screencast from 23-10-2024 05:16:05 ΜΜ.webm](https://github.com/user-attachments/assets/4493b083-bae1-4116-b284-0c755d51b260)


Caesarh97:
Hi @K-Markopoulos, 
Thank you for working on this. When testing it out, I added a spotter (not owned by Aqualink) and added its API token. I updated the page and went back to the dashboard. You can see the screenshot below. The Status "deployed" works fine, but I can not see the "Add a custom API token" box being checked or any value in the token textbox. 

Regarding the second point in your changes, I didn’t fully understand it. If it's possible to show the last values that were entered for the API token, that's great. If not, that's fine and no need to do too much work on it. This view is locked to the SuperAdmin and the Site Manager for the specific site, so there's no need to mask the token or anything fancy (unless it has already been done). The "EDIT SITE DETAILS" view is only for the site manager of the specific site and the SuperAdmins to keep track of and troubleshoot what data has been entered into the database. 

Screenshot 2024-10-23 at 8 59 51 AM

K-Markopoulos:
Hi @Caesarh97,
If I'm not wrong, the generated preview link (https://aqualink-app-1043.surge.sh/) includes only the frontend changes of the PR. The backend changes have to be manually deployed to be available/testable and the spotter api token fix has some backend changes. @ericboucher can you help us with the backend deployment in order to test this PR?

The 2nd point is describing some technical things about the implementation. Basically, with the current implementation, once the token is persisted, it won't be fully visible in the UI. The user will be able to see only the last 4 characters of it. If they set "123456789", they will see "****6789".

Caesarh97:
> Hi @Caesarh97, If I'm not wrong, the generated preview link (https://aqualink-app-1043.surge.sh/) includes only the frontend changes of the PR. The backend changes have to be manually deployed to be available/testable and the spotter api token fix has some backend changes. @ericboucher can you help us with the backend deployment in order to test this PR?
> 
> The 2nd point is describing some technical things about the implementation. Basically, with the current implementation, once the token is persisted, it won't be fully visible in the UI. The user will be able to see only the last 4 characters of it. If they set "123456789", they will see "****6789".

ericboucher (approved):

packages/api/src/sites/sites.entity.ts
```diff
@@ -151,10 +151,12 @@ export class Site {
   @Column({ nullable: true, select: false, type: 'character varying' })
   contactInformation?: string | null;
 
-  hasHobo: boolean;
+  hasHobo?: boolean;
```
ericboucher, line 154: Was this change needed?

ericboucher (approved):

packages/api/src/sites/sites.entity.ts
K-Markopoulos: Yes, since it was a required property in `Site` TS expected [sites.service.ts::findeOne](https://github.com/aqualinkorg/aqualink-app/blob/ea8792db1aafb17fc0174f14589f23c41b08d7db/packages/api/src/sites/sites.service.ts#L209) to return this property as well. The spread operator `...site` tricked TS into thinking that `hasHobo` is included in the response, even though it was actually not. After removing the spread operator and explicitly defining the properties, TS throws an error that `hasHobo` is missing. I had to make this property optional to overcome this error and since it's not always defined, it makes sense to be optional.

ericboucher: Ok, the alternative would be to actually have it defined no? Since it can only be true or false?

Issue #1022, opened 2024-09-18
New default dashboard card


Caesarh97:
We already have over 800 displayed sites and only 39 deployed buoys. Once we have the Reef Check sites available, there will be even more sites without a buoy connected to them. Because of this, it would be great if we changed the "Buoy Observation" dashboard card. 

When a site does not have a buoy, the default mode should be a new dashboard card instead of “Buoy Observation”. The new card should show the change in temperature over the past 7 days compared to the 7 days before that. It should show the increase / decrease in absolute terms (e.g. +1C, the previous temperature e.g. 27c and an arrow up or down - blue if colder, red if hotter). This would come from SST and would only be shown if we have no spotter. 

Below are 3 examples of how the dashboard could look like. Note that I included two images for each example. We could also consider displaying the 7-day average temperature at the bottom, as it forms the basis for the calculated change. Please let me know what you think would be best. 

**UPDATE: We should do Example 2.** 

Example 1
![Example 1](https://github.com/user-attachments/assets/d439781f-c7b4-4452-9d93-9fe8175c7d47)
Example 1
![Example 1](https://github.com/user-attachments/assets/bc337e10-93d0-4a4c-a840-5d72f1394b52)

Example 2
![Example 2](https://github.com/user-attachments/assets/f2249f8a-ee4a-4a69-8bf4-019c0c992b6e)
Example 2
![Example 2](https://github.com/user-attachments/assets/add6986b-23e7-47e6-9008-996ded4b3d5b)

Example 3
![Example 3](https://github.com/user-attachments/assets/f5e24848-293b-49ef-9f2d-cb8a3339abb4)
Example 3
![Example 3](https://github.com/user-attachments/assets/002e52b8-d5ff-4894-a6d2-5c146742921a)

PR #1046, opened 2024-10-25
Add temperature change card in site dashboard

K-Markopoulos:
## Issue #1022 

Add 7-days temperature change card in place of Buoy Observation card when site doesn't have a spotter.

- Calculate the change based on `satelliteTemperature` from site's `dailyData`:  avg of 7 last days minus avg of 8-14 days back
- Calculate the avg temp of the last 7 days.

image

### DIFF
--- a/packages/website/src/common/SiteDetails/index.tsx
+++ b/packages/website/src/common/SiteDetails/index.tsx
@@ -43,7 +43,7 @@ import Map from './Map';
 import SketchFab from './SketchFab';
 import FeaturedMedia from './FeaturedMedia';
 import Satellite from './Satellite';
-// import Sensor from './Sensor';
+import Sensor from './Sensor';
 import CoralBleaching from './CoralBleaching';
 import Waves from './Waves';
 import OceanSenseMetrics from './OceanSenseMetrics';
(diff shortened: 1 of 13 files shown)


packages/website/src/common/SiteDetails/index.tsx
```diff
@@ -42,7 +43,7 @@ import Map from './Map';
 import SketchFab from './SketchFab';
 import FeaturedMedia from './FeaturedMedia';
 import Satellite from './Satellite';
-import Sensor from './Sensor';
+// import Sensor from './Sensor';
```
ericboucher, line 46: Note that we still want to display the Sensor card when there is a spotter/buoy with live data for that site

Caesarh97:
Hi @K-Markopoulos,

I think many of the changes look great. However, I'd like to add some more information about the card. We would like the number in color to be calculated by the past 7 days compared to the 7 days before that. As Eric mentioned, this card should be shown if there's no buoy data available. If a site has water quality data and no buoy, we should still display the heat stress card together with the water quality card like before. I think we would want to show the temperature and one decimal number (ex. 22.6°C, just like you've done in the example). 

I think this calculation would be the most accurate to calculate the past 7 days compared to the 7 days before that. 
(sum of most recent 7 days) / 7 = (average temp for past 7 days)
(Sum of past 14-8 days / 7 = (average temp for past 14-8 days)
(average temp for past 7 days) - (average temp for past 14-8 days) = average 7-day change in temperature for the past 7 days compared to the previous 7 days.

Example:
Past 14 days (Day 14 - Day 1. Left to right)
23, 24, 26, 22, 19, 20, 24, 22, 25, 26, 25, 23, 23, 22
(Sum of 22, 25, 26, 25, 23, 23, 22) / 7 = 23.7 (average most recent 7 days)
(Sum of 23, 24, 26, 22, 19, 20, 24) / 7 = 22.6 (average past 14-8 days)
23.7-22.6=1.1
The average 7-day change in temperature for the past 7 days compared to the previous 7 days is 1.1°C.

K-Markopoulos:
@Caesarh97 Thanks for the detailed explanation, I've updated the logic. Please let me know if anything seems off.

Caesarh97:
> the past 7 days compared to the 7 days before that.

@K-Markopoulos That looks great! Thank you!

ericboucher (approved):

PR #1049, opened 2024-11-08
Add water quality data to site model and use it to power site filter

yaodingyd:
fix https://github.com/aqualinkorg/aqualink-app/issues/1026
/claim #1026

Add `waterQuality` to `site` model that would include list of water quality metrics.

The logics to determine water quality metrics is from https://github.com/aqualinkorg/aqualink-app/blob/56dbb9c8a3115699ab74c97bf53e1962a67cf82e/packages/website/src/common/SiteDetails/index.tsx#L157-L174

which looks at `LatestData` entity (and forecast data, omitted by now) and check if they have `hui` source or enough `sonde` metrics.

If this is the correct direction, I can work on a follow up to remove the checks in frontend and consolidate to the backend, so we have single source of truth. This also avoid multiple data fetching to get water quality data


ericboucher:
I think it might be due to GitHub protections around variables for external contributors... It's super annoying. Let me add you to the repo if that's ok with you?

yaodingyd:
I think PR has to be opened in this repo (not in forks) for API CI to pass, like https://github.com/aqualinkorg/aqualink-app/pull/1075. I'll recrease this PR

PR #1052, opened 2024-11-13
Add Reef Check entities

K-Markopoulos:
Integrate provided Reef Check Data into Aqualink.

- Add Reef Check entities: `ReefCheckSite`, `ReefCheckSurvey`, `ReefCheckOrganism`, `ReefCheckSubstrate`
- Add script to parse the xlsx files and create the entities.
- Add basic api controllers for sites and surveys

image

### DIFF
--- a/packages/api/scripts/reef-check.ts
+++ b/packages/api/scripts/reef-check.ts
@@ -1,0 +1,700 @@
+/* eslint-disable fp/no-mutating-methods */
+/* eslint-disable fp/no-mutation */
+import xlsx from 'node-xlsx';
+import { DataSource, DataSourceOptions } from 'typeorm';
+import { Logger } from '@nestjs/common';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+import { groupBy, keyBy, uniqWith } from 'lodash';
+import fs from 'fs';
+import { ReefCheckSubstrate } from '../src/reef-check-substrates/reef-check-substrates.entity';
+import { ReefCheckOrganism } from '../src/reef-check-organisms/reef-check-organisms.entity';
+import { ReefCheckSurvey } from '../src/reef-check-surveys/reef-check-surveys.entity';
+import { Site } from '../src/sites/sites.entity';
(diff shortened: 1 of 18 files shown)


packages/api/src/reef-check-surveys/reef-check-surveys.entity.ts
```diff
@@ -0,0 +1,301 @@
...
+  @VirtualColumn({
+    query: (survey) => `
+      SELECT satellite_temperature
+      FROM daily_data
+      WHERE DATE(daily_data.date) = DATE(${survey}.date)
+      AND daily_data.site_id = ${survey}.site_id
+      LIMIT 1
+    `,
+  })
+  satelliteTemperature: string;
```
K-Markopoulos, line 300: This is based on surveys service

https://github.com/aqualinkorg/aqualink-app/blob/343573b7277e5414cac60507aae6cff3fc36bbf9/packages/api/src/surveys/surveys.service.ts#L247-L252

@ericboucher please verify whether this is the right way to compute the satellite temperature.

image

packages/api/scripts/reef-check.ts
```diff
@@ -0,0 +1,700 @@
+/* eslint-disable fp/no-mutating-methods */
```
ericboucher, line 1: Let's add a bit more comments in this file to make sure we will know how to adapt it and run it again with new data (and/or overlapping data).

packages/api/src/reef-check-surveys/reef-check-surveys.entity.ts
ericboucher: So ultimately we are pulling data from daily_data ? I think that's correct but it also means that we need to update the daily data table for all the site/survey date combos

ericboucher:
Great work, let's add comments and also expand the README to make a note of this addition

packages/api/src/reef-check-surveys/reef-check-surveys.entity.ts
K-Markopoulos: Indeed that would require to add the daily_data entry for the date of the survey

ericboucher (approved):

PR #1064, opened 2024-12-03
Add Reef check surveys in Site page

K-Markopoulos:
Based on `reef-check-survey-page` branch. Will switch target to `master` once the base branch is merged.

- Add reef check surveys to surveys list in Site's page.
- Show both type of Surveys in the same sorted list. Each type with its own card.
- Create table to show a sample of survey stats.
- [x] Add Tests

image


ericboucher:
Nice work @K-Markopoulos 
I deployed to programize so progress can be tested here - https://aqualink-app-programize.web.app/sites/5427

One thing to update is moving the survey "date line" to the left to make more room fo the table as you can see from the designs

PR #1069, opened 2024-12-11
Improve map performance - use points on canvas

K-Markopoulos:
The goal is to improve initial load of the map and provide smooth scrolling and zooming. The main cause is that we are rendering (and re-rendering many times) 5.000+ markers at once.

## Changes
- Incrementally mount site markers. Mount only visible markers, to avoid mounting all at once and freezing the UI. Keep the mounted to avoid re-mounting.
- Replace `siteOnMap` selector with `isSelected` selector which caused all markers to re-render when selected site was changed.
- Add back marker clustering.
- Move "fly to selected site" logic to map component and remove double rendering.
- [x] Add back offset markers functionality.
- [x] Investigate cluster icon color not matching alert levels.

image


ericboucher:
@K-Markopoulos could you rebase on the last version of the reef-check frontend so that we can deploy it to programize for testing?

K-Markopoulos:
> @K-Markopoulos could you rebase on the last version of the reef-check frontend so that we can deploy it to programize for testing?

@ericboucher  Sure but what's the plan for testing? This PR doesn't have any overlapping changes with the reef-check frontend PRs. The generated link: https://aqualink-app-1069.surge.sh/ shows the reef-check sites on the map as well.

K-Markopoulos:
Comparing different clustering options


|Options|Screenshot|Full map|
|---|---|---|
|maxClusterRadius={80}, sizes (1k, 100, 10, 1): [100,64,48,36]|image|image|
|maxClusterRadius={60}, sizes: [50,44,36,32] |image|image|
| background: max alert color |image|image|

Caesarh97:
@K-Markopoulos I like the clustering icons. I like when we clearly show how many sites that experience what heat stress. I think the white center clarify this the most. 

If we would color the center, then we should display the most abundant heat stress level for the specific area. I think we should maybe separate the center and the outside circle a little bit more if we’d choose this choice. See example below. 

Which do you think would be best?

Untitled design

ericboucher:
Let's try a final time to improve map performance using geojson grouping.
If that doesn't work then we can stick with the clustering and upgrade it to match this figma: https://www.figma.com/design/5FeX5lEkdcRNZRaY8WtcqC/Aqualink?node-id=690-306&t=fZTLw6H98ekXnnSW-0

ericboucher:
In the end we used points, see PR #1082

image (2)

ericboucher (approved):

PR #1076, opened 2024-12-17
Add water quality data to site model and add site filter

yaodingyd:
fix https://github.com/aqualinkorg/aqualink-app/issues/1026
/claim #1026

Add `waterQuality` to `site` model that would include list of water quality metrics.

The logics to determine water quality metrics is from https://github.com/aqualinkorg/aqualink-app/blob/56dbb9c8a3115699ab74c97bf53e1962a67cf82e/packages/website/src/common/SiteDetails/index.tsx#L157-L174

which looks at `LatestData` entity (and forecast data, omitted by now) and check if they have `hui` source or enough `sonde` metrics.

If this is the correct direction, I can work on a follow up to remove the checks in frontend and consolidate to the backend, so we have single source of truth. This also avoid multiple data fetching to get water quality data


packages/api/src/utils/site.utils.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable fp/no-mutating-methods */
```
ericboucher, line 1: Would it be easy to do this inline instead of for the whole file?

packages/api/src/sites/sites.entity.ts
```diff
@@ -167,6 +167,8 @@ export class Site {
 
   maskedSpotterApiToken?: string;
 
+  waterQuality?: string[];
```
K-Markopoulos, line 170: I suggest we rename this field to indicate this is a list of strings(Sources), maybe:

```suggestion
  waterQualitySources?: string[];
```

packages/api/src/utils/site.utils.ts
```diff
@@ -278,6 +278,52 @@ export const hasHoboDataSubQuery = async (
   return hasHoboDataSet;
 };
 
+export const hasWaterQualityDataSubQuery = async (
```
K-Markopoulos, line 281: `has` prefix is usually used for functions that return boolean, but this one returns a map. It might be better to rename or return a boolean instead of the map, since we are only interested in a specific site and not all the map.

packages/api/src/utils/site.utils.ts
```diff
@@ -278,6 +278,52 @@ export const hasHoboDataSubQuery = async (
...
+  latestDataRepository: Repository<LatestData>,
+): Promise<Map<number, string[]>> => {
+  const latestData: LatestData[] = await latestDataRepository
+    .createQueryBuilder('water_quality_data')
+    .select('site_id', 'siteId')
+    .addSelect('metric')
+    .addSelect('source')
```
K-Markopoulos, line 288: We can rewrite this to a single `select`

```suggestion
    .select('site_id', 'siteId', 'metric', 'source')
```

packages/api/src/utils/site.utils.ts
```diff
@@ -278,6 +278,52 @@ export const hasHoboDataSubQuery = async (
...
+    .where(`source != '${SourceType.HOBO}'`)
```
K-Markopoulos, line 289: We are looking for `hui` and `sonde` sources so let's specify that in the `where` clause to only get these. Also, it will be more performant to filter by `siteId` here and not load the whole table (or view) in memory.

packages/api/src/utils/site.utils.ts
```diff
@@ -278,6 +278,52 @@ export const hasHoboDataSubQuery = async (
...
+    .getRawMany();
+
+  const sondeMetrics = [
+    'odoConcentration',
+    'cholorophyllConcentration',
+    'ph',
+    'salinity',
+    'turbidity',
+  ];
```
K-Markopoulos, line 298: We can use the metrics in snake_case as they are stored in DB instead of writing them in camel case and converting the metric in each iteration. We can use `Metric` enum directly.

packages/api/src/utils/site.utils.ts
yaodingyd: on a second look, this is needed since the first select also adds alias `site_id` => `sideId`

K-Markopoulos: You are right, my bad. Please ignore this comment

packages/api/src/utils/site.utils.ts
K-Markopoulos: Nice work, ignore my comment regarding filtering, it doesn't make sense, I though this was the findOne function.

K-Markopoulos (approved):
Nice work, looks good :rocket:

PR #1079, opened 2025-01-10
Add 200.html in surge deployment to fix routing

K-Markopoulos:
Fix routing for surge deployments (UI temp deployment for PR builds) by proving a 200.html as mentioned in the [docs](https://surge.sh/help/adding-a-200-page-for-client-side-routing).

We can now directly access specific pages and not only the home page (`/`)
-> https://aqualink-app-1079.surge.sh/sites/123

### DIFF
--- a/.github/workflows/build_website.yml
+++ b/.github/workflows/build_website.yml
@@ -61,6 +61,11 @@ jobs:
         working-directory: packages/website
         run: yarn build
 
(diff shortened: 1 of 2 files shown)


ericboucher (approved):

ericboucher:
Nice find. I remember trying something like that in the past which did not work. Great to see you found the right way to do it!

PR #1083, opened 2025-01-20
Fix Reef check issues

K-Markopoulos:
- Fix typo in table header DIDEASES -> DISEASES
- Show empty cell when data is null (instead of `null°C`)
- Hide observation box when there are no data
- Update yarn lock and snapshots

Check: /6762/reef_check_survey/c4e52389-e1d2-4cf0-a389-cdea16b059d0


K-Markopoulos:
> I don't have any concerns with these changes, apart from the yarn.lock file changes. Why were these necessary? Can these be moved to a separate PR?
> 
> Image of Graham C **Graham C**
> 
> Reviewed with ❤️ by [PullRequest](https://www.pullrequest.com)

Looks like the yarn.lock is not in sync with package.json in master branch. package.json was updated but yarn.lock changes were not committed. Since this is a bucket PR to address a few minor issues, I thought I should address this inconsistency in this PR as well.

ericboucher (approved):

PR #1090, opened 2025-02-03
Add new map filters

### LINKED ISSUE #1066: Reef Check - new filter button for the map
Hello, 

We want to populate the map with Reef Check data, and we want the users to be able to filter the map using the current metrics and the Reef Check data. In the image below, you can see what we want this filter to look like. 

- We want to have `All Sites and Data`, `Heat stress status` (drop-down), `Sensor and data types` (drop-down), `species` (drop-down), `Reef composition` (drop-down), and `Anthropogenic impact` (drop down). 

- You should be able to select and filter multiple categories and items at once. 

`All Sites and Data` should be applied by default. When selecting an item in a category, all sites on Aqualink that match that option or options will populate the map. 
- For Reef Check, all items that have >0 in count OR higher than 0 in % will show up. `Species`, `Reef composition`, and `Anthropogenic impact` are Reef Check data.
- For `Heat Stress Status`, all sites that are experiencing the current selected status/statuses will populate the map. 
- `Sensor and Data Types` - just like they are populating the map today, but multiple types can be selected. 

I suggest that when any item in any category gets checked, the `All Sites and Data` will be unchecked. If possible, when nothing is selected, the `All Sites and Data` will be rechecked by default. 

Please note that if an item isn't selected, it does not mean that it can't be shown on the map. For example, if only the item `Active buoys` is checked, no items in the `Heat Stress Status` category will be checked. That would still mean that all different Heat Stress Statuses show up as long as there's a buoy there. 

For `Species`, there are A LOT of species. Too many to display in checkboxes in the filter. Therefore, I've compiled the species into a shorter list ONLY for filter purposes. Please see this [spreadsheet](https://docs.google.com/spreadsheets/d/1f5zly_WJjc9sFCj8MyGkrPrHpw-EBMRUVGFuBN9jBfs/edit?usp=sharing). 

Please note that `All Sites and Data` should be spelt with capital letters for Sites and Data. 

The filter button will be 

image

### LINKED ISSUE #1078: Change color search bar
Hello,

Can we change the color of the search icon, the small triangle, and the text "Search" in the search bar to black #000000? 

[https://www.figma.com/design/5FeX5lEkdcRNZRaY8WtcqC/Aqualink?node-id=411-5&t=O6P9MhH1uyGRiyXe-4](https://www.figma.com/design/5FeX5lEkdcRNZRaY8WtcqC/Aqualink?node-id=411-5&t=O6P9MhH1uyGRiyXe-4)

K-Markopoulos:
Add new map filters [figma](https://www.figma.com/design/5FeX5lEkdcRNZRaY8WtcqC/Aqualink?node-id=469-1122&t=Vk6UMnCLezVolK76-4). Closes #1066 

- Add `reefCheckData` property to each site in `/sites` response which includes some attributes for the site in order to use them for filtering in UI.
- Add new modal for filters.
- Refactor sites store, `filters` are in state and displayedSites is a selector now.
- Expand by default categories with active filters 
- Store filters in URL, so they can be shared with URL . Load filters on page load.

Other:
- Change search bar color to black. Closes #1078

image

image


ericboucher:
You can maybe address https://github.com/aqualinkorg/aqualink-app/issues/1078 as well in this PR since we're touching at the main page?

packages/api/src/utils/site.utils.ts
```diff
@@ -325,6 +326,56 @@ export const getWaterQualityDataSubQuery = async (
...
+): Promise<
+  Dictionary<{ siteId: number; organism: string[]; substrate: string[] }>
+> => {
+  const organisms: { siteId: number; organism: string[] }[] =
+    await reefCheckSurveyRepository
+      .createQueryBuilder('survey')
+      .select('survey.site_id', 'siteId')
+      .addSelect('json_agg(distinct rco.organism)', 'organism')
+      .leftJoin('reef_check_organism', 'rco', 'rco.survey_id = survey.id')
+      .where('rco.s1 > 0')
```
ericboucher, line 344: Would it simplify a bit to so something like this? `.where('(rco.s1 + rco.s2 + rco.s3 + rco.s4) > 0')`
Or maybe not? Not sure if it would make it faster

K-Markopoulos: I would assume that with multiple _simple_ predicates (`s1 > 0 or s2 > 0`) postgres would be able to do more optimizations and therefore it would be faster, but after analyzing the query plan and the execution time it seems they are equivalent. It likely won't make it faster or slower, but it would make it simpler, so let's go with your suggestion.

ericboucher (approved):

packages/website/src/store/Sites/constants.ts
```diff
@@ -0,0 +1,72 @@
...
+  ],
+  siteOptions: [
+    { value: 'activeBuoys', label: 'Active buoys' },
+    { value: 'liveStreams', label: 'Live streams' },
+    { value: '3DModels', label: '3D Models' },
+    { value: 'hoboLoggers', label: 'HOBO loggers' },
+    { value: 'waterQuality', label: 'Water quality' },
+    { value: 'reefCheckSites', label: 'Reef Check Sites' },
+  ],
+  species: {
```
ericboucher, line 17: Let's order the lists alphabetically or by coherence

K-Markopoulos: Sure, let me know if the new order is as expected

PR #1095, opened 2025-02-13
Add request to download data link

### LINKED ISSUE #1080: Add "Request to download reef check data" button
[Figma](https://www.figma.com/design/5FeX5lEkdcRNZRaY8WtcqC/Aqualink?node-id=282-564&t=02aqN2YwTIoVlitR-0)

Finalize the behavior of the button (what data and in which shape they will be downloaded) and add the button along with its functionality.

K-Markopoulos:
Add link to external google form to request to download reef check data.
Closes #1080 

Example: https://aqualink-app-1095.surge.sh/sites/4158/reef_check_survey/6163f960-1c87-4e23-936f-0f80af847d02
image


Caesarh97:
@K-Markopoulos, This site is missing the satellite temperature card in the top right corner. Pete's comment only said not to show the satellite data card if no data was available. 

I think an even better solution would be to show the satellite temperature from the day the survey was conducted. I know you're backfilling satellite temperature, but I don't know how far back. Would this be possible? If not, maybe we should specify that the satellite data is today's data. 

Screenshot 2025-02-13 at 8 26 19 AM

ericboucher:
@Caesarh97  not sure I understand your ask here. We only backfill 90 days in the past so no reef check surveys won't be covered. I'm not sure it makes sense to show today's date tbh

Caesarh97:
@ericboucher You're right. It doesn't fit in. I think it would look more aesthetically pleasing if we did something like the images below. What do you think?

PR #1093, opened 2025-02-10
Add pages metadata

K-Markopoulos:
(Waiting for NextJS migration to me merged first)

Add metadata content to enhance preview links shared on social media and messages - closes #641 

- Content is based on #1050 
- Some sites have `name: null` and we fallback to `Site ${id}`. Feel free to suggest otherwise.
- Env var `NEXT_PUBLIC_WEBSITE_BASE_URL` is now required at build time, to generate absolute URLs for images
- We use `/sites` api on each page generation because of caching on the same request.

Example:
image

PR #1102, opened 2025-02-20
Migrate from create-react-app to rsbuild & vitest, use cloudfare worker to serve render metadata

yaodingyd:
Three major changes:
1. replace create-react-app with rsbuild
2. replace jest with vitest
3. add cloudflare worker to do server render with site specific meta info

The current app architecture is still the same, when testing frontend changes, we can still use surge to deploy static asset and verify. The change to production is instead of using firebase hosting for static asset, we use a cloudflare worker to serve and pre-render meta info, if applies.

The workflow change is instead of deploy to firebase, we will deploy the CF worker. I think we also need to wire up the domain.

This is very preliminary that CF worker will fetch and render everything. We could fetch all sites info, store them in KV (which also have free-tier and more than enough for under 100k read/write per day) so meta info is cached.

For reviewing: majority change is in snapshot. The part needs review is mostly `worker.ts` and all new config files

## Feature highlights
- No app architecture change, still running as SPA
- No build step change
- One single server function (worker.ts) which reads index.html, replace meta info and serve it
- Minor change to deployment process

## Testing

Testing url: https://aqualink.yaodingyd.workers.dev/sites/2919

## Cost

The major cost would be requests for worker, and free tier has 100k per day, 1k per minute limit.
[The static assets are free and unlimited](https://developers.cloudflare.com/workers/static-assets/billing-and-limitations/).

So that means if the sites has under 100k page visit per day, we are good to stay in free tier and no cost added.


ericboucher:
Thanks a lot! This is great work and we do need to move away from create-react-app

@K-Markopoulos actually started a transition to nextJS because we had the call to add page specific metadata. But he was hitting a few blockers. Do you have some thoughts on the matter? Happy to discuss over slack or email.

See #1085 and #1093

yaodingyd:
First of all K-Markopoulos is doing an amazing migration work! However I'm wondering to use next.js just for its SSG output, considering the cost of allover architecture change, is the good fit here. Since we are not on Vercel, and I have no intention to migrating to, we cannot take advantage of incremental build, so each SSG is a fresh build for all pages. In that case, using a build tool (rsbuild/rspress) and keeping current architecture might be simpler. 

Furthermore, I'm wondering if using SSG is good here, since we will regenerate all pages on every change and with sites increasing, the build job will also scale up. What if we use a cloud function to just replace the meta info in generated html? that way each specific page is only partially pre-rendered and that's the only change we need. We can only keep the staging preview functionality, only output one single build output index.html.

I'll try to test this hypothesis and update here with more results.

yaodingyd:
Screenshot 2025-02-22 at 4 17 53 PM

https://aqualink.yaodingyd.workers.dev/sites/2919

I think it's definitely doable! I'll add detail later

yaodingyd:
I can move the worker setup into packages/website so they are colocated. I'll update this PR to have everything

yaodingyd:
@ericboucher  this is ready for final review

ericboucher:
Getting there! There is a small issue with the site page backbutton, I probably messed up the merge of #1115

PR #1132, opened 2025-05-08
Improve collections

ericboucher:
- Default to sat map for collections
- Filter out empty buoy columns in collection site table


ericboucher:
@Caesarh97 https://aqualink-app-1132.surge.sh/collections/minderoo

PR #1141, opened 2025-05-22
fix home page's metadata

yaodingyd:
Homepage's meta is not working as expected:

Screenshot 2025-05-21 at 8 15 23 PM

It looks that CF worker function will not handle the root path (/), it defaults to serving static assets like index.html from the ASSETS bucket.

So in this PR, we update the default index.html to have the correct meta, and in worker to correctly replace these meta  (to avoid two title/descriptions) for other routes.


packages/website/package.json
```diff
@@ -118,8 +118,5 @@
     "surge": "^0.23.1",
     "vitest": "^3.0.7",
     "wrangler": "^4.12.0"
-  },
```
yaodingyd, line 121: some clean up, volta is my personal choice of js tooling manager, though it's pretty good https://volta.sh/

PR #1145, opened 2025-06-02
Collection names

Caesarh97:
Hi Eric, I've added the collection names now. I added them to two files since the worker.ts has been added now.


Caesarh97:
Hi @ericboucher,

I wanted to share a quick update on the latest push. I was finally able to resolve the issue preventing `yarn test -u` from running, which turned out to be an incompatible Node.js version on my local environment. I did this with the help of Google AI Assistant.

After updating and running a clean `yarn install`, everything works correctly. The `yarn test -u` generated the updated test snapshots, and the `yarn install` created a completely new `yarn.lock` file, so I wanted to give you a heads up on why that change is there.

Please let me know if everything looks good from your end.

I see this has now created a merge conflict in the `yarn.lock` file. Would you be able to take a look at it?

Thanks!

ericboucher (approved):

PR #1152, opened 2025-06-20
add map info dialog

### LINKED ISSUE #1149: Map info icon
Hi @ericboucher, 

We want to add some info and guidance to the map. The idea is to add an info icon button below the full-screen icon on the map. When clicking on the info icon, a pop-up window will appear. This can work exactly the same as the filter window that pops up. Same size, white background, and scrollable. The text should be black with blue titles. The text and image content should look like [this page](https://highlights.aqualink.org/map-how-to-use). Text size, line and paragraph spacing, and font size should be reduced from the content in the link. The final version should look like the images below. 

Would this be a relatively easy addition? Please let me know if there's anything I can do to make this easier and faster for you. 

Image

Image

Image

yaodingyd:
Fix #1149
/claim #1149

https://github.com/user-attachments/assets/ff818587-9cc0-4c34-b01c-f65fecac4136

### DIFF
--- a/packages/website/src/routes/HomeMap/Map/InfoDialog/index.tsx
+++ b/packages/website/src/routes/HomeMap/Map/InfoDialog/index.tsx
@@ -15,7 +15,7 @@ import Tune from '@mui/icons-material/Tune';
 import CloseIcon from '@mui/icons-material/Close';
 import { spotter } from 'assets/spotter';
 import { hobo } from 'assets/hobo';
-import AlertLevelLegend from '../alertLevelLegend';
+import { AlertLevelLegendContent } from '../alertLevelLegend';
 
 interface IconDescriptionProps {
   icon: React.ReactNode;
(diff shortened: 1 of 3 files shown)


ericboucher:
Great start ! Thanks a lot for taking up this task. A few remarks:
- Some of the icons don't work on my end. They should be readily available through the libs we use without the need for external links. If it's tricky, let's create simple svgs or pngs in the code.
- The app structure should be more Reacty, in particular, using Typography and maybe moving it to its' own file? wdyt?

Caesarh97:
Hi @yaodingyd, Thanks for adding the feature! It works very well on a desktop for me. However, it looks a little bit weird on mobile. The no alert, watch, warning, etc. buttons are covering the text. The icons in the top are not lined up properly. Would it be possible to fix those two things?

IMG_0148
IMG_0149

Caesarh97:
@yaodingyd The top layer icon looks great. However, the alert icons have two issues: 

**Desktop:** The icons are covering the text on the desktop now. 


**Mobile:** The icons have a square shape, where the text inside the icons doesn't fit. Suggestion: Let's reformat the icons to display in two rows on mobile, like this:

Row 1: no alert watch warning
Row 2: alert 1 alert 2

Would that work, you think?

yaodingyd:
I change the font size to be responsive to fit in one line. WDYT?

Screenshot 2025-06-24 at 8 39 54 PM

Caesarh97:
That's perfect! Thank you so much!

ericboucher (approved):

PR #1160, opened 2025-08-01
More Collection names

Caesarh97:
Hi Eric, 

I added some more collection names to make some prettier and more informative URLs. 

When running yarn test -u. I ran into some issues unrelated to these changes. See Slack for more info.


packages/website/src/routes/Dashboard/index.tsx
```diff
@@ -32,6 +32,12 @@ export const collections: Record<string, number> = {
   tnc: 837,
   hawaii: 838,
   malaysia: 839,
+  KAUST: 850,
```
ericboucher, line 35: let's do lower case only please for consistency

PR #1161, opened 2025-08-03
Ensure map is ready when doing fly to

yaodingyd:


packages/website/src/routes/HomeMap/Map/index.tsx
```diff
@@ -58,12 +58,25 @@ const currentLocationMarker = L.divIcon({
...
+  onMapReady: () => void;
 }) => {
-  useMapEvents({
+  const map = useMapEvents({
     baselayerchange: (e) => onBaseLayerChange(e),
   });
+
+  useEffect(() => {
+    if (map) {
+      const timer = setTimeout(() => {
```
ericboucher, line 72: Is this necessary?

This seems to be enough:

```
  useEffect(() => {
    if (map) {
      onMapReady();
    }
  }, [map, onMapReady]);
```

but maybe I missed something?

ericboucher (approved):

Issue #1162, opened 2025-08-15
See the map/data for a date in the past


ericboucher:
It would be interesting to be able to see the world / Aqualink map and site data at a set date that is not today but in the past.
For example, to look back to at coral bleaching events in Australia - https://www.bbc.com/news/articles/cvg3pp52m65o

Let's investigate how hard it would be to adapt the Map component first and then the sites page,

piyzard:
/attempt https://github.com/aqualinkorg/aqualink-app/issues/1162

sirrodgepodge:
I'm attempting this bounty

justinleeyang:
Quick follow-up on #1162: I am still actively working on this. Please let me know if you want me to narrow scope to a smaller first PR.

sungdark:
我来认领

SyedHannanMehdi:
Working on this — CashClaw autonomous agent (@SyedHannanMehdi). PR coming shortly.

SyedHannanMehdi:
Working on this — CashClaw autonomous agent (@SyedHannanMehdi). PR coming shortly.

SyedHannanMehdi:
Working on this — CashClaw autonomous agent (@SyedHannanMehdi). PR coming shortly.

SyedHannanMehdi:
Working on this — CashClaw autonomous agent (@SyedHannanMehdi). PR coming shortly.

SyedHannanMehdi:
Working on this — CashClaw autonomous agent (@SyedHannanMehdi). PR coming shortly.

apollo-jpg:
/attempt #1162

I have already submitted PR #1278. Implementation plan/status:
- Add optional historical date query support to GET /sites.
- Populate map collection data from the latest daily_data row on or before the selected date.
- Wire a map date control through the website store and service layer.
- Keep current live/latest behavior unchanged when no date is selected.

I am following up on CI failures before requesting review.

tejas3732:
Hi, I’d like to work on this bounty, but I see existing PRs #1235 and #1278.

Before I duplicate work: are you still open to an alternative implementation, or is there a specific gap in the current PRs that I can address? I can focus on making the historical-date map flow merge-ready with tests and screenshots/video.

wangedmund77-cmyk:
/attempt #1162

Plan:
- Review the existing open PRs and maintainer feedback first so I avoid duplicating work or repeating rejected approaches.
[...]

PR #1177, opened 2025-10-18
View historical map and site data 

Excellencedev:
/claim #1162 
/closes #1162 

Add a date picker to the map to view site data and WMS overlays as of a selected past date.

The backend /sites endpoint now accepts an optional at parameter to return collectionData as of that timestamp, derived from time-series data. The frontend Home Map includes a date picker that updates site markers and passes the selected date as a TIME parameter to the WMS SST Anomaly layer. Sofar layers remain current as they do not support time parameters.

Demo - https://github.com/user-attachments/assets/fcaab15d-2682-477e-95c7-6ef7df7d3e8e


Excellencedev:
@ericboucher Review please

packages/website/src/routes/HomeMap/Map/sofarLayers.tsx
```diff
@@ -90,6 +91,8 @@ export const SofarLayers = ({ defaultLayerName }: SofarLayersProps) => {
 
 interface SofarLayersProps {
   defaultLayerName?: MapLayerName;
+  // ISO time parameter for WMS time-enabled layers
```
ericboucher: Are they all time-enabled? How can we check that we only display and overlay data for the same date?

Excellencedev: Not all layers were time-enabled originally, only WMS layers supported the `time` parameter. To ensure consistent historical data display across all overlays, I've added time parameter support to SOFAR layers as well, assuming their API supports historical queries (similar to WMS). I've added a code comment to verify with SOFAR API documentation that the NOAA Coral Reef Watch model supports historical data. If it doesn't support arbitrary historical dates, we may need to limit historical viewing to the date ranges available in the SOFAR data.

ericboucher:
@Excellencedev what's blocking you? You can generate mock data locally, but you can also email me at `hello at ovio.org` and I will send you a read only access to one of our databases
Anything else blocking you?

Excellencedev:
@ericboucher I will finish up this PR in the following days. Stay tuned
I will also provide demo video !

Excellencedev:
@ericboucher I have the demo video for this PR ready showing a working version where changing the dates changing the colors of the spotters and markers.
Pls review and merge

https://github.com/user-attachments/assets/fcaab15d-2682-477e-95c7-6ef7df7d3e8e

PR #1278, opened 2026-05-11
feat: add historical date view for map sites

apollo-jpg:
Fixes #1162

## Summary
- add an optional `date` query parameter to `GET /sites`
- when `date` is provided, populate site `collectionData` from the latest daily-data row on or before that date
- add a map date control on the home map and reload site overview data when it changes
- keep normal latest-data behavior unchanged when no date is selected
- add API regression coverage for historical map collection data

## Verification
- `./node_modules/.bin/tsc -p packages/api/tsconfig.json --noEmit`
- `./node_modules/.bin/jest --config ./packages/api/test/jest.json --runInBand packages/api/test/app.spec.ts` was attempted, but this local install fails before tests execute because the native `sharp` module cannot be loaded on darwin-arm64 due to a code-signature error.
- `./node_modules/.bin/tsc -p packages/website/tsconfig.json --noEmit` was attempted; it reports existing project-wide type errors unrelated to this change, including `tap` on an existing Leaflet `MapContainer`, React type duplication around `HashLink`, and existing mock/site type mismatches.

/claim #1162


apollo-jpg:
Pushed a CI stabilization follow-up for this PR in `41abd6ee`:

- Updated audit-forced dependencies (`axios`, `fast-xml-parser`, `hono`, `uuid`) so `yarn audit` passes.
- Made `BACKEND_BASE_URL` optional only under `NODE_ENV=test`, matching fork CI where that env var is not present.
- Gated the live Sofar API tests behind `SOFAR_API_TOKEN`, so fork CI without protected secrets skips those external integration checks instead of failing with 401/empty data.
- Froze the HomeMap snapshot test date and updated the snapshot for the new date control.

Local verification run:
- `yarn run audit` passes.
- `tsc -p packages/api/tsconfig.json --noEmit` passes.
- `SOFAR_API_TOKEN= NODE_ENV=test jest --config ./packages/api/test/jest.json --runInBand packages/api/src/utils/sofar.test.ts packages/api/src/workers/dailyData.test.ts` skips the external-token tests cleanly.

apollo-jpg:
Follow-up pushed in `bc3bfce3` after the first rerun:

- Ran the API lint path locally with Yarn 1 and fixed the Prettier warnings from the Sofar test gate.
- Reset the survey `deleteFile` mock between success/failure cases so the e2e delete tests do not inherit stale mock state.

Local verification:
- `yarn --cwd packages/api lint` passes with `--max-warnings 0`.
- `tsc -p packages/api/tsconfig.json --noEmit` passes.
- `yarn run audit` passes.

apollo-jpg:
Pushed `89fe48f1` to cover the remaining fork-CI survey delete failures. The Google Cloud delete mock now targets `GoogleCloudService.prototype.deleteFile`, so it covers all provider instances created by the app module during e2e tests. Re-verified locally: API lint passes and API TypeScript check passes.

apollo-jpg:
Pushed `1976a55f` after the latest API rerun. The remaining fork-CI failures were only storage-delete e2e cases while `GCS_BUCKET` is empty in the fork workflow, so those storage-dependent delete cases now skip when GCS config is unavailable. Local API lint and TypeScript check pass.

ericboucher:

PR #1289, opened 2026-05-20
Home map: add "As of" date selector

RaphaelS-git:
Fixes #1162

Changes:
- API: supports `asOf` (ISO-8601) on `GET /sites` and `GET /sites/:id` to compute `collectionData` as-of a timestamp.
- Website: adds an "As of" date picker overlay on the home map, re-fetching sites with `asOf`. Includes a "Latest" reset button.

Notes:
- `packages/website` tests pass (vitest).
- `packages/api` tests currently fail in CI/local when required env vars / external service auth are missing; failures are unrelated to this change.


ericboucher:

PR #1296, opened 2026-05-25
Add historical site date support

duongynhi000005-oss:
/claim #1162

Adds `date` query support to sites APIs, historical collectionData mapping from `daily_data`, and a homepage date picker/URL sync for map + site refetches. Includes API integration tests.


ericboucher:

PR #1300, opened 2026-05-28
Add historical date support to map and site pages

jamilahmadzai:
/claim #1162

Closes #1162

## Summary
- Add optional `date` support to `GET /sites` and `GET /sites/:id`, backed by the latest daily-data row on or before the selected day.
- Add URL-backed historical date controls on the map and site detail pages.
- Preserve the selected date when opening site pages from map popups and selected-site cards.
- Bound site-detail daily chart requests to the selected day so the detail view, cards, and chart window stay aligned.
- Add focused API and website coverage for the historical query flow.
- Make the historical-date snapshots deterministic by fixing their clock and mock daily-data timestamp.
- Keep fork CI independent of unavailable backend/GCS/SOFAR secrets by supplying test-only defaults and skipping token-backed live integration tests when the token is absent.

## Demo
https://files.catbox.moe/awrqiu.mp4

## Payout
Algora payout: @jamilahmadzai

## Validation
- `YARN_IGNORE_ENGINES=1 corepack yarn install --frozen-lockfile`
- ESLint on all changed TypeScript/TSX files with `--max-warnings 0`
- `YARN_IGNORE_ENGINES=1 corepack yarn workspace api test --runInBand src/utils/sofar.test.ts src/workers/dailyData.test.ts` (5 token-backed tests skipped as intended without `SOFAR_API_TOKEN`)
- `YARN_IGNORE_ENGINES=1 corepack yarn workspace api build`
- `YARN_IGNORE_ENGINES=1 corepack yarn workspace website build`
- `YARN_IGNORE_ENGINES=1 corepack yarn workspace website test src/routes/SiteRoutes/Site/index.test.tsx src/routes/HomeMap/index.test.tsx src/services/siteServices.test.ts` (3 files, 6 tests passed)
- `git diff --check origin/master...HEAD`

The focused API e2e path requires the repository's Postgres test service, which is not available in this local environment; the API compiles successfully and GitHub Actions covers the database-backed suite.

### DIFF
--- a/packages/api/src/collections/dto/collection-data.dto.ts
+++ b/packages/api/src/collections/dto/collection-data.dto.ts
@@ -24,6 +24,13 @@ type CollectionDataDtoType = Partial<
 >;
 
 export class CollectionDataDto implements CollectionDataDtoType {
+  @ApiPropertyOptional({
+    description:
+      'Timestamp of the daily observation selected for a historical request',
+    example: '2024-04-10T23:59:59.999Z',
+  })
+  observationDate?: Date;
+
   @ApiPropertyOptional({ example: 28.05 })
   bottomTemperature?: number;
 
--- a/packages/api/src/config/config.service.ts
+++ b/packages/api/src/config/config.service.ts
@@ -38,7 +38,9 @@ class ConfigService {
     return value;
   }
 
-  API_URL = this.getValue('BACKEND_BASE_URL', true);
+  API_URL =
+    this.getValue('BACKEND_BASE_URL', process.env.NODE_ENV !== 'test') ||
+    'http://localhost/api';
 
   public ensureValues(keys: string[]) {
     keys.forEach((k) => this.getValue(k, true));
--- a/packages/api/src/sites/dto/filter-site.dto.ts
+++ b/packages/api/src/sites/dto/filter-site.dto.ts
@@ -5,6 +5,7 @@ import {
   IsInt,
   IsEnum,
   IsBooleanString,
+  IsDateString,
 } from 'class-validator';
 import { Type } from 'class-transformer';
 import { ApiProperty } from '@nestjs/swagger';
@@ -37,4 +38,9 @@ export class FilterSiteDto {
   @IsOptional()
   @IsBooleanString()
   readonly hasSpotter?: string;
+
+  @ApiProperty({ example: '2024-04-15', required: false })
+  @IsOptional()
+  @IsDateString()
+  readonly date?: string;
 }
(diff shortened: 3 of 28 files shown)


jamilahmadzai:
Refreshed this PR onto current `master` (`169ae614`) and force-with-lease pushed head `e6fa0fb6`.

The refresh removes the old dependency/security lockfile churn from the PR, keeps only the historical-date feature and regression coverage, and adds a deterministic clock to the date snapshots.

Current local validation:
- changed-file ESLint: pass
- API build: pass
- website build: pass
- focused website tests: 3 files / 6 tests pass
- diff check: pass

The API e2e tests still require the repository Postgres test service; that service is not available locally, so GitHub Actions remains the source of truth for the DB-backed suite.

jamilahmadzai:
Follow-up for the API CI failure: the job received empty `BACKEND_BASE_URL`, GCS, and SOFAR secret values on this fork PR. I restored the prior branch's four targeted test/config safeguards without restoring any dependency or lockfile churn, and pushed head `24c9593b`.

Local verification for the follow-up:
- changed-file ESLint: pass
- API build: pass
- SOFAR/daily-data integration files: 5 token-backed tests skipped as intended without `SOFAR_API_TOKEN`
- diff check: pass

The database-backed e2e suite will run in GitHub Actions with its Postgres service.

saphid:
AI-assisted review disclosure: I reviewed this PR for issue #1162 using Codex and a synthetic-data-only local reproduction.

I found one narrow requested-date correctness gap. getHistoricalCollectionData deliberately selects the latest daily_data row on or before the requested date, but the returned collectionData does not include that row date. SiteDetails then constructs historicalTimestamp from the requested date itself. If a site has no row on the requested day, an older observation is therefore labelled in the cards as though it came from the requested day.

Synthetic reproduction: requested date 2024-04-15; selected source row 2024-04-10T23:59:59.999Z; mapped collectionData has no observation date; the site card timestamp is constructed from 2024-04-15. The mismatch reproduces under UTC, UTC+14, and UTC-12; the current focused tests all pass because they use a row from the exact requested day and do not assert observation-date provenance.

Suggested narrow fix: select and map the actual daily_data.date through the historical API result, use that value for card UpdateInfo timestamps, and add one sparse-date mapping test plus timezone-boundary coverage. I have not opened a competing implementation because maintainers retained this fuller map + site PR while closing map-only duplicates. I can prepare a focused follow-up if that is the preferred contribution path.

jamilahmadzai:
Thanks for flagging the historical-date issues. The API now returns the actual observation timestamp, and historical cards use that timestamp rather than relabeling older data with the selected day. Missing historical observations do not fall back to live values. I also added request-order guards for both the map and site views, including switching back to Latest while an older request is still in flight.

The current head is a234bb31. Both GitHub jobs pass their lint, build, and test steps, including the database-backed historical API tests and the new timestamp/request-order regressions. Their overall status is still red because the dependency security audit reports advisories in the unchanged lockfile; I reproduced the same audit failures on the previous head. No audit checks were disabled. Dependency remediation remains separate from this historical-date fix.

PR #1303, opened 2026-05-29
feat: add historical date viewing for map layers

misakaAron:
Closes #1162

Add a date picker to the homepage map allowing users to browse historical SOFAR data layers.

Changes:
- Date picker overlay positioned at bottom-right of the map
- Selected date passed as time query parameter to SOFAR API tile URLs
- Clear button to reset to latest real-time data
- Max date capped at today to prevent requesting future data
- Zero external dependencies (uses native HTML date input via MUI TextField)


ericboucher:

PR #1308, opened 2026-06-01
feat: add historical date picker to map and site list (#1162)

cyrillical00:
Closes #1162

Adds ability to view Aqualink map alert levels at any past UTC date.

**API changes:**
- : new optional Mon Jun  1 18:18:27 UTC 2026 query param (ISO YYYY-MM-DD)
- : queries  table for a specific day, same logic as  view but date-scoped
- : uses historical path when  is set, live path otherwise

**Frontend changes:**
- : forwards optional date as query param
-  / : accept date arg; re-fetch on date change
- : date picker UI (URL param ), re-fetches sites on change


ericboucher:

PR #1312, opened 2026-06-04
Add historical date support to home map

PiyushVerma-001:
Closes #1162.

## Summary
- Add an optional `date` query parameter to `GET /sites`.
- Use the nearest daily satellite record on or before the selected date to populate map/table `collectionData`.
- Add a shareable `?date=` date picker to the home map and keep navbar search requests aligned with the selected map date.
- Add API regression coverage for historical site collection data.

## Validation
- `git diff --check HEAD~1 HEAD`

I also attempted a full dependency install with the repository's required Node 20/Yarn 1.22.21 runtime, but the install could not complete because package fetching repeatedly hit network reachability/timeouts in my local environment.

### DIFF
--- a/packages/api/src/sites/dto/filter-site.dto.ts
+++ b/packages/api/src/sites/dto/filter-site.dto.ts
@@ -5,6 +5,7 @@ import {
   IsInt,
   IsEnum,
   IsBooleanString,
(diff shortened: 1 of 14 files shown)


ericboucher:

PR #1320, opened 2026-06-11
Ai assistant clean local storage

Caesarh97:
Users who had chatted with the AI assistant during the period when the greeting bug was active have contaminated chat histories stored in localStorage. Those histories contain greeting-format responses which the model then reproduces for every follow-up message, even after the underlying prompt fix was deployed.

Changing the localStorage key from `chat-history-${siteId}` to `chat-history-v2-${siteId}` forces all existing users onto a clean history on their next visit. No data is deleted, and old entries are simply ignored.

A guard is also in place to prevent this bug in the future.


ericboucher (approved):
It could be worth creating a shared constant work the key ?

Caesarh97:
> creating a shared constant work the key ?

Is my new commit what you had in mind? @ericboucher

PR #1328, opened 2026-07-04
Wave data: Add Open-Meteo API and replace Sofar Wave Model

Caesarh97:
Replaces wave data source with Open-Meteo Marine API. Wind stays on Sofar/GFS (Phase 2).

Sofar's wave model was snapping ~937 sites to land cells and hitting rate limits across ~6,000 sites/hour, leaving ~2,193 sites with no wave data. Open-Meteo's cell_selection=sea + batched multi-coordinate requests (100 sites/call) fixes both.
Changes:

- New SourceType.OPEN_METEO enum + migration (enum-add only, no data rewrite — avoids a transient mislabeling bug found in testing)
- hindcast-wind-wave.ts: wave now fetched via batched Open-Meteo calls; wind untouched; original getForecastData preserved for generate-stormglass-csv.ts compat
- Dashboard card: fixed hardcoded SOFAR MODEL badge → separate WIND/WAVE badges linking to correct sources (was silently mislabeling wave-sourced data as wind data in the non-Spotter case)

Follow-up (not in this PR): ~40% of sites (2,523/6,368) have wind data older than 3 days (29 completely missing). This can be added in a second PR.

Note: the Security Audit CI check is currently failing on master too (confirmed independently). Seems to be a pre-existing vulnerability in firebase-admin/multer/hono/typeorm/axios transitive deps, unrelated to this PR's changes. Not something I've touched or can safely fix here.

UPDATE:
One thing I couldn't finish myself due to permissions: I purchased the Open-Meteo $29/mo paid tier, and added OPEN_METEO_API_KEY to .env + functionsConfig.ts. But pushing it to Firebase's remote config for aqualink-programize needs yarn config:cloud-functions:programize run from packages/api, and I'm hitting a 403 (missing serviceusage.serviceUsageConsumer on that project, even as Owner). Could you run that command, or grant whatever role's missing? The key is sent via Slack.

### DIFF
--- a/packages/website/src/common/Chip/index.tsx
+++ b/packages/website/src/common/Chip/index.tsx
@@ -45,6 +45,7 @@ const useStyles = makeStyles((theme: Theme) =>
     button: {
       padding: 0,
       height: '100%',
+      minWidth: 58,
     },
   }),
 );
--- a/.github/workflows/build_api.yml
+++ b/.github/workflows/build_api.yml
@@ -50,9 +50,9 @@ jobs:
         with:
           path: |
             **/node_modules
-          key: ${{ runner.os }}-yarn-v2-${{ hashFiles('yarn.lock') }}
+          key: ${{ runner.os }}-yarn-v3-${{ hashFiles('yarn.lock') }}
           restore-keys: |
-            ${{ runner.os }}-yarn-v2-
+            ${{ runner.os }}-yarn-v3-
 
       - name: Install dependencies if needed.
         id: install
--- a/.github/workflows/build_api.yml
+++ b/.github/workflows/build_api.yml
@@ -50,9 +50,9 @@ jobs:
         with:
           path: |
             **/node_modules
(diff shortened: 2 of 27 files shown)


ericboucher:
@copilot update libs to fix api and frontend security CI step

packages/website/src/common/Chip/index.tsx
```diff
@@ -45,6 +45,7 @@ const useStyles = makeStyles((theme: Theme) =>
     button: {
       padding: 0,
       height: '100%',
+      minWidth: 58,
```
ericboucher, line 48: Why is this needed?

Caesarh97: MUI's Button has a default minWidth: 64px. Fine for chips with an icon, but our new icon-less WIND/WAVE badges are shorter than that, so they had extra empty padding. This overrides it to fit the content instead.

ericboucher:
@copilot fix security audit

ericboucher:
@copilot fix ci security audits

PR #1349, opened 2026-08-02
feat: browse historical map data by date

genaaredes-ui:
## Summary
- add a date query to retrieve the latest available daily map snapshot on or before that date
- add a URL-persisted map date picker for historical browsing
- add API coverage for historical daily data and keep the latest-data path unchanged

## Validation
- git diff --check
- TypeScript transpile validation for all 10 changed TS/TSX files
- Full Yarn tests/build could not run locally because the workspace has no installed dependencies or package manager

Resolves #1162


ericboucher:

PR #1350, opened 2026-08-02
feat: view site data for a historical date

FrancoLDG:
## Summary

- Add an optional `date=YYYY-MM-DD` filter to the public sites endpoint.
- Resolve each metric to the latest recorded value on or before the selected date.
- Add a map date picker and reload site data when the date changes.

Closes #1162

## Validation

- `git diff --check`
- Full build/tests not run locally because the repository dependencies were unavailable in the execution environment.

/claim #1162


ericboucher:

PR #1211, opened 2026-01-11
Migrate vitest to rstest

yaodingyd:
Finish the migration to rsbuild tool chain to remove the last one of using vitest/vite and switch to rstest. This makes the compelet tooling on rstack

PR #1356, opened 2026-08-12
Close the delete confirmation dialog once the deletion succeeds

### LINKED ISSUE #654: 404 when deleting a survey
While deleting a survey, I got the following 404 error. However, I think that what happened is that I deleted the survey on my first click, but the box stayed on. So when I clicked a second time, I got a 404. But the survey was indeed deleted and doesn't show up anymore.

Let's investigate this behavior and make sure the flow works as expected.

Screen Shot 2022-01-03 at 10 34 44 AM

eastagiletracker:
This PR proposes closing the delete confirmation dialog once the deletion succeeds, so a second click can no longer re-send the request and surface a 404 (Fixes #654). We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/228. You can sign in with your GitHub ID to claim ownership of the project.

## The fix

`DeleteButton` never closed its confirmation dialog on the success path. `onDelete` awaited `onConfirm()`, called `onSuccess?.()` and then left `open` set to `true`, so after a successful delete the box stayed on screen with "Yes" still armed. Clicking it again re-sent the delete for a resource that was already gone, and the 404 came back into the dialog's error alert — exactly the sequence described in #654 ("I deleted the survey on my first click, but the box stayed on. So when I clicked a second time, I got a 404"). Nothing guarded the in-flight window either, so a quick double-click fired two concurrent deletes for the same resource.

The change is confined to `packages/website/src/common/DeleteButton/index.tsx`: call the existing `handleClose()` after `onSuccess?.()`, return early when a request is already in flight, and set `disabled: loading` on both dialog actions so the buttons reflect that state. The failure path is deliberately untouched — on error the dialog still stays open and still shows the message, which is what makes the error recoverable. This is the shared delete control, so both call sites benefit: survey deletion in `SiteDetails/Surveys/SurveyCard` and data-upload deletion in `SiteRoutes/UploadData/HistoryTable`.

Reproduced on `master` at 03d3ac2 with three tests added to the existing `packages/website/src/common/DeleteButton/index.test.tsx`. Against the unmodified component (`yarn test` in `packages/website`):

```
 ✓ Delete Button > should render with given state from Redux store
 × Delete Button confirmation dialog > closes the dialog once the deletion succeeds
   → expected true to be false // Object.is equality
 × Delete Button confirmation dialog > does not issue a second deletion while the first one is in flight
   → expected "spy" to be called 1 times, but got 2 times
 ✓ Delete Button confirmation dialog > keeps the dialog open and surfaces the error when the deletion fails
 Tests  2 failed | 2 passed (4)
```

With the change applied all four pass. The third test is a control that passes both before and after, pinning the error behaviour so the fix cannot quietly swallow it. Restoring the old `index.tsx` and re-running turns the first two red again, which is the quickest way to verify the diff does what it says.

The full website suite was run before and after on the same tree: 65 files passed with one pre-existing failure (`src/routes/MonitoringRoutes/PromptsEditor/index.test.tsx`, a snapshot that is already red on `master` and unrelated to this change) in both runs, going from 90 to 93 passing tests. No new failures. `yarn lint` in `packages/website` is clean, and the existing `DeleteButton` snapshot is unchanged, since `disabled: false` renders identically to the previous markup.

One note on merge order: PR #1211 (vitest to rstest) touches test tooling broadly and may want a trivial rebase here if it lands first.

## How this was managed

This work was tracked on a board imported from this repository's own issues and pull requests, on the story for issue #654: https://eastagiletracker.com/projects/228/stories/97604 — the board itself is at https://eastagiletracker.com/projects/228.

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)


ericboucher:
Thanks for your contribution @eastagiletracker !!

PR #1361, opened 2026-08-24
New Grok model added

Caesarh97:
Updated the Grok model to a new model. Added `reasoning_effort: 'low',` for faster response time. Now we're matching production with a 5-7 second response time for greetings and short questions. Grok 4.6 has a default effort level of High, but responses took between 20 and 25 seconds, which I think is way too long, especially for the greeting message.

### DIFF
--- a/packages/api/src/utils/aiGrokService.ts
+++ b/packages/api/src/utils/aiGrokService.ts
@@ -2,7 +2,7 @@ import axios from 'axios';
 import { buildPromptWithContext, PromptMap } from './prompts';
 
 const GROK_API_URL = 'https://api.x.ai/v1/chat/completions';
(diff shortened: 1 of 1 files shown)


packages/api/src/utils/aiGrokService.ts
```diff
@@ -2,7 +2,7 @@ import axios from 'axios';
 import { buildPromptWithContext, PromptMap } from './prompts';
 
 const GROK_API_URL = 'https://api.x.ai/v1/chat/completions';
-const GROK_MODEL = 'grok-4-fast-reasoning'; // Fast and cost-effective
+const GROK_MODEL = 'grok-4.6'; // Fast and cost-effective
```
ericboucher: Change or remove the outdated comment

ericboucher (approved):

